PR #5, opened 2020-12-14
Add database growth scripts

bill-long:


dpaulson45:
@bill-long should add a README.md that provides how to use each of the scripts vs just looking at the comments. 

- space dump analysis should provide that this requires the database to be dismounted, OR being run on a suspended copy providing the issue is occurring there. The other thing is it should only be attempted to be done after the other script doesn't provide anything.

- Mailbox size comparison, should include how to collect the data for the before and how long to wait prior to running the script. Then we might need to include a Mailbox Folder Stats comparison as well.

bill-long:
@dpaulson45 I like having documentation at the top of the script for several reasons:

* The documentation travels with the script. If you download it and you're trying to use it somewhere, you don't have to go find a separate file. You just cat the script.
* When you change the script, you don't have to remember to update some other file. Sort of like in C#, where we describe a method and its parameters right above the method.

However, I agree that we need to think about where to store information about troubleshooting logic for an issue, i.e. first check x with this script, then check y with that other script. Maybe that should go in a README in this folder.

dpaulson45:
@bill-long maybe do a little bit of both like you said. This way when view the directory on the GitHub site, they know what each file is for and roughly how to use it prior to just clicking on it.

dpaulson45 (approved):
Approving for now, but would like to see a README down the road to assist.

PR #8, opened 2020-12-23
Setup Scripts

dpaulson45:


bill-long:
Do we really need to put authors in script comments? People can just look at the commit history. Ideally, we'd have more people contributing, expanding, and fixing these scripts than some predefined author.

bill-long:
We also need to decide on a consistent set of styling rules and shift-alt-F these files in VSCode. For example, in some places we have Javascript-style braces (my own preference), and it some places we have C# style braces. In some places we have <# #> multiline comments, and in other places we have # multiline comments. But we can address styling consistency in a separate PR.

bill-long (dismissed):

bill-long (approved):

PR #23, opened 2021-02-06
SourceSideValidation updates

bill-long:


dpaulson45 (approved):

PublicFolders/SourceSideValidations/SourceSideValidations.ps1
```diff
@@ -41,23 +41,23 @@ $folderData = [PSCustomObject]@{
 $ipmSubtree | ForEach-Object { $folderData.ParentEntryIdCounts[$_.ParentEntryId] += 1 }
 $ipmSubtree | ForEach-Object { $folderData.EntryIdDictionary[$_.EntryId] = $_ }
 $nonIpmSubtree | ForEach-Object { $folderData.NonIpmEntryIdDictionary[$_.EntryId] = $_ }
-$anyDatabaseDown = $false
+$script:anyDatabaseDown = $false
```
dpaulson45, line 44: Could use `private` vs `script` scope if you wanted to. 

Inside the foreach loop, would need to have the scope set to `script` to get around the bug. However, the variable wont be set for the whole script this way.

PR #65, opened 2021-03-05
Fixed urls to use archive.org as  blog removed

Absoblogginlutely:
**Issue:**
URLs in the readme refer to the now defunct You Had Me at Helo blog. The links just redirect to the default Exchange community blog as opposed to the specific blog post

**Reason:**
Updating links so that an archived copy of the original blog post is viewable so the users can read about the scripts purpose

**Fix:**
Updated links to go to archive.org's copy of the original blog post. 

**Validation:**
Links in readme.md work.


dpaulson45:
Not going to include the archive.org location. Just put in the correct location to the blogs

#66

PR #68, opened 2021-03-05
Rewrite for remoting, objects and parallel processing

FriedrichWeinmann:
**Issue:**
Original script while doing its job didn't scale well

**Reason:**
Needed this to scale out to hundreds of Exchange Servers

**Fix:**
Rewrite as functions with remoting and parallel processing capabilities


bill-long:
Thank you! This is a huge change, so we'll need to do some testing before merging this.

bill-long:
Example output from this version:

image005

bill-long:
This branch seems to contain the original blog commands without the performance improvements from Test-Hafnium. For example:

image
...
image

Versus the current Test-Hafnium:

image
...
image

The current Test-Hafnium takes 6 seconds. The rewrite takes over 2 minutes. We need to bring over all the performance work.

bill-long:
Example output from latest revision. Note we now simply pipe the servers to the script:

image

dpaulson45 (approved):

dpaulson45 (approved):

PR #105, opened 2021-03-06
Report oldest log times

### LINKED ISSUE #98: Feature request: Show log retention time for Security/Test-ProxyLogon.ps1

It might be a problem, that Security/Test-ProxyLogon.ps1 checks logfiles which don't date back to a potential exploitation. If the logs are gone, nothing can be done about it, but at least it should be clear, that if the logs just go back for e.g. 3 days, it doesn't cover the perdiod until 26th Feb when the exploiting started.

**Describe The Request**
Please include the few lines of code to show the date range covered by the checked logfiles:
$exchangePath\Logging\ECP\Server\*.log
$exchangePath\Logging\OABGeneratorLog
$exchangePath\Logging\HttpProxy

**Additional context**
There are It providers, which have discovered, that disk space costs money and unless the clients pays extra for it, they reduce the log retention time to some days or even hours. That might cover basic debugging needs but not security.

We're not even talking about attackers which might have deleted parts of the logs ;)

bill-long:
Fixes #98 

Still need to discuss the right way to output this. It's kind of a mess.


bill-long:
Example with the days approach:

image

bill-long:
Example with OutPath:

image

bill-long:
I think this is still too ugly. The massive table below the "Nothing suspicious" message really clutters things up.

dpaulson45:
Should add a way to break up the results of each server. When i first read that, didn't realize that the results were from different servers. 

Might be ideal to do in the following order:

- write to host "results for server X"
- proxy logs ages 
- proxy logs test results - good or bad

bill-long:
_Without_ the new log file age stuff, the output from the current version of the script is pretty clean:

image

So we need a way to preserve that while also showing something about the log file ages.

2d4d:
How about merging this? Even a bad solution is better than no solution here. Lot's of orgs will scan their logs and think they're green, not even realizing that their logs don't cover a significant time span.

Josegto1 (approved):

bill-long:
We'll probably go with this for now. Still not great, but at least it indents and only takes up 1 line if your shell is wide enough.

image

dpaulson45 (approved):

Issue #179, opened 2021-03-09
[Issue] Get-AgeInDays fails on non-English locales?


abochmann:
**Describe the issue**
We have a couple of systems where the LogAgeDays.csv contains empty values only. The log files exist though, and are accessible when running the (Get-ChildItem ...).CreationTime - call manually. No error is reported by the script

Our current suspicion is that this happens when the Exchange host system is a Windows with a locale other than English. 

**Expected behavior**
The log age reporting works regardless of system locale.

**Script Output**
Example empty csv:
```
#TYPE System.Management.Automation.PSCustomObject
"Oabgen","Ecp","AutodProxy","EasProxy","EcpProxy","EwsProxy","MapiProxy","OabProxy","OwaProxy","OwaCalendarProxy","PowershellProxy","RpcHttpProxy"
"","","","","","","","","","","",""
```

bill-long:
Hrm, I changed my regional settings but it seems to be fine with commas:

![image](https://user-images.githubusercontent.com/4518572/110493725-834afe80-80b8-11eb-88c0-0444d7b01a92.png)

It's even perfectly happy to return decimals for some servers and commas for others, depending on the regional settings of the target server:

![image](https://user-images.githubusercontent.com/4518572/110493829-98c02880-80b8-11eb-9609-6cc5f815aeb7.png)

abochmann:
Hrm.

In our testing, we removed the Get-AgeInDays subroutine and then got the actual date in the output. We'll try some more to get this tracked down.

2d4d:
we got negative numbers from one machine:

#TYPE System.Management.Automation.PSCustomObject
"Oabgen","Ecp","AutodProxy","EasProxy","EcpProxy","EwsProxy","MapiProxy","OabProxy","OwaProxy","OwaCalendarProxy","PowershellProxy","RpcHttpProxy"
"-55,5","-55,2","-55,4","-55,4","-55,4","-55,4","-55,4","-55,4","-55,4","-55,4","-55,4","-55,4"

abochmann:
I tried the version from #183 and now get correct output 👍

```
#TYPE System.Management.Automation.PSCustomObject
"Oabgen","Ecp","AutodProxy","EasProxy","EcpProxy","EwsProxy","MapiProxy","OabProxy","OwaProxy","OwaCalendarProxy","PowershellProxy","RpcHttpProxy"
[...]

PR #183, opened 2021-03-09
Fix for issue where Get-AgeInDays swaps day and month when parsing $dateString parameter.

Magnus1990P:
**Issue:**
The Get-AgeInDays function swapped month and date during the TryParse causing incorrect age calculation.
The OS had LCID = 1044.
Eg. dateString input  "07.02.2021 17.00.46", time format string "dd.mm.yyyy HH.MM.ss"

**Reason:**
Improving the age calculation function when ran locally on the exchange server.

**Fix:**
Removed the String conversion of the parameter, so it does not convert to string.
Checks if the variable contains data and is of type [DateTime]. If not it returns an empty string.
If DateTime, it grabs the current timestamp (Get-Date) and subtracts the supplied parameter value from the current timestamp. Then returns the age in TotalDays as a string with 1 decimal point.
Uses the CultureInfo to set a default LCID

**Validation:**
Tested on a system with LCID 1044, which fixed the age calculation error.

### DIFF
--- a/Security/Test-ProxyLogon.ps1
+++ b/Security/Test-ProxyLogon.ps1
@@ -182,11 +182,11 @@ process {
 
                 function Get-AgeInDays {
                     param ( $dateString )
-                    if ( $dateString ){
+                    if ( $dateString -and $dateString -as [DateTime] ){
                         [System.Globalization.CultureInfo]$provider = [System.Globalization.CultureInfo]::InvariantCulture
                         $CURTIME = Get-Date
                         $age = $CURTIME.Subtract($dateString)
-                        return $age.TotalDays.ToString("N2")
+                        return $age.TotalDays.ToString("N1")
                     }
                     return ""
                 }
--- a/Security/Test-ProxyLogon.ps1
+++ b/Security/Test-ProxyLogon.ps1
@@ -181,17 +181,13 @@ process {
                 }
 
                 function Get-AgeInDays {
-                    param (
-                        [string]
-                        $dateString
-                    )
-
-                    $date = [DateTime]::MinValue
-                    if ([DateTime]::TryParse($dateString, [ref]$date)) {
-                        $age = [DateTime]::Now - $date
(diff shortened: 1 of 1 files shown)


Security/Test-ProxyLogon.ps1
```diff
@@ -181,17 +181,13 @@ process {
...
-                        [string]
-                        $dateString
-                    )
-
-                    $date = [DateTime]::MinValue
-                    if ([DateTime]::TryParse($dateString, [ref]$date)) {
-                        $age = [DateTime]::Now - $date
+                    param ( $dateString )
+                    if ( $dateString -and $dateString -as [DateTime] ){
+                        [System.Globalization.CultureInfo]$provider = [System.Globalization.CultureInfo]::InvariantCulture
```
bill-long, line 186: $provider is being flagged as an unused variable. Does this line do anything?

Also, can you run `.build\CodeFormatter.ps1 -Save` please? This will enforce the formatting rules and save the changes. If you run it a second time, it should then be clean, except for any PSAnalyzer errors.

bill-long:
This might fix #179

Security/Test-ProxyLogon.ps1
bill-long: I took care of the formatting piece but I'm still not sure about this $provider.

Magnus1990P: Sorry, the "[System.Globalization.CultureInfo]$provider = [System.Globalization.CultureInfo]::InvariantCulture" was left as part of trying to use the [DateTime]::TryParseExact  function.

Shall I generate a new pull request with this removed?

bill-long: Nah we can just update this one. I just wanted to check and make sure I wasn't missing something.

bill-long (approved):

bill-long:
Thank you!

Issue #108, opened 2021-03-07
This Tools is not run for Ex2010  ?


rezatbahtera:
Unable TO find type [PS Credential]: make sure that the assembly containing this step loaded 
![image](https://user-images.githubusercontent.com/59155410/110228802-9df56a00-7f36-11eb-9e88-19546efc8144.png)

bill-long:
Hmm... this _was_ working on 2010 initially. Investigating.

bill-long:
It's definitely broken when running _locally_ on 2010, and there are several changes needed to get it working due to the older version of PowerShell.

One option is to Enable-PSRemoting on 2010 and then run it from somewhere else, using the -ComputerName parameter to point it to the 2010 server. That does work.

Another option is to run the blog commands manually on 2010.

Keep in mind that the script is of limited use on 2010 even in a fully working state. The commands from the blog post search several folders that don't exist on 2010. I think the goal for this script will be to run the tests we can, remotely, against 2010, and fail gracefully where the tests don't apply to 2010.

dinonovak:
Confirming SBS 2011 when run locally is throwing the same error.
Please fix as this is the only server in infrastructure so no remote option possible.

mahmoudshoaala:
Try to bind the $credential function with = get-credential "domain\username"

dbacon247:
This script requires Powershell 3.0 and later, Exchange 2010 by default I believe has Powershell 2.0 installed.

Wizzdk:
Check if your shortcut to "Exchange Management Shell" includes "-version 2.0" in the parameters.
If so, remove this and try again.
![Exchange Management Shell](https://user-images.githubusercontent.com/4739672/110307573-5ab2fe00-7fff-11eb-8d92-1658d52521cd.png)

JeremyTBradshaw:
The script should start with `#Requires -Version 4` to prevent people trying to use it in the wrong version of PowerShell.

bill-long:
You can also remove the PSCredential requirement if that's the blocker. I haven't actually see a case where we need to pass credentials. Just delete this line:

`[pscredential]`

Or change it to some other data type:

`[string]`

live-motion:
I change [pscredential] to [string] and got it:

[PS] D:\Downloads>.\Test-ProxyLogon.ps1 -OutPath .\logs
cmdlet Write-Progress at command pipeline position 1
Supply values for the following parameters:
(Type !? for Help.)
Status:

[...]

PR #188, opened 2021-03-09
Allow exporting functions

bill-long:
- Drop version requirement to 3
- Remove hard PSCredential requirement
- Allow exporting functions so objects can be retrieved and
  piped for further processing.

image

See #108 for discussion.


bill-long:
Note this does have a breaking change for anyone who was already calling functions directly. The shape of the Cve26855 report object has changed. It now contains two things:

Hits - contains the old data, just the hits from the files.
FileList - full name of files containing hits, for collection or further investigation.

dpaulson45 (approved):

PR #201, opened 2021-03-09
Improvements to Get-Cve26855

nashiooka:
**Issue:**
Improve performance of Get-Cve26855

**Reason:**
1) using Select-String as a bool doesn't require reading the whole file.
2) Using Get-ChildItem to test file existence and prevent downstream errors is comparatively slow compared to other methods.
3) Using += to incrementally add to an array is slow. Especially if repeated.

**Fix:**
1) Use -Quiet switch to limit file read to the first match.
2) Use Test-Path to validate file before attempting to run Select-String
3) Use [ArrayList] in place of typical arrays.

**Other**

- Use variable with Join-Path to avoid extra "\" in the log path
- assign output properties to an array to cleanup Select-Object command.
- Removing unnecessary $filelist variable and array subexpression wrapping.

**Validation:**
Ran against files with spoofed bad entries and results looked ok...

### DIFF
--- a/Security/Test-ProxyLogon.ps1
+++ b/Security/Test-ProxyLogon.ps1
@@ -117,7 +117,6 @@ process {
                     $sw = New-Object System.Diagnostics.Stopwatch
                     $sw.Start()
 
-                    # Start you loop here:
                     For ( $i = 0; $i -lt $files.Count; ++$i ) {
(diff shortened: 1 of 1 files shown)


bill-long (dismissed):

bill-long (approved):

bill-long:
Looks good! Thanks Steven!

PR #203, opened 2021-03-09
Fix null ref error when response header is missing

justinhendricksmsft:
**Issue:**
When response header is missing it fails with ERROR: Script execution failed (use -d to debug)

**Reason:**
Fix null ref error

**Fix:**
Check that it's not null before string.match

**Validation:**
Provide if applicable


justinhendricksmsft:
Don't want to merge

bill-long (approved):

PR #239, opened 2021-03-10
Minor Improvement to Get-SuspiciousFiles

nashiooka:
**Issue:**
Nothing specific.

**Reason:**
Minor readability and performance improvements.

**Fix:**
> Store various filter strings and paths in variables for easier reference.
> Consolidate 2 `ForEach()` loops used for searching for lsass.*dmp files to 1 loop `ForEach-Object{}` referencing an array of the 2 paths.
> Convert `ForEach()` loop searching for archive files to `ForEach-Object{}`
> Use the same loop with an if block to substitute for original `Where{}`. This approach is usually faster and is ideal since already using `ForEach-Object{}`

**Validation:**
Tested by creating bogus .dmp & archive files in relevant locations.  Seems to work.


bill-long (approved):

bill-long:
Thank you!

Issue #260, opened 2021-03-11
Issue Test-proxyLogon.ps1 on Powershell 2.0


swissiws:
Windows 2008 R2, Exchange 2010, Powershell 2.0

You must provide a value expression on the right-hand side of the '-' operator.
At C:\scripts\Test-ProxyLogon.ps1:228 char:48
+                             If ( $_.Extension - <<<< in $zipFilter ) {
    + CategoryInfo          : ParserError: (:) [], ParseException
    + FullyQualifiedErrorId : ExpectedValueExpression

bill-long:
PowerShell 2 is not supported. It requires at least 3. Unfortunately it ignores the Requires directive since it isn't on the first line of the file.

Because Exchange 2010 is only vulnerable to CVE-2021-26857, the script only performs one of the tests from the blog:

Get-EventLog -LogName Application -Source "MSExchange Unified Messaging" -EntryType Error | Where-Object { $_.Message -like "*System.InvalidCastException*" }

You either run this manually, or you can enable PS Remoting and run the script from a different machine that has at least PowerShell 3.

PR #289, opened 2021-03-11
Minor Improvements

nashiooka:
**Issue:**
Nothing serious

**Reason:**
Minor readability & performance.

1. `Get-LogAge` and it's helper function `Get-AgeInDays`
2. `Get-Csv26858`
3. `Get-SuspiciousFiles`

**Fix:**
1. `Get-LogAge` & `Get-AgeInDays`; A few things to shorten lines. Splatting & pipeline functionality. Also a var to shorten path arguments.
2. `Get-Csv26858`, use `Test-Path` instead of `Get-ChildItem`,  while ensuring the Cve26858 object property is not a 1 null element array.
3. Change `Get-Suspicious` files to use `-contains` instead of `-in`.  Despite #Requires 3.0, this should address [issue# 260](https://github.com/microsoft/CSS-Exchange/issues/260),  Ex2010, PoSh 2.0 compatibility.

**Validation:**
Retested


Security/src/Test-ProxyLogon.ps1
```diff
@@ -207,7 +210,7 @@ begin {
 
                     Get-ChildItem -Path $env:ProgramData -Recurse -ErrorAction SilentlyContinue |
                         ForEach-Object {
-                            If ( $_.Extension -in $zipFilter ) {
+                            If ( $zipFilter -contains $_.Extension ) {
```
FriedrichWeinmann, line 213: This works equally well, but won't prevent some issues with PSv2:
PSCustomObject in the line below will likely only return hashtables, which in turn will cause interesting reports (but can be addressed on the caller side)

Security/src/Test-ProxyLogon.ps1
```diff
@@ -220,7 +223,11 @@ begin {
                 }
 
                 function Get-AgeInDays {
-                    param ( $dateString )
+                    param (
+                        [Parameter( ValueFromPipeline = $true )]
+                        $dateString
+                    )
```
FriedrichWeinmann, line 229: If you add acceptance from pipeline, make sure to put the main logic into a process block.
It may work, as we only offer one item each time, but should be considered good practice anyway.

If you don't add any of begin/process/end blocks, all your code executes in end.

FriedrichWeinmann (commented):
Mostly a nice cleanup, thanks for the upgrades.
Mostly agree with everything, but would prefer pipelining implemented right if we add it at all in Get-AgeInDays

Security/src/Test-ProxyLogon.ps1
nashiooka: RE: pipeline.  I presume you mean all the trappings of an advanced function.  I'm happy to do that, but I share your concern that it may be overkill for little ROI.

RE: 2.0, That's a great point, I was myopically responding to the specific complaint in #260.  While I'm not fond of the `New-Object` syntax maybe we should move to it.

However, 2.0 is so old at this point I'm not even sure it'll accept the hash as an arg to -Property.  I'd have to install the 2.0 engine just to test, but per other MS comment's on #260 I'm not sure how far we want to go to accommodate 2.0. Especially considering the `[PSCustomObject]@{...}` is leveraged elsewhere.

Security/src/Test-ProxyLogon.ps1
nashiooka: That I do know. I agree and prefer a more robust syntax (even if verbose) anyhow.  It works in this case because we only send a single value that then persists through to `End{}` See other comment, I'm not sure it's worth it.

bill-long: That's actually causing a build failure:

image

Hey Steven, please run `.build\CodeFormatter.ps1 -Save`. That will fix the formatting and save the changes automatically, and it will report any PSScriptAnalyzer issues that need to be corrected manually, such as the process block thing.

Once CodeFormatter comes back clean, if you push to this branch again the build should succeed. Thanks!

nashiooka: Understood. I'm sorry I'm having trouble running CodeFormatter due to environment security.  We can abandon this pull request for now, and I'll refile as soon as I can. Sorry for the trouble.

bill-long:
Closing this one for now. We can revisit in the future.

PR #305, opened 2021-03-11
Parallelizing Get-Cve26855 and fixing single hit scenario

SharmaAkash1:
**Sharing to get any feedback on the changes.**

**Issue:**

Get-Cve26855 is currently single threaded and hence slow. The script also fails to show output when there is only hit recorded in the httpproxy logging folder. 

**Changes/Fixes:**

1. Parallelized the search of log files in the Logging\HttpProxy folder. Steps:

          1. We create batches of the files
          2. Each batch of files is searched using a job created using start-job and the results are then accumulated
          3. A common progress bar is shown for all jobs. 
                  a. The write-progress function in GetCve26855LogsScriptBlock creates the progress object for each of the jobs
                  b. The progress objects of all jobs are then utilized to update the common progress bar in showJobProgress.

2. Fixed the following scenarios where log detection does not work.

          1. If there is only one hit (one attack) in the https proxy logs
          2. If there is only one log file in the httpproxy folder (not likely to happen as IIS would ensure that there are more files in httpsproxy log folder structure)
          3. Cause of above: Powershell causes automatic conversion of System.Collections.ArrayList to PSCustomObject if there is only one object in the list.
          4. Fix:
                    i. We explicitly ensure that the hits and fileCount objects are arrays even in case when there is only 1hit/ 1 file(with dangerous logs)
                    ii. Also for the issue 2b we just strongly type the files object that stores the paths of files in the logging/httpproxy folder hirearchy.
          
**Testing/validation**

In progress

### DIFF
--- a/Security/src/Test-ProxyLogon.ps1
+++ b/Security/src/Test-ProxyLogon.ps1
@@ -1,9 +1,12 @@
-﻿# Checks for signs of exploit from CVE-2021-26855, 26858, 26857, and 27065.
+# Checks for signs of exploit from CVE-2021-26855, 26858, 26857, and 27065.
 #
 # Examples
 #
 # Check the local Exchange server only and save the report:
 # .\Test-ProxyLogon.ps1 -OutPath $home\desktop\logs
+#
+# Provide the number of threads using the -Numthreads (defaults to 5 if not specified):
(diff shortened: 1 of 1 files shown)


SharmaAkash1:
Need to put this on hold for some time

bill-long:
Let's close this one for now. We can revisit.

PR #334, opened 2021-03-12
CopyMissingDlls

dpaulson45:
**Issue:**
Seeing a lot of cases that are missing DLLs after setup
Created a script that will copy the DLLs from the ISO to the server.


.build/Build.ps1
```diff
@@ -28,7 +28,7 @@ New-Item -Path $distFolder -ItemType Directory | Out-Null
 $scriptFiles = Get-ChildItem -Path $repoRoot -Directory |
     Where-Object { $_.Name -ne ".build" } |
     ForEach-Object { Get-ChildItem -Path $_.FullName *.ps1 -Recurse } |
-    Where-Object { ! $_.Name.Contains(".Tests.ps1") }
+    Where-Object { ! $_.Name.Contains(".Tests.ps1") -and ! $_.Name.Contains(".NotPublished.ps1") }
```
bill-long, line 31: Since we spelled out `-and` can we spell out `-not` for consistency, instead of `!`

Setup/DllMappings/DllMaper.NotPublished.ps1
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    foreach ($fileName in $group.Group.FileName) {
+        $currentMappings.DllFileNames.Add($fileName)
+    }
+
+    $mapperObject.Add($currentMappings)
+}
+
+<#For 2013
```
bill-long, line 193: Is this intentionally left in?

bill-long (changes requested):

Setup/DllMappings/DllMaper.NotPublished.ps1
dpaulson45: Yep, but i should be able quickly change it to use a variable actually that way i don't need it.

bill-long (approved):

PR #341, opened 2021-03-12
Providing output from script to indicate IssuesFound & Fixing same na…

SharmaAkash1:
The following changes have been made:

1. We are returning a simple object that contains the attribute IssueFound. This allows another script calling test-proxylogon to figure out if issues were found.

2. Issue: If the attacker places files with same names in each of the folders c:\root", "$env:WINDIR\temp", $env:ProgramData,  current script would try to copy all 3 files into the suspicious files repo and hence overwrite one another.

To prevent this we are attaching a file number to the name of the file. This ensure files with same name are not overwritten.

Also in a test scenario there were some folder created with names like Hello.Zip, etc. Copy-Item is unable to overwrite these during copy and leads to exceptions being thrown.  To fix such issues we just recreate the suspicious files folder when we run the script again.


bill-long:
We solved this a slightly different way.

PR #357, opened 2021-03-15
Proper Check for Admin

dpaulson45:
Can't have a failed admin check call into ` Set-LogActivity ` as this sets values in the registry. Can't set values in the registry without being an Admin.

Running without Admin before fix:

image

After fix:

image


dbacon247 (changes requested):
Please change write-host to write-error on line 721

dpaulson45:
@dbacon247 done

dbacon247 (approved):

PR #400, opened 2021-03-20
Display MU message in Color Red

xzong:
Feedback on the color of display MU message to change it to Red.


dbacon247:
Fixes in PR #411

PR #481, opened 2021-04-19
Add AutoUpdate feature, fix bug in SU checking

ethanb-msft:
**Issue:**
There was a bug in how we determine the latest SU. 

**Reason:**
Fixes a bug adds a feature to help customers stay current.
Also added autoupdate feature.

**Fix:**
Changed logic for determining latest SU to also match SUs higher than March SU.

**Validation:**
Provide if applicable

### DIFF
--- a/Security/README.md
+++ b/Security/README.md
@@ -12,8 +12,9 @@ This script contains mitigations to help address the following vulnerabilities.
 
 * CVE-2021-26855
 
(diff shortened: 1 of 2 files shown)


bill-long:
@dpaulson45 I wrote Test-ScriptVersion as a proof of concept while these folks were deciding how to do their version checks. There was a thread on the best way to validate the signature. Turns out, the best way to do it seems to be the way we are doing it in Test-ScriptVersion.

So, this might get switched to consume Test-ScriptVersion in the future. For now we're going to merge it with its own implementation so we can get the fix out.

bill-long (approved):

PR #512, opened 2021-04-22
Moved Get-VisualCRedistributableVersion to Shared

dpaulson45:
Also moved the dependent of Invoke-ScriptBlockHandler to Shared as well. Removed the Class that we no longer call because of the move to a Shared structure.


Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-OperatingSystemInformation.ps1
```diff
@@ -68,7 +68,7 @@ Function Get-OperatingSystemInformation {
         -ScriptBlockDescription "Getting Current Time Zone" `
         -CatchActionFunction ${Function:Invoke-CatchActions}
     $osInformation.TLSSettings = Get-AllTlsSettingsFromRegistry -MachineName $Script:Server -CatchActionFunction ${Function:Invoke-CatchActions}
-    $osInformation.VcRedistributable = Get-VisualCRedistributableVersion
+    $osInformation.VcRedistributable = Get-VisualCRedistributableVersion -ComputerName $Script:Server
```
bill-long, line 71: Can we make this a parameter to Get-OperatingSystemInformation instead of assuming a script-scoped variable will be present somewhere?

bill-long (changes requested):

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-OperatingSystemInformation.ps1
bill-long: This problem is all over HealthChecker, so we will address it in a future PR focused on that.

bill-long (approved):

PR #548, opened 2021-04-26
Enforce BOM at release and at dev time

### LINKED ISSUE #528: [Issue] format issue in Get-MRMDetails.ps1
**Describe the issue**
Get-MRMDetails.ps1 failed to run, error shows single quote mark issue in line 51.
image

**Script Output**
image

bill-long:
Issue:

We do not use a consistent encoding across the repo. Some scripts have BOMs and some don't. When we release, we currently release scripts with no BOMs.

See [https://docs.microsoft.com/en-us/powershell/scripting/dev-cross-plat/vscode/understanding-file-encoding](https://docs.microsoft.com/en-us/powershell/scripting/dev-cross-plat/vscode/understanding-file-encoding).

Per that article, in Windows PowerShell, which is where our released scripts will usually be executed, the default encoding is Windows-1252. This means that a file with no BOM, like we currently release, is _not_ assumed to be UTF8. This will be problematic if we need to release scripts that include Unicode characters.

In PowerShell 6+, the default encoding is UTF8, so a BOM is not needed. VSCode will create files with UTF8 no BOM by default.

Proposal:

This PR changes our build process to always produce released scripts as UTF8 _with_ a BOM. This ensures that Windows PowerShell does not assume they are Windows-1252.

This PR also changes our code formatter script to enforce UTF8 _with_ a BOM at dev time. This ensures that our scripts behave the same at dev time as they do at release.

Additional Notes:

This change is agnostic to other files in the repo, such as text files.

We're taking a dependency on the EncodingAnalyzer module from PSGallery to check for a BOM.

Fixes #528 where Windows PowerShell can't figure out what to do with a curly quote because it thinks it's Windows 1252. Though we should probably change those to non-curly quotes anyway.

### DIFF
--- a/.build/Build.ps1
+++ b/.build/Build.ps1
@@ -1,4 +1,4 @@
-[CmdletBinding()]
+﻿[CmdletBinding()]
 param (
 
 )
@@ -100,7 +100,7 @@ $scriptFiles | ForEach-Object {
     $expandedScript.Insert(0, "")
     $expandedScript.InsertRange(0, $disclaimer)
 
-    Set-Content -Path (Join-Path $distFolder $scriptName) -Value $expandedScript
+    Set-Content -Path (Join-Path $distFolder $scriptName) -Value $expandedScript -Encoding utf8BOM
     $scriptVersions += [PSCustomObject]@{
         File    = $scriptName
         Version = $buildVersionString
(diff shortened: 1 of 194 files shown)


bill-long:
I'm still debating the no BOM at dev time. While it's nice to match the modern defaults, this means we could potentially see different behavior in the released script versus the dev time script.

tweekerz (approved):

dpaulson45 (approved):

Issue #536, opened 2021-04-26
[Issue] HealthChecker - Exception when DisabledComponents is set to 0xffffffff


dpaulson45:
**Describe the issue**
Customers reported the following issue:

```
Exception setting "IPv6DisabledComponents": "Cannot convert value "0xffffffff" to type "System.Double". Error: "Input 
string was not in a correct format.""
At C:\Temp\HealthChecker.ps1:2947 char:5
+     $osInformation.NetworkInformation.IPv6DisabledComponents = Invoke ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], SetValueInvocationException
    + FullyQualifiedErrorId : ExceptionWhenSetting
 
at Get-OperatingSystemInformation, C:\Temp\HealthChecker.ps1: line 2947
at Get-HealthCheckerExchangeServer, C:\Temp\HealthChecker.ps1: line 2787
at Main, C:\Temp\HealthChecker.ps1: line 5615
at <ScriptBlock>, C:\Temp\HealthChecker.ps1: line 5655
at <ScriptBlock>, <No file>: line 1
```

This can be caused when `DisabledComponents` is set to `0xffffffff` which is `-1`.  We don't recommend to set this value as it can cause a system startup delay:

_Additionally, system startup will be delayed for five seconds if IPv6 is disabled by incorrectly, setting the DisabledComponents registry setting to a value of 0xffffffff. The correct value should be 0xff. For more information, see Internet Protocol Version 6 (IPv6) Overview._

**Expected behavior**
We should not run into an exception. We should instead display a warning that the current value may cause issues.

**Script Output**
```
Exception setting "IPv6DisabledComponents": "Cannot convert value "0xffffffff" to type "System.Double". Error: "Input 
string was not in a correct format.""
At C:\Temp\HealthChecker.ps1:2947 char:5
+     $osInformation.NetworkInformation.IPv6DisabledComponents = Invoke ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], SetValueInvocationException
    + FullyQualifiedErrorId : ExceptionWhenSetting
 
[...]

PR #573, opened 2021-05-03
Check for misconfigured DisabledComponents -1 value

lusassl-msft:
**Issue:**
We now check if the `DisabledComponents` DWORD is set to `0xffffffff (-1)`. This setting causes a system startup delay of 5 seconds. For more information see issue #536 

**Fix:**
First we check for a value of `0xffffffff (-1)`, otherwise for not equal (ne) 255.

**Validation:**
Validated in lab


dpaulson45:
Waiting to determine if we need to add a way to check to see if the registry is set to the correct type as well.

dpaulson45 (approved):

PR #577, opened 2021-05-06
[New feature] Download Domains configuration check

lusassl-msft:
**Feature description**
We introduced the `Download Domains` feature with Exchange 2016 CU18 and Exchange 2019 CU7. The feature was introduced to address `CVE-2021-1730 Spoofing Vulnerability` and should be configured.

For more information, see the FAQ section: https://msrc.microsoft.com/update-guide/vulnerability/CVE-2021-1730

**Fix:**
We check if the `Download Domains` feature is enabled or not. We show a red warning if it's available but not configured (disabled).
If the feature is enabled, then we check the configuration.
We compare the `ExternalDownloadHostName` and `InternalDownloadHostName` with the OWA `ExternalUrl` and `InternalUrl` and show a red warning if they are configured to use the same namespace (like owa.contoso.com).

**Validation:**
Validated in lab. Possible Pester test case.

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
@@ -1495,6 +1495,64 @@
         }
     }
 
+    Function Test-DownloadDomainsConfiguration {
+        param(
+            [Parameter(Mandatory = $true)][object]$OwaVDirObject
+        )
+        Write-VerboseOutput("Calling: Test-DownloadDomainConfiguration")
+
+        <#
+        Unknown 0
+        Download Domains disabled 1
+        Download Domains enabled and configured as expected 2
+        Download Domains enabled and external download host name = internal/external owa url 4
+        Download Domains enabled but external download host name not set 8
+        Download Domains enabled and internal download host name = internal/external owa url 16
+        Download Domains enabled but internal download host name not set 32
+        #>
+
+        $downloadDomainsStatus = 0
+
+        try {
+            $downloadDomainsEnabled = (Get-OrganizationConfig -ErrorAction Stop).EnableDownloadDomains
+        } catch {
+            Write-VerboseOutput("Failed to run Get-OrganizationConfig. Download Domains result may not accurate")
+            Invoke-CatchActions
+            return $downloadDomainsStatus
+        }
+
+        if ($downloadDomainsEnabled) {
+            $downloadDomainsStatus += 2
+
+            if (![String]::IsNullOrEmpty($OwaVDirObject.ExternalDownloadHostName)) {
+
+                if (($OwaVDirObject.ExternalDownloadHostName -eq $OwaVDirObject.ExternalUrl.Host) -or
+                    ($OwaVDirObject.ExternalDownloadHostName -eq $OwaVDirObject.InternalUrl.Host)) {
+                    $downloadDomainsStatus += 4
+                }
+            } else {
+                Write-VerboseOutput("'ExternalDownloadHostName' is not configured")
+                $downloadDomainsStatus += 8
+            }
+
+            if (![String]::IsNullOrEmpty($OwaVDirObject.InternalDownloadHostName)) {
+
+                if (($OwaVDirObject.InternalDownloadHostName -eq $OwaVDirObject.ExternalUrl.Host) -or
+                    ($OwaVDirObject.InternalDownloadHostName -eq $OwaVDirObject.InternalUrl.Host)) {
+                    $downloadDomainsStatus += 16
+                }
+            } else {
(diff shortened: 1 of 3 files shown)


Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
```diff
@@ -1495,6 +1495,64 @@
         }
     }
 
+    Function Test-DownloadDomainsConfiguration {
```
dpaulson45, line 1498: Do not add this function here into the `Invoke-AnalyzerEngine.ps1` file. This function should only be analyzing the object.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
```diff
@@ -1495,6 +1495,64 @@
...
+        Download Domains enabled and external download host name = internal/external owa url 4
+        Download Domains enabled but external download host name not set 8
+        Download Domains enabled and internal download host name = internal/external owa url 16
+        Download Domains enabled but internal download host name not set 32
+        #>
+
+        $downloadDomainsStatus = 0
+
+        try {
+            $downloadDomainsEnabled = (Get-OrganizationConfig -ErrorAction Stop).EnableDownloadDomains
```
dpaulson45, line 1517: This is the main reason why this function shouldn't be in the `Invoke-AnalyzerEngine`, we shouldn't have any 'get' type commands that needs to be run on the customer's environment. 

We are already running `Get-OrganizationConfig` in `Get-ExchangeInformation`. Add two attributes to the `HealthChecker.ExchangeInformation` class. 

- GetOrganizationConfig //store the results from Get-OrganizationConfig
- EnableDownloadDomains //quick access to the the value that you are looking for

Update the code to look something like this

```
        try {
$organizationConfig =  (Get-OrganizationConfig -ErrorAction Stop)
            $exchangeInformation.GetOrganizationConfig = $organizationConfig 
        } catch {
            Write-Yellow "Failed to run Get-OrganizationConfig."
            Invoke-CatchActions
        }

if ($null -ne $organizationConfig) {
//save values
} else {
Write-ScriptDebug "MAPI HTTP Enabled and download domains enabled results not accurate"
}
```

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
```diff
@@ -1689,6 +1749,116 @@
         }
     }
 
+    #Description: Check for CVE-2021-1730 vulnerability
+    #Fix available for: Exchange 2016 CU18+, Exchange 2019 CU7+
+    #Fix: Configure Download Domains feature
+    #Workaround: N/A
+
+    if ((($exchangeInformation.BuildInformation.MajorVersion -eq [HealthChecker.ExchangeMajorVersion]::Exchange2016) -and
+            ($exchangeCU -ge [HealthChecker.ExchangeCULevel]::CU18)) -or
```
dpaulson45, line 1758: Looks like we are adding in the CU here because we are checking for the configuration that is only applicable if you are on the CU containing the fix.

At the very least, we should add in an `else` statement calling out in the debug log that we weren't on the correct CU to hit this area of code if we think the `CVE-2021-1730` in the other section being called out is enough there. Which i think it should be for the most part.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
dpaulson45: We can probably keep the function here for now. But we need to update it so it doesn't have the requirement of running any "get" commands. We should already have the data here to be analyzed.

dpaulson45 (changes requested):

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
lusassl-msft: Okay, should I create a project file for this new feature under: `DataCollection\ExchangeInformation` ?

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
lusassl-msft: Done

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
lusassl-msft: Done.

dpaulson45 (approved):

Issue #539, opened 2021-04-26
HealthChecker - Get organization hybrid configuration


dpaulson45:
We should collect information whether a server (which is part of an Exchange organization) has configured hybrid (by running HCW) or not. This would be helpful in developing further testing on this topic.

Migrated From: https://github.com/dpaulson45/HealthChecker/issues/471

lusassl-msft:
Closed for now. First implementation with PR #604.

PR #604, opened 2021-06-01
Check for Exchange Organization Hybrid Configuration

lusassl-msft:
**Description:**
This is the first draft of the Exchange Organization Hybrid Configuration implementation. We query the information via `Get-HybridConfiguration` and store the results in `public class ExchangeInformation` as `public object GetHybridConfiguration;`.
We only run the validation of these values when the server is no Edge Transport server and Exchange version is greater or equal Exchange2013. We simply skip the checks when no Hybrid Configuration is available. 

image

Work item #539

**Validation:**
Validated in Exchange 2016 Hybrid lab with Edge Transport server in place; possible Pester unit test case

### DIFF
--- a/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeInformation.ps1
+++ b/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeInformation.ps1
@@ -372,6 +372,12 @@
 
         if ($buildInformation.ServerRole -ne [HealthChecker.ExchangeServerRole]::Edge) {
             $exchangeInformation.ApplicationPools = Get-ExchangeAppPoolsInformation
+            try {
+                $exchangeInformation.GetHybridConfiguration = Get-HybridConfiguration -ErrorAction Stop
+            } catch {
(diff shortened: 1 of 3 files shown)


Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeInformation.ps1
```diff
@@ -372,6 +372,13 @@
 
         if ($buildInformation.ServerRole -ne [HealthChecker.ExchangeServerRole]::Edge) {
             $exchangeInformation.ApplicationPools = Get-ExchangeAppPoolsInformation
+            try {
+                $hybridConfiguration = Get-HybridConfiguration -ErrorAction Stop
```
dpaulson45, line 376: Let's just do the following, as there isn't a reason to store the results in a variable. 

```
$exchangeInformation.GetHybridConfiguration = Get-HybridConfiguration -ErrorAction Stop
```

dpaulson45 (changes requested):

dpaulson45 (approved):

PR #609, opened 2021-06-22
Canthv0 avtest

Canthv0:
**Issue:**
AV scanning can have a negative impact on Exchange.
Many cases can be caused by incorrectly configured scanning.

**Reason:**
Implemented script to place an EICAR file in all locations that are specified by PG that should be excluded from AV scanning.
Will then report if the file is removed.


bill-long:
Couple of minor formatting issues. `.build\CodeFormatter.ps1 -Save` will fix one of them. Other one will be a manual fix.

bill-long (approved):

PR #622, opened 2021-07-08
Update ExPerfAnalyzer.md

gabymuntean:
Please check the differences between the script versions.
Thank you!

**Issue:**

**Reason:**

**Fix:**
Short description of the fix

**Validation:**
Provide if applicable


lusassl-msft:
@gabymuntean I can't see any issue here. The link we use triggers the download of the latest release (which is currently version `21.06.29.0910`).

dpaulson45:
Closing PR, we are not linking the Markdown page to a raw source code page. Only release script files that is downloadable.

If you are noticing an issue, open an issue that provides what you are running into.

PR #638, opened 2021-07-23
Updated Visual C Redistributeable code check to be in Shared

### LINKED ISSUE #629: SetupAssist and HealthChecker look for different VC++ 2013 versions
SetupAssist:

image

HealthChecker:

image

Whatever we do here, we might want to change this check into a shared function so the code stays in sync between the two scripts.

dpaulson45:
Placed the logic for the check to see if each Visual C version is up to date within Shared. This way we can call it from each script it is located in the same way. 

Resolved #629


Shared/VisualCRedistributableVersionFunctions.ps1
```diff
@@ -0,0 +1,89 @@
...
+Function Get-VisualCRedistributable2013Information {
+    return [PSCustomObject]@{
+        VersionNumber = 201367256
+        DownloadUrl   = "https://support.microsoft.com/en-us/topic/update-for-visual-c-2013-redistributable-package-d8ccd6a5-4e26-c290-517b-8da6cfdf4f10"
+        DisplayName   = "Microsoft Visual C++ 2013*"
+    }
+}
+
+#Returns a 0 if no version is detected
+#Returns a 1 if version is detected, but the the version we wanted.
```
lusassl-msft, line 58: Typo: you forgot the (**not** the version we want)

lusassl-msft (dismissed):
Just the typo in the comment. Everything else looks fine.

bill-long (approved):

PR #645, opened 2021-07-27
New feature: ScriptUpdateOnly

lusassl-msft:
**Feature description:**
Many customers don't have an internet connectivity on their Exchange machines. These customers must manually download the latest version of the script from the internet by visiting `https://aka.ms/ExchangeHealthChecker`. 

The new feature `ScriptUpdateOnly` allows them to run the script without elevated permissions and without Exchange Management Tools installed on a machine with internet connectivity to perform an auto update only. Makes it easier for them to get the latest version of the script.

### DIFF
--- a/Diagnostics/HealthChecker/HealthChecker.ps1
+++ b/Diagnostics/HealthChecker/HealthChecker.ps1
@@ -59,6 +59,8 @@
     No version check is performed when this switch is used.
 .PARAMETER SaveDebugLog
     The debug log is kept even if the script is executed successfully.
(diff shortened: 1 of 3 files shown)


dpaulson45 (approved):
Not fully sure I understand the benefit this has as it is still additional work that the customer needs to do. However, I will approve the PR.

lusassl-msft:
Sure but it’s easier because customers don’t have to remember the link for the HC. Just run the script with the newly introduced parameter and that’s it. So it’s more an “comfort” feature. Not really needed but it makes life easier.

Issue #646, opened 2021-07-27
KB50004778 discovered but vulnerability still exists - Health Checker


tizianosartori:
**Provide Version Number**
21.07.23.1605

**Describe the issue**
HelthChecker script detect "Security Update for Exchange Server 2013 Cumulative Update 23 (KB5004778)" about Exchange Information but "Security Vulnerability: CVE-2021-34470" still displayed

**Expected behavior**
No "Security Vulnerability: CVE-2021-34470" display

dpaulson45:
This is caused by the schema update have not been applied yet. Please follow the steps in the blog article. 

https://techcommunity.microsoft.com/t5/exchange-team-blog/released-july-2021-exchange-server-security-updates/ba-p/2523421

dpaulson45:
@lusassl-msft might need to change the wording when the SU is installed, but not the Schema update.

lusassl-msft:
Yeah, we should do so. I'll take care of it.

tizianosartori:
> This is caused by the schema update have not been applied yet. Please follow the steps in the blog article.
> 
> https://techcommunity.microsoft.com/t5/exchange-team-blog/released-july-2021-exchange-server-security-updates/ba-p/2523421

Thanks, I haven't found this documentation yet
Now HealthChecker run without errors

PR #648, opened 2021-07-27
CVE-2021-34470 check improved

lusassl-msft:
**Issue:**
CVE-2021-34470 check improved to address #646.

**Reason:**
It drives confusion if we show that the system is vulnerable to CVE-2021-34470 even if the SU has been installed. 

**Fix:**
We now check if `KB5004778` or `KB5004779` or `KB5004780` are in place. If that's the case, we perform testing against the schema and check for the required schema change. If the schema change has not been performed, we highlight that `/PrepareSchema` is required to finalize the July 2021 SU installation to become fully protected against CVE-2021-34470.

**Validation:**
Lab

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
@@ -1957,25 +1957,31 @@ Function Invoke-AnalyzerEngine {
             ($exchangeCU -lt [HealthChecker.ExchangeCULevel]::CU10))) {
         Write-VerboseOutput("Testing CVE: CVE-2021-34470")
 
+        $displayWriteTypeColor = $null
+        if ($exchangeInformation.BuildInformation.MajorVersion -eq [HealthChecker.ExchangeMajorVersion]::Exchange2013) {
+            Test-VulnerabilitiesByBuildNumbersForDisplay -ExchangeBuildRevision $buildRevision -SecurityFixedBuilds "1497.23" -CVENames "CVE-2021-34470"
+        }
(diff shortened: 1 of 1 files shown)


Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
```diff
@@ -1966,7 +1966,13 @@ Function Invoke-AnalyzerEngine {
             $Script:AllVulnerabilitiesPassed = $false
             Write-VerboseOutput("Unable to query 'ms-Exch-Storage-Group' Exchange Schema class information")
         } elseif ($exchangeInformation.msExchStorageGroup.Properties.posssuperiors -eq "computer") {
-            $details = "{0}`r`n`t`tSee: https://portal.msrc.microsoft.com/en-us/security-guidance/advisory/{0} for more information." -f "CVE-2021-34470"
+            if ($exchangeInformation.BuildInformation.KBsInstalled -notlike "*KB5004778*" -and
```
dpaulson45, line 1969: This might work for now, but doesn't the KB get removed after the next security update is installed?

lusassl-msft: Nope we keep the history. We drop the history after a newer CU is installed but in this case the vulnerability is addressed too because we do / force the schema update.

lusassl-msft: Will modify the check to use the security fixed build number.

dpaulson45 (approved):

PR #660, opened 2021-07-29
Health Checker - Shared Files Updates

dpaulson45:
Did some major changes.
- Removed the `extern` directory from `DataCollection`
- Reworked all those files to be able to easily move to Shared if need be
- Updated some of the current `Shared` files the Health Checker uses to be using a better standard
- Placed all the required functions at the top of the source files where they are being used to dot load them
- Removed `Write-VerboseOutput` function and are now just using  `Write-Verbose`

@lusassl-msft to also test before approving


Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-AllNicInformation.ps1
```diff
@@ -0,0 +1,313 @@
...
+    )
+    begin {
+
+        Function Get-NicPnpCapabilitiesSetting {
+            [CmdletBinding()]
+            param(
+                [string]$NicAdapterComponentId
+            )
+            begin {
+                if ($NicAdapterComponentId -eq [string]::Empty) {
```
bill-long, line 22: It would be nice if we could move this validation into the param() declaration. For example, you could ValidateLength to make sure it has at least one character in it. For example:
image

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-ServerOperatingSystemVersion.ps1
```diff
@@ -0,0 +1,38 @@
...
+            "*Server 2012*" { $osReturnValue = "Windows2012" }
+            "*Server 2016*" { $osReturnValue = "Windows2016" }
+            "*Server 2019*" { $osReturnValue = "Windows2019" }
+            "Microsoft Windows Server Standard" { $osReturnValue = "WindowsCore" }
+            "Microsoft Windows Server Datacenter" { $osReturnValue = "WindowsCore" }
+            default { $osReturnValue = "Unknown" }
+        }
+    }
+    end {
+        Write-Verbose "Returned: '$OsReturnValue'"
```
bill-long, line 35: Case on line 35 doesn't match 36, 31, etc.

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-Smb1ServerSettings.ps1
```diff
@@ -0,0 +1,60 @@
...
+#>
+Function Get-Smb1ServerSettings {
+    [CmdletBinding()]
+    param(
+        [string]$ServerName = $env:COMPUTERNAME,
+        [scriptblock]$CatchActionFunction
+    )
+    begin {
+        Write-Verbose "Calling: $($MyInvocation.MyCommand)"
+        $smb1Status = 0
```
bill-long, line 22: We should either define our bit flags to make this readable (`$Failed = 1`, `$Installed = 2`, etc, then `$smb1Status += $Installed`), or not use a bitmask. Can we turn this into a PSCustomObject with a bunch of named booleans?

Shared/Confirm-ExchangeShell.ps1
```diff
@@ -1,111 +1,95 @@
...
-            & $CatchActionFunction
-        }
-
-        if (!$LoadExchangeShell) {
-            return [PSCustomObject]@{
-                ShellLoaded = $false
+    process {
+        try {
+            $currentErrors = $Error.Count
+            Get-ExchangeServer -ErrorAction Stop | Out-Null
```
bill-long, line 24: We should add -ResultSize 1 or use a cheaper/faster cmdlet to test, like Get-OrganizationConfig or Get-ADServerSettings.

lusassl-msft (changes requested):

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-AllNicInformation.ps1
dpaulson45: That is a good point, i will work on looking through all the current Health Checker parameters to see if there is a `throw` based off parameter information that we can replace with a `Validate` in the `param()`

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-Smb1ServerSettings.ps1
dpaulson45: @bill-long would it be better to create a `enum` class that we can instead of `$Failed = 1`, `$Installed = 2`. It would be something like this: 

```
enum Smb1Status
{
    Failed = 1
    Installed = 2
}
...

$smb1Status += [Smb1Status]::Failed 
```

Shared/Confirm-ExchangeShell.ps1
dpaulson45: Will Add `-ResultsSize 1` as we have run into issues with `Get-OrganizationConfig` not working due to RBAC for some users.

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-LmCompatibilityLevelInformation.ps1
```diff
@@ -1,12 +1,22 @@
...
 
     [HealthChecker.LmCompatibilityLevelInformation]$ServerLmCompatObject = New-Object -TypeName HealthChecker.LmCompatibilityLevelInformation
-    $ServerLmCompatObject.RegistryValue = Invoke-RegistryGetValue -RegistryHive "LocalMachine" -MachineName $Script:Server -SubKey "SYSTEM\CurrentControlSet\Control\Lsa" -GetValue "LmCompatibilityLevel" -CatchActionFunction ${Function:Invoke-CatchActions} -DefaultValue 3
+    $ServerLmCompatObject.RegistryValue = Get-RemoteRegistryValue -RegistryHive "LocalMachine" `
+        -MachineName $Script:Server `
+        -SubKey "SYSTEM\CurrentControlSet\Control\Lsa" `
+        -GetValue "LmCompatibilityLevel" `
+        -CatchActionFunction ${Function:Invoke-CatchActions}
+
+    if ($null -eq $ServerLmCompatObject.RegistryValue) {
```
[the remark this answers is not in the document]
dpaulson45: We actually need to save the results to it's own variable before setting it to `$ServerLmbCompatObject.RegistryValue`. That should solve this issue as the default value of `Get-RemoteRegistryValue` is to return a `$null` value.

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-ServerOperatingSystemVersion.ps1
dpaulson45: Too bad CodeFormatter does catch this.

Shared/Confirm-ExchangeShell.ps1
dpaulson45: @bill-long @lusassl-msft Let me know what you think. 

Actually, `Get-ExchangeServer` doesn't have an option for `-ResultSize`.

https://docs.microsoft.com/en-us/powershell/module/exchange/get-exchangeserver?view=exchange-ps

Because there has been a few cases where `Get-OrganizationConfig` isn't able to run by some people, don't want this function to fail because of that. So i think that should be out of the question. 

`Get-ExchangeServer` is the most common command to get information about the server that is typically run in scripts, or at least the main ones that i have written. We could do `Get-ADServerSettings` which does give us more RBAC options to help confirm that we are in an Exchange Management Shell, or we can include a parameter to pass an `-Identity` to `Get-ExchangeServer`


image

Shared/Confirm-ExchangeShell.ps1
```diff
@@ -1,111 +1,95 @@
...
-                Invoke-CatchActionErrorLoop -CurrentErrors $currentErrors
-            } catch {
-                Write-HostWriter("Failed to Load Exchange PowerShell Module...")
-                if ($null -ne $CatchActionFunction) {
-                    & $CatchActionFunction
+                    Write-Verbose "Imported Module. Trying Get-Exchange Server Again"
+                    Get-ExchangeServer -ErrorAction Stop | Out-Null
+                    $passed = $true
+                    Write-Verbose "Successfully loaded Exchange Management Shell"
+                    Invoke-CatchActionErrorLoop -CurrentErrors $currentErrors
```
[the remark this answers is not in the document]
dpaulson45: Fixed in 129e574c5b184a9c546248071cd75621c1366701

Shared/Confirm-ExchangeShell.ps1
```diff
@@ -1,111 +1,95 @@
...
+            Build       = ((Get-ItemProperty -Path $setupKey -Name "MsiBuildMajor" -ErrorAction SilentlyContinue).MsiBuildMajor)
+            Revision    = ((Get-ItemProperty -Path $setupKey -Name "MsiBuildMinor" -ErrorAction SilentlyContinue).MsiBuildMinor)
+            EdgeServer  = $passed -and (Test-Path $setupKey) -and (Test-Path $edgeTransportKey)
+            ToolsOnly   = $passed -and (Test-Path $setupKey) -and (!(Test-Path $edgeTransportKey)) -and `
+            ($null -eq (Get-ItemProperty -Path $setupKey -Name "Services" -ErrorAction SilentlyContinue))
+            RemoteShell = $passed -and (!(Test-Path $setupKey))
+        }
 
-    Invoke-CatchActionErrorLoop -CurrentErrors $currentErrors
+        Invoke-CatchActionErrorLoop -CurrentErrors $currentErrors
```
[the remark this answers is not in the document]
dpaulson45: Fixed in 129e574c5b184a9c546248071cd75621c1366701

Shared/Confirm-ExchangeShell.ps1
```diff
@@ -1,111 +1,95 @@
...
-        if (!$LoadExchangeShell) {
-            return [PSCustomObject]@{
-                ShellLoaded = $false
+    process {
+        try {
+            $currentErrors = $Error.Count
+            Get-ExchangeServer -ErrorAction Stop | Out-Null
+            Write-Verbose "Exchange PowerShell Module already loaded."
+            $passed = $true
+            Invoke-CatchActionErrorLoop -CurrentErrors $currentErrors
```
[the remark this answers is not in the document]
dpaulson45: Fixed in 129e574c5b184a9c546248071cd75621c1366701

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-LmCompatibilityLevelInformation.ps1
```diff
@@ -1,12 +1,22 @@
...
 
-    Write-VerboseOutput("Calling: Get-LmCompatibilityLevelInformation")
+    Write-Verbose "Calling: $($MyInvocation.MyCommand)"
 
     [HealthChecker.LmCompatibilityLevelInformation]$ServerLmCompatObject = New-Object -TypeName HealthChecker.LmCompatibilityLevelInformation
-    $ServerLmCompatObject.RegistryValue = Invoke-RegistryGetValue -RegistryHive "LocalMachine" -MachineName $Script:Server -SubKey "SYSTEM\CurrentControlSet\Control\Lsa" -GetValue "LmCompatibilityLevel" -CatchActionFunction ${Function:Invoke-CatchActions} -DefaultValue 3
+    $ServerLmCompatObject.RegistryValue = Get-RemoteRegistryValue -RegistryHive "LocalMachine" `
+        -MachineName $Script:Server `
+        -SubKey "SYSTEM\CurrentControlSet\Control\Lsa" `
+        -GetValue "LmCompatibilityLevel" `
```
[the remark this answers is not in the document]
dpaulson45: Added in f766d67b0a3b5fcbaf25cd997a7b9d60a701d529

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-LmCompatibilityLevelInformation.ps1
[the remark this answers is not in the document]
dpaulson45: Fixed in 3d77f2b7346e1a011753f5ba7a4e83a5630b68c0

Shared/Confirm-ExchangeShell.ps1
bill-long: If we're just trying to confirm the command is available, we test for:

`($null -ne (Get-Command Get-ExchangeServer -ErrorAction SilentlyContinue))`

If True, something has loaded a command called Get-ExchangeServer.

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-AllNicInformation.ps1
dpaulson45: Looks like this was the only place under Health Checker that was doing it. 

Fixed in 5c6998e65f876db823c540f08bce57e10f9e2471

Shared/Confirm-ExchangeShell.ps1
dpaulson45: Yes, we can do that. However, the secondary purpose of this functionality of the function is to determine if we are in a stale or idle session. If we are, when we run `Get-ExchangeServer` it ends up reconnecting the connection. It handles the errors that might occur and if we failed to re-connect it tries to create a new session similar as how we connect when launching the shortcut of EMS. That is why we can't just test to see if the cmdlet is there.

bill-long: We could explicitly check for an open session:

image
...
image

Though if we have to run Get-ExchangeServer (or some other random cmdlet) to re-establish it anyway, I guess there's not much benefit, and we should just stick with the random cmdlet call.

dpaulson45: It does reconnect most of the time just fine. However, it also throws errors while reconnecting. Within the Health Checker, there is built in logic to determine if there were errors that weren't handled. If we determine there was an error we didn't handle, at the end of the script we provide to email with the debug log to determine the cause. 

The other option like I provided was to pass a server name too. That will also limit the call and make it faster for large environments. The other option might be to do `Get-ADServerSettings`.

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-Smb1ServerSettings.ps1
bill-long: Actually the fact that you're using += is itself a problem. I even retyped it myself without thinking. This is a bitmask. We must use bitwise operators to avoid bugs. For example:

image

Oops, I accidentally set my favorite day to Saturday by adding Friday twice.

Personally, I would never use a bitmask. This article sums up my opinion nicely: https://dev.to/somedood/bitmasks-a-very-esoteric-and-impractical-way-of-managing-booleans-1hlf

You're really just creating a collection of booleans, and shoving them all into one value, making them very difficult to work with, when you could just do this:

```
[PSCustomObject]@{
  Failed = $false
  Installed = $true
  ...
}
```

Shared/Confirm-ExchangeShell.ps1
bill-long: Yeah, passing a single server name would be as good as -ResultSize 1.

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-Smb1ServerSettings.ps1
dpaulson45: Should be fully addressed in 9f1adc803ebfdaee432ad4363aa00d12591dc131

bill-long: That looks SO much better.

Shared/Confirm-ExchangeShell.ps1
dpaulson45: Fixed in 02dd28d07e442085124aff8d8009d802572c3f7a

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
```diff
@@ -1548,16 +1548,16 @@ Function Invoke-AnalyzerEngine {
...
     if ($osInformation.BuildInformation.MajorVersion -gt [HealthChecker.OSServerVersion]::Windows2012) {
         $displayValue = "False"
         $writeType = "Green"
 
-        if ($smb1Status -band 1) {
+        if (-not ($smb1Settings.SuccessfulGetInstall)) {
             $displayValue = "Failed to get install status"
             $writeType = "Yellow"
-        } elseif ($smb1Status -band 2) {
+        } elseif ($smb1Settings.Installed) {
```
lusassl-msft, line 1560: `$smb1Settings.Installed` doesn't work. Must be: `$smb1settings.WindowsFeature.Installed`

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
```diff
@@ -1572,10 +1572,10 @@ Function Invoke-AnalyzerEngine {
     $writeType = "Green"
     $displayValue = "True"
 
-    if ($smb1Status -band 8) {
+    if ($smb1Settings.SuccessfulGetBlocked) {
```
lusassl-msft, line 1575: Same here - we don't have SuccessfulGetBlocked available in this context.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
```diff
@@ -1572,10 +1572,10 @@ Function Invoke-AnalyzerEngine {
...
         $displayValue = "Failed to get block status"
         $writeType = "Yellow"
-    } elseif ($smb1Status -band 16) {
+    } elseif (-not($smb1Settings.IsBlocked)) {
```
lusassl-msft, line 1578: Doesn't work as well.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerEngine.ps1
```diff
@@ -1548,16 +1548,16 @@ Function Invoke-AnalyzerEngine {
...
     $additionalDisplayValue = [string]::Empty
-    $smb1Status = $osInformation.Smb1ServerSettings.Smb1Status
+    $smb1Settings = $osInformation.Smb1ServerSettings
 
     if ($osInformation.BuildInformation.MajorVersion -gt [HealthChecker.OSServerVersion]::Windows2012) {
         $displayValue = "False"
         $writeType = "Green"
 
-        if ($smb1Status -band 1) {
+        if (-not ($smb1Settings.SuccessfulGetInstall)) {
```
lusassl-msft, line 1557: `$smb1Settings.SuccessfulGetInstall` is not available.

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-OperatingSystemInformation.ps1
```diff
@@ -90,6 +115,6 @@ Function Get-OperatingSystemInformation {
     $osInformation.Smb1ServerSettings.WindowsFeature = $getSmb1ServerSettings.WindowsFeature
     $osInformation.Smb1ServerSettings.Smb1Status = $getSmb1ServerSettings.Smb1Status
```
lusassl-msft, line 116: This is no longer available because of the modified logic in `Get-Smb1ServerSettings`

lusassl-msft: `int Smb1Status` is also still present in `public class Smb1ServerSettings`

lusassl-msft (changes requested):

lusassl-msft (approved):

PR #663, opened 2021-07-30
New iteration of New-LoggerObject

dpaulson45:
this is to replace New-LoggerObject and make it easier to use within a script.


lusassl-msft (approved):
Validated in lab. Looks good so far.

PR #669, opened 2021-08-04
Test-ScriptVersion doesn't work if proxy with auth is in place

lusassl-msft:
**Issue:**
`Test-ScriptVersion` doesn't work if proxy with auth is in place. Customers with such a configuration are getting the following return: `The remote server returned an error: (407) Proxy Authentication Required.`

**Reason:**
This is caused by the fact that we do not provide credentials for authentication by default.

**Fix:**
Make use of `[System.Net.CredentialCache]::DefaultNetworkCredentials` to use the network credentials of the current security context (https://docs.microsoft.com/en-us/dotnet/api/system.net.credentialcache.defaultnetworkcredentials?view=net-5.0). 

**Validation:**
~~Please do not merge yet - will be validated together with affected customers (especially the `([System.Net.WebRequest]::GetSystemWebproxy()).IsBypassed("https://github.com") -eq $false` condition needs to be validated).~~
Validated with customer and in lab.


Shared/Test-ScriptVersion.ps1
```diff
@@ -90,6 +90,11 @@ function Test-ScriptVersion {
     try {
         $versionsUrl = "https://github.com/microsoft/CSS-Exchange/releases/latest/download/ScriptVersions.csv"
         [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+        if (([System.Net.WebRequest]::GetSystemWebproxy()).IsBypassed("https://github.com") -eq $false) {
+            $webClient = New-Object System.Net.WebClient
+            $webClient.Headers.Add("User-Agent", "PowerShell")
+            $webClient.Proxy.Credentials = [System.Net.CredentialCache]::DefaultNetworkCredentials
```
dpaulson45, line 96: We don't end up using `$webClient` when Invoking the web request. Does this do anything?

lusassl-msft: Yes, it affects the current PS session. There is no need to set this anywhere else.

lusassl-msft:
Validated with customer. PowerShell `Invoke-WebRequest` relies on the proxy settings set via Internet Explorer. I've adjusted the check and added a new function to validate if a proxy has been set or not.

Just set a non-existing proxy server - `Invoke-WebRequest` failed as expected in this case (proxy does not exist):
image

Without proxy server:
image

We then decide, based on the return, to use the network credential of the current security context or not for `System.Net.WebClient` class (only valid for the current PS session).

bill-long (approved):

PR #677, opened 2021-08-06
Fixes #676

### LINKED ISSUE #676: [Issue] Test-CVE-2021-33768 fails looking for Schema Master in non-English environment
From blog post comments:

```
Cannot index into a null array.
At C:\SETUP\CVE-2021-34470_Exchange\Test-CVE-2021-34470.ps1:55 char:92
+ $schemaMaster = (netdom query fsmo | Select-String "Schema master\s+(\S+)").Matches.Groups[ <<<< 1].Value
+ CategoryInfo : InvalidOperation: (1:Int32) [], ParentContainsErrorRecordException
+ FullyQualifiedErrorId : NullArray
```

Note this likely means HealthChecker role checks also fail (maybe silently?) since we use the same approach in both.

bill-long:


lusassl-msft (approved):

Security/src/Test-CVE-2021-34470.ps1
```diff
@@ -65,7 +65,7 @@ if ($ApplyFix) {
         $storageGroupSchemaEntry.Properties["possSuperiors"] | Out-File $OutputFile -Append
     }
 
-    $isSchemaAdmin = $null -ne (whoami /groups | sls "\\Schema Admins\s+Group")
+    $isSchemaAdmin = $null -ne (whoami /groups | sls "\\Schema Admins\s+")
```
dpaulson45, line 68: I am not 100% this works correctly like we think it does. I have seen a few issues with SetupAssist where they are part of the group vs sub groups vs directly added to it. 

Test by created a secondary group (ChildGroup) place your Admin into ChildGroup. Place ChildGroup as a member of "Schema Admins". 

If that doesn't work, should we address that now?

dpaulson45 (changes requested):

PR #688, opened 2021-08-09
Fixed issue with CodeFormatter script not calling out all the rules

### LINKED ISSUE #682: [Issue] Invoke-ScriptAnalyzer isn't catching all the rules like it use to
**Describe the issue**
Recently found that an alias was able to get by the build process. After looking into it, we found that it wasn't catching a few of the rules. 

**Expected behavior**
We should be using all the default rules by default, then excluding the rules that we want. 

**Additional context**

Removing this section of the `PSScriptAnalyzerSettings.psd1` allows all the rules to take effect.
```
    IncludeRules        = @(
        'PSPlaceOpenBrace',
        'PSPlaceCloseBrace',
        'PSUseConsistentWhitespace',
        'PSUseConsistentIndentation',
        'PSAlignAssignmentStatement',
        'PSUseCorrectCasing'
    )
```
Not fully sure why this was this way to begin with, as it was working initially. A possibility is that something was changed when merging HealthChecker or ExchangeLogCollector if they had these settings. Either way, need to revert this change and which requires going through all the error the script is now producing.

dpaulson45:
Removed the IncludeRules section of the `PSScriptAnalyzerSettings.psd1` file and addressed all the rules that were now being flagged for all the scripts

Resolved #682


.build/CodeFormatter.ps1
```diff
@@ -1,6 +1,6 @@
 ﻿# Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
-
+[Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseDeclaredVarsMoreThanAssignments', '', Justification = '$fileFailed is being used.')]
```
bill-long, line 3: This should be plural. $filesFailed

Diagnostics/ManagedAvailabilityTroubleshooter/ManagedAvailabilityTroubleshooter.ps1
```diff
@@ -553,7 +577,7 @@ function CheckifthiscanbeaknownissueusingMonitor {
     if ($Script:lastProbeerror) {
         if ($Script:lastProbeerror.Exception -like "*The underlying connection was closed*") {
             Write-Host -foreground yellow "This probe error message related to underlying connection closed has been seen when connection for loopback adapter has been blocked at lower level before reaching Exchange`n"
-            Write-Host -foreground yellow "You can check in IIS Default Web Site /Actions pane / Bindings that “All Unassigned” is used and this has not been changed to only allow specific IP.`n"
+            Write-Host -foreground yellow "You can check in IIS Default Web Site /Actions pane / Bindings that `“All Unassigned`” is used and this has not been changed to only allow specific IP.`n"
```
bill-long, line 580: Can we replace these curly quotes with normal quotes? Seems like an encoding issue waiting to happen.

bill-long (changes requested):

bill-long (approved):

bill-long (approved):

PR #689, opened 2021-08-11
Matt simple

Canthv0:
**Issue:**
Script failed to run after downloading.
Was generating large numbers of null exception errors even after correcting basic script failure.

**Reason:**
Correct errors to bring script back to function.

**Fix:**
Fixed issue with Parameters.
Added null check for modified properties.

**Validation:**
Ran script against customer provided dataset and received correct output.


bill-long (changes requested):
```
RuleName                             Severity ScriptName                   Line Message
--------                             -------- ----------                   ---- -------
PSUseDeclaredVarsMoreThanAssignments Warning  Get-SimpleAuditLogReport.ps1 68   The variable 'ResultSet' is assigned
                                                                                but never used.
```

Please run .build\CodeFormatter and fix any reported issues until clean.

Admin/Get-SimpleAuditLogReport.ps1
```diff
@@ -193,10 +191,7 @@ Process {
...
     # Add the object to the array to be returned
-    $ResultSet = $ResultSet + $Result
+    $Result
 }
 # Final steps
-End {
-    # Return the array set
-    Return $ResultSet
-}
+End {}
```
dpaulson45, line 197: should remove this if it is empty.

dpaulson45 (changes requested):

Admin/Get-SimpleAuditLogReport.ps1
Canthv0: <shrug> was left there for consistency with the Begin / Process / End methodology but doesn't matter either way so took it out.

Canthv0:
Everything fixed should be good to go.

bill-long (approved):

Admin/Get-SimpleAuditLogReport.ps1
dpaulson45: True, but there wasn't even a Process section in the script. Which i am surprised that we don't get a rule error on this being that we set the one parameter to be accepted from the pipeline.

dpaulson45 (approved):

PR #691, opened 2021-08-11
Wording fixed

lusassl-msft:
`Test-ScriptVersion` returns `$false` if the version is current and no script update is required or when we run into a state where no update can be performed (for example due to no internet access). 

**Reason:**
Due to the way how `Test-ScriptVersion` works, we cannot distinguish what the reason for a non-update state is. The wording was adjusted to highlight that no update was performed which might be by design (version current). 
Long term resolution would be to adjust `Test-ScriptVersion` to differentiate between a successful update, no update required or no update possible and return the appropriate value.


bill-long:
My original use of the result value from Test-ScriptVersion was just so the caller could say "Do I need to restart the script due to an update, or not?" I'm not sure if the caller needs any more information than that.

lusassl-msft:
Thanks Bill. Doesn’t have a high prio. Will think about it later. We should be good for the moment with the current approach.

PR #695, opened 2021-08-16
Update Health Checker

### LINKED ISSUE #537: HealthChecker - IIS Web App Settings
Had a customer having an app pool getting restarted frequently. This was caused by a IIS Web App Pool setting change. They added `Request Limit` value to their app pool. This is an accumulative number of all the requests that come into the app pool, so once the limit is reached, the app pool is recycled. 

**Describe Check Request**
In `Get-ExchangeAppPoolsInformation` we should look at also collecting more information like the following and storing it. 

```
&$appcmd list apppool MSExchangeMapiMailboxAppPool /text:*
```
Then determine how to extract the data for the settings we want to look at and display to the screen. 

Migrated From: https://github.com/dpaulson45/HealthChecker/issues/511

### LINKED ISSUE #692: Health Checker should check Recycling settings
Setting Recycling Settings on an IIS vdir will cause the app pool to recycle periodically, resulting in a short outage.

Health Checker should validate these settings to make sure we are not configured for self-inflicted outages.

dpaulson45:
List of updates performed:

- updated links to aka.ms links and removed some links to 3rd party locations that weren't really needed anymore.
- inactive component to red write type
- created a Write-Host override to better handle color options on machines
- added functionality for column formatting with colorization
- addressed Web App Pool restarts (Resolved #692 Resolved #537)


lusassl-msft (approved):
Validated in lab and compared with output of the previous version. Looks good so far.

PR #700, opened 2021-08-17
Update Emerging-Issues.md

JasonDebug:
Extended info for HMAC error in emerging issues.

**Issue:**
No need to recreate Auth cert if it's valid

**Reason:**
Additional info for emerging issue to avoid unnecessarily recreating an auth cert in an org

**Fix:**
Updated text.

**Validation:**
Recent cases.


Batre-MSFT:
Jason, it's called out on the [public document](https://docs.microsoft.com/en-us/exchange/troubleshoot/administration/cannot-access-owa-or-ecp-if-oauth-expired) to fix the issue

Issue #653, opened 2021-07-28
[Issue] Could not establish trust relationship for the SSL/TLS secure channel


tweekerz:
**Describe the issue**
 .\Test-AMSI.ps1 on Exchange 2016 CU21 but get below SSL/TLS connection error. Any idea? 

Test-AMSI : The underlying connection was closed: Could not establish trust relationship for the SSL/TLS secure channel.
At C:\Batch\AMSI\Test-AMSI.ps1:124 char:1
+ Test-AMSI
+ ~~~~~~~~~
+ CategoryInfo : NotSpecified: (:) [Write-Error], WriteErrorException
+ FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Test-AMSI

If you are using Microsoft Defender then AMSI may be disabled or you are using a AntiVirus Product that may not be AMSI capable (Please Check with your AntiVirus Provider for Exchange AMSI Support)

**Expected behavior**
Expect to be able to run even if the certificate is not trusted

**Script Suggestion**
if (-not ([System.Management.Automation.PSTypeName]'ServerCertificateValidationCallback').Type)
{
$certCallback = @"
    using System;
    using System.Net;
    using System.Net.Security;
    using System.Security.Cryptography.X509Certificates;
    public class ServerCertificateValidationCallback
    {
        public static void Ignore()
        {
            if(ServicePointManager.ServerCertificateValidationCallback ==null)
            {
                ServicePointManager.ServerCertificateValidationCallback += 
                    delegate
                    (
                        Object obj, 
                        X509Certificate certificate, 
                        X509Chain chain, 
                        SslPolicyErrors errors
                    )
                    {
                        return true;
                    };
            }
        }
    }
"@
    Add-Type $certCallback
 }
[ServerCertificateValidationCallback]::Ignore()

bill-long:
```
[Net.ServicePointManager]::ServerCertificateValidationCallback = {$true}
```

This simple approach _should_ work. Let me know if not.

PR #702, opened 2021-08-18
updates to AMSI script

### LINKED ISSUE #652: [Issue] add German localization for IIS HTTP 400 Error
**Describe the issue**
Running the Test-AMSI script on an Exchange server that has a different OS language than English (German in my case) will lead to the warning message that AMSI may be disabled, as the script checks for the English error message of the code 400 / Bad Request that is returned. The test is working fine despite of this, code 400 is returned, and the detection is logged. And a Microsoft 365 Defender alert was triggered, as we've onboarded the Exchange servers to Defender for Endpoint. Of course, you can change the script and add the error message in the OS language if you want it to display the correct result.

**Expected behavior**
A correct localized message

tweekerz:
Fixes #652 #653


lusassl-msft (approved):

lusassl-msft:
Looks like there is an formatting issue.

Admin/Test-AMSI.ps1
```diff
@@ -1,11 +1,9 @@
-﻿# Copyright (c) Microsoft Corporation.
-# Licensed under the MIT License.
```
bill-long, line 2: Compliance header was removed. It is required.

Admin/Test-AMSI.ps1
```diff
@@ -1,11 +1,9 @@
...
-
-[CmdletBinding(DefaultParameterSetName = "TestAMSI", HelpUri = "https://aka.ms/css-exchange")]
+﻿[CmdletBinding(DefaultParameterSetName = "TestAMSI", HelpUri = "https://aka.ms/css-exchange")]
 param(
     [Parameter(ParameterSetName = 'TestAMSI', Mandatory = $true, Position = 0)]
     [ValidateNotNullOrEmpty()]
     [string]$ExchangeServerFQDN,
+    [switch]$IgnoreSSL,
```
bill-long, line 6: Trailing whitespace on line 6.

bill-long (changes requested):

bill-long:
`.build\CodeFormatter.ps1 -Save` will fix both issues.

Admin/Test-AMSI.ps1
```diff
@@ -37,14 +35,29 @@ function Test-AMSI {
...
+        public class TrustAllCertsPolicy : ICertificatePolicy {
+            public bool CheckValidationResult(
+                ServicePoint srvPoint, X509Certificate certificate,
+                WebRequest request, int certificateProblem) {
+                return true;
+            }
+        }
+"@
+        Add-Type -TypeDefinition $IGSSL
+        [System.Net.ServicePointManager]::CertificatePolicy = New-Object TrustAllCertsPolicy
```
bill-long, line 51: CertificatePolicy is obsolete. You can delete all this code and add one line instead:

`[Net.ServicePointManager]::ServerCertificateValidationCallback = {$true}`

lusassl-msft: Oh, didn’t know that. @tweekerz https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager.certificatepolicy?view=netframework-4.8 for reference.

bill-long: Due to runspace errors when we try to use a PowerShell method for the callback, we'll have to set the ServerCertificateValidationCallback property from C# like so:

```
Add-Type @"
    using System;
    using System.Net;
    using System.Net.Security;
    using System.Security.Cryptography.X509Certificates;
    public class ServerCertificateValidationBehavior
    {
        public static void Ignore()
        {
            ServicePointManager.ServerCertificateValidationCallback += 
                delegate(Object obj, X509Certificate certificate, X509Chain chain, SslPolicyErrors errors)
                {
                    return true;
                };
        }
    }
"@

[ServerCertificateValidationBehavior]::Ignore()
```

We should also remove this when we're done by setting it back to null. This, we can do from PowerShell:

```
[Net.ServicePointManager]::ServerCertificateValidationCallback = $null
```

```

bill-long (approved):

bill-long:
One oddity to note is that the shell continues to ignore certificate errors for some time after that setting is used. We seem to be caching the Invoke-WebRequest session or something, and I haven't found a way to make the $null immediately take effect. It would be nice if we could eventually address this somehow.

image

PR #721, opened 2021-08-25
Force columns to fit within window width

bill-long:
Out-Columns will now auto-wrap, using spaces to break lines when possible.


bill-long:
Example output from merging this code with dpaul-devSetupAssist branch:

image

lusassl-msft (approved):

PR #725, opened 2021-08-26
Fix NIC settings fallback to WMI

lusassl-msft:
**Issue:**
Fallback to WMI to query NIC settings was not working properly in some scenarios. If WMI was used to query the NIC settings, certain information was missing (IP addresses ...). 

**Reason:**
There was no proper implementation to return the IP addresses in WMI scenario. We don't have all the information available which we get when using CIM with the latest cmdlets so this should be considered as a fallback just in case. 

**Fix:**
`Throw` when we don't get any result via `Get-NetworkConfiguration`. Added some logic to get the IP addresses, subnet mask and gateway. 

**Validation:**
Validated in lab and with affected customer. Please do not merge yet. Might require some more adjustments.


lusassl-msft:
Received feedback from customer that the WMI fallback works with the fix proposed.

dpaulson45:
Except the code is written correctly, took a closer look at it last night and understood what was going on a little bit more. Will push out additional changes here today.

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-AllNicInformation.ps1
```diff
@@ -240,6 +242,56 @@ Function Get-AllNicInformation {
                                 break
                             }
                         }
+
+                        $ipv4Gateway = @()
+                        $ipv6Gateway = @()
+                        foreach ($gateway in $adapterConfiguration.DefaultIPGateway) {
```
dpaulson45, line 252: Is it really possible to have multiple DefaultIPGateways for both ipv6 and ipv4 on the same nic?

lusassl-msft: Yes, I think it was possible. Will verify.

dpaulson45: Okay, well the new code that I added should handle this.

lusassl-msft: Validated in lab. We're good to merge this change into main.

dpaulson45 (approved):

PR #726, opened 2021-09-03
SourceSideValidations refactor

bill-long:
This is a large refactoring of SourceSideValidations. Highlights:

* Produces a single CSV file containing all the test results.
* Includes a new test for problematic characters in folder names.
* Includes a new test that checks for folders that are completely empty - that is, the folder itself has no items, all its children have no items, and all their children have no items, etc. These results are informational in nature, but they can be used to prune the hierarchy if desired.
* If you forget what the script said in the results summary, such as what commands to run, you can use the -SummarizePreviousResults switch to re-summarize the last run.
* Subsequent runs of the script do not overwrite the results. Instead, we rename the current file with its timestamp, and then we create a new ValidationResults.csv.
* A -SlowTraversal option is now available for environments where there are so many folders that the normal approach times out. This option is slower, but it can recover from a disconnected PowerShell session.
* Tests can be omitted by specifying the -Tests option. This is for scenarios where fixes have been performed and now you want to rerun only specific tests.


PublicFolders/src/SourceSideValidations/Get-FolderData.ps1
```diff
@@ -21,66 +30,99 @@ function Get-FolderData {
...
+                $folderData.ItemCounts = $itemCountResult.ItemCounts
+                foreach ($errorParam in $itemCountResult.Errors) {
+                    $errorResult = New-TestResult @errorParam
+                    $folderData.Errors.Add($errorResult)
                 }
             }
         }
+
+        $folderData.ItemCounts | ForEach-Object { $folderData.ItemCountDictionary[$_.EntryId] = $_.ItemCount }
+        $folderData.ItemCounts | Export-Csv $PSScriptRoot\ItemCounts.csv
```
dpaulson45, line 125: Same thing here, we are importing and exporting the same data.

PublicFolders/src/SourceSideValidations/Get-IpmSubtree.ps1
```diff
@@ -6,57 +6,117 @@ function Get-IpmSubtree {
...
+                    DumpsterEntryId   = if ($Folder.DumpsterEntryId) { $Folder.DumpsterEntryId.ToString() } else { $null }
+                    FolderSize        = $Folder.FolderSize
+                    HasSubfolders     = $Folder.HasSubfolders
+                    ContentMailbox    = $Folder.ContentMailboxName
+                    MailEnabled       = $Folder.MailEnabled
+                    MailRecipientGuid = $Folder.MailRecipientGuid
+                }
+
+                $null = $ipmSubtree.Add($result)
+
```
dpaulson45, line 94: `$impSubtree` isn't within the script block is this a mistake? 

If it is outside of the scope of the script block, might be best to use `$Script:impSubtree` to better declare it?

PublicFolders/src/SourceSideValidations/Get-IpmSubtree.ps1
```diff
@@ -6,57 +6,117 @@ function Get-IpmSubtree {
...
+                    ParentEntryId     = $Folder.ParentFolder.ToString()
+                    DumpsterEntryId   = if ($Folder.DumpsterEntryId) { $Folder.DumpsterEntryId.ToString() } else { $null }
+                    FolderSize        = $Folder.FolderSize
+                    HasSubfolders     = $Folder.HasSubfolders
+                    ContentMailbox    = $Folder.ContentMailboxName
+                    MailEnabled       = $Folder.MailEnabled
+                    MailRecipientGuid = $Folder.MailRecipientGuid
+                }
+
+                $null = $ipmSubtree.Add($result)
```
dpaulson45, line 93: Why are we doing `$null = something`? Should is this the best way to do it vs a pipe to `Out-Null`?

PublicFolders/src/SourceSideValidations/Get-ItemCountJob.ps1
```diff
@@ -0,0 +1,93 @@
...
+
+            $maxRetries = 5
+            for ($retryCount = 1; $retryCount -le $maxRetries; $retryCount++) {
+                try {
+                    $stats = Get-PublicFolderStatistics $folder.EntryId | Select-Object EntryId, ItemCount
+                    [void]$itemCounts.Add($stats)
+                    break
+                } catch {
+                    # Only retry Kerberos errors
+                    if ($_.ToString().Contains("Kerberos")) {
```
dpaulson45, line 54: Is "Kerberos" only going to be in English?

PublicFolders/src/SourceSideValidations/Get-ItemCountJob.ps1
```diff
@@ -0,0 +1,93 @@
...
+                            Write-Progress @progressParams -Status "Retry $retryCount of $maxRetries. Error: $($_.Message)"
+                            Start-Sleep -Seconds 5
+                            $remainingMilliseconds = $retryDelay.TotalMilliseconds - $sw.ElapsedMilliseconds
+                            if ($remainingMilliseconds -lt 0) { $remainingMilliseconds = 0 }
+                            Write-Progress @progressParams -Status "Retry $retryCount of $maxRetries. Will retry in $([TimeSpan]::FromMilliseconds($remainingMilliseconds))"
+                            Start-Sleep -Seconds 5
+                        }
+
+                        Get-PSSession | Remove-PSSession
+                        Import-PSSession (New-PSSession -ConfigurationName Microsoft.Exchange -ConnectionUri "http://$Server/powershell" -Authentication Kerberos) -AllowClobber | Out-Null
```
dpaulson45, line 66: Not doing the same as the beginning of the function. We aren't doing `-AllowClobber` in the first attempt. Unless that is on purpose, but this is the end of the function so not sure why it would be different here.

PublicFolders/src/SourceSideValidations/Tests/Name/Test-FolderName.ps1
```diff
@@ -1,12 +1,7 @@
 ﻿# Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
```
dpaulson45, line 3: missing dot loading of `New-TestResult`

dpaulson45 (changes requested):

PublicFolders/src/SourceSideValidations/Get-IpmSubtree.ps1
bill-long: It's in the begin block, so not a mistake. Everything within begin is accessible from process. No need for a qualifier.

PublicFolders/src/SourceSideValidations/Get-IpmSubtree.ps1
bill-long: Actuall $null = is faster and [void] is even faster. Out-Null is the slowest approach:

https://docs.microsoft.com/en-us/powershell/scripting/dev-cross-plat/performance/script-authoring-considerations

I'm gradually migrating to [void] even though I don't like how it looks. I went ahead and updated it in this location.

PublicFolders/src/SourceSideValidations/Get-ItemCountJob.ps1
bill-long: I have no idea. This code path has only been used in one specific environment. We'll have to cross that bridge when we get another case where -SlowTraversal is needed.

PublicFolders/src/SourceSideValidations/Get-FolderData.ps1
```diff
@@ -21,66 +30,99 @@ function Get-FolderData {
...
-                if ($null -ne $job.ItemCounts) {
-                    $folderData.ItemCounts = $job.ItemCounts
-                    $folderData.ItemCounts | Export-Csv $PSScriptRoot\ItemCounts.csv
-                }
+            if ($null -ne $job.ItemCounts) {
+                $folderData.ItemCounts = $job.ItemCounts
             }
         }
 
+        $folderData.IpmSubtree | Export-Csv $PSScriptRoot\IpmSubtree.csv
```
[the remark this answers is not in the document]
bill-long: Fixed.

PublicFolders/src/SourceSideValidations/Get-FolderData.ps1
bill-long: Fixed.

PublicFolders/src/SourceSideValidations/Get-ItemCountJob.ps1
bill-long: Fixed.

PublicFolders/src/SourceSideValidations/Tests/Name/Test-FolderName.ps1
bill-long: Fixed.

bill-long:
Added the docs to this PR.

PublicFolders/src/SourceSideValidations/Get-ItemCountJob.ps1
dpaulson45: sounds good. Just figured i should bring that up as we have seen that with Rob's script.

dpaulson45 (approved):

PR #730, opened 2021-09-06
Tabular results for SSV

bill-long:
SourceSideValidations now produces results as a table. Old output:

image

New output:

image


dpaulson45 (approved):

bill-long:
I disabled PSUseSingularNouns for now, because it is flagging dozens of nouns in other scripts that it didn't care about in 1.19.

bill-long:
Something wonky is going on in the case where it updates PSScriptAnalyzer to 1.20. It stops recognizing the PSScriptAnalyzer commands even though the module is loaded:

image

I tried removing the module and importing again, but that didn't fix it. Had to launch a new PowerShell instance and then it was fine.

I can't figure out how to fix this right now, so if this happens to you, just launch a new pwsh.

dpaulson45 (approved):

PR #742, opened 2021-09-24
Add SSV template

bill-long:


PublicFolders/SourceSideValidations/SourceSideValidations.ps1
```diff
@@ -45,165 +45,170 @@ param (
...
+        $errorColor = $configuredErrorColor
+    }
 
-$warningColor = "Yellow"
-$configuredWarningColor = (Get-Host).PrivateData.WarningForegroundColor
-if ($configuredWarningColor -is [ConsoleColor]) {
-    $warningColor = $configuredWarningColor
-}
+    $warningColor = "Yellow"
+    $configuredWarningColor = (Get-Host).PrivateData.WarningForegroundColor
```
dpaulson45, line 64: In stead of using this.... shouldn't we use the Write-Host from the shared location that does handle this?

dpaulson45 (changes requested):

PublicFolders/SourceSideValidations/SourceSideValidations.ps1
bill-long: Those colors get passed to Out-Columns. I never pass them to Write-Host.

dpaulson45 (approved):

PR #786, opened 2021-11-17
First integration of an AMSI check

lusassl-msft:
**Description:**
This PR implements a check regarding the Exchange Server AMSI Integration. This first attempt performs testings by retrieving overrides via `Get-SettingOverride`. We show a yellow warning if AMSI is disabled on server or organizational level. We assume that Exchange Server AMSI Integration is enabled and working if no override exists or if it's set to `Enabled=True`. 

I think there is more room for optimizations but I think it's a good start to build an useful check.


lusassl-msft:
@tweekerz feel free to have a look and comment on this (since you wrote the AMSI blog post: https://techcommunity.microsoft.com/t5/exchange-team-blog/more-about-amsi-integration-with-exchange-server/ba-p/2572371)

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
```diff
@@ -0,0 +1,57 @@
...
+
+. $PSScriptRoot\..\..\Helpers\Invoke-CatchActions.ps1
+Function Get-ExchangeAMSIConfigurationState {
+
+    begin {
+        Write-Verbose "Calling: $($MyInvocation.MyCommand)"
+    } process {
+        try {
+            Write-Verbose "Trying to query AMSI configuration state"
+            $amsiConfiguration = Get-SettingOverride | Where-Object { ($_.ComponentName -eq "Cafe") -and ($_.SectionName -eq "HttpRequestFiltering") } -ErrorAction Stop
```
dpaulson45, line 14: Need to move the `-ErrorAction Stop` to before the pipe and have it be with `Get-SettingOverride`

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
```diff
@@ -0,0 +1,57 @@
...
+
+            if (($null -ne $amsiConfiguration) -and
+                ($amsiConfiguration.Count -eq 1)) {
+                Write-Verbose "Setting override detected for AMSI configuration"
+                $amsiConfigurationQuerySuccessful = $true
```
dpaulson45, line 19: Move `$amsiConfigurationQuerySuccessful = $true` to right after `Get-SettingOverride`, no need to have it set to `true` in all the if statements

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
```diff
@@ -0,0 +1,57 @@
...
+                Switch ($amsiConfiguration.Parameters.Split("=")[1]) {
+                    ("False") { $amsiState = $false }
+                    ("True") { $amsiState = $true }
+                    Default { $amsiState = "Unknown" }
+                }
+
+                if ($null -eq $amsiConfiguration.Server) {
+                    $amsiOrgWideSetting = $true
```
dpaulson45, line 27: remove the if statement and do the following instead:

```
$amsiOrgWideSetting = $null -eq $amsiConfiguration.Server
```

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
```diff
@@ -0,0 +1,57 @@
...
+                } else {
+                    $amsiOrgWideSetting = $false
+                }
+            } elseif (($null -ne $amsiConfiguration) -and
+                ($amsiConfiguration.Count -gt 1)) {
```
dpaulson45, line 32: How do we get multiple overrides? How do we determine which one is active in Exchange, might need to change this logic

bill-long:
There are some test failures.

```
Starting discovery in 1 files.
Discovery found 31 tests in 152ms.
Running tests.
##[error] [-] Testing Health Checker by Mock Data Imports.Basic Exchange 2019 CU11 Testing HyperV.Display Results - Security Settings 114ms (111ms|3ms)
##[error]  Expected 71, but got 73.
##[error]  at $Script:ActiveGrouping.Count | Should -Be 71, D:\a\1\s\Diagnostics\HealthChecker\Tests\HealthChecker.Tests.ps1:127
##[error]  at <ScriptBlock>, D:\a\1\s\Diagnostics\HealthChecker\Tests\HealthChecker.Tests.ps1:127
##[error] [-] Testing Health Checker by Mock Data Imports.Mocked Calls.Testing Standard Mock Calls 830ms (829ms|1ms)
##[error]  Expected Invoke-CatchActions to be called 1 times exactly but was called 2 times
##[error]  at Should -Invoke @__params__p, C:\Program Files\WindowsPowerShell\Modules\Pester\5.3.1\Pester.psm1:14586
Tests completed in 16.49s
Tests Passed: 29, Failed: 2, Skipped: 0 NotRun: 0
```

lusassl-msft:
Will make some more adjustments here by adding Pester unit test.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+. $PSScriptRoot\..\..\Helpers\Invoke-CatchActions.ps1
+Function Get-ExchangeAMSIConfigurationState {
+
+    begin {
+        Write-Verbose "Calling: $($MyInvocation.MyCommand)"
```
dpaulson45, line 10: In here we should have the default values set for, and remove the additional locations in the catch or where we set them multiple times to avoid additional lines of code

```
$amsiState
$amsiOrgWideSetting
$amsiConfigurationQuerySuccessful
```

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeInformation.ps1
```diff
@@ -420,6 +421,12 @@ Function Get-ExchangeInformation {
             Write-Verbose "MAPI HTTP Enabled and Download Domains Enabled results not accurate"
         }
 
+        if ($OSMajorVersion -ge [HealthChecker.OSServerVersion]::Windows2016) {
```
dpaulson45, line 424: Maybe we should add a check for Edge here as well?

Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityAMSIConfigState.ps1
```diff
@@ -0,0 +1,71 @@
...
+
+        [Parameter(Mandatory = $true)]
+        [object]$HealthServerObject,
+
+        [Parameter(Mandatory = $true)]
+        [object]$DisplayGroupingKey
+    )
+
+    Write-Verbose "Calling: $($MyInvocation.MyCommand)"
+    $amsiInformation = $HealthServerObject.ExchangeInformation.AMSIConfiguration
```
dpaulson45, line 18: This should move inside the if statement of taking the action. 

Also add `$amsiWriteType = "Yellow"` here and remove the additional lines of where it is set to Yellow.

dpaulson45 (changes requested):

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
lusassl-msft: Done.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
lusassl-msft: It's possible to configure multiple overrides by using different IDs but the same component / section combination. Not sure how Exchange handles this case but it's possible to create these conflicting configurations (validated in lab). I have no idea yet how to handle these cases at best so the current idea is to write a warning that we cannot determine if the configuration is accurate or not. But agree, we should re-think this part and find a better solution in the near future.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeInformation.ps1
lusassl-msft: Yeah, good point. Idea was to filter only on OS version and to do the rest of the Exchange role / CU logic within the `Get-AMSIConfigurationState` function. But good point since I'm not sure if the `Get-SettingOverride` cmdlet is available on Edge Transport or not which may cause FP if it's not available.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
lusassl-msft: Done. Will be addressed with the next commit.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
lusassl-msft: Done. Will be addressed with the next commit.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
lusassl-msft: Done. Will be addressed with the next commit.

Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityAMSIConfigState.ps1
lusassl-msft: Done. Will be addressed with the next commit.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeInformation.ps1
lusassl-msft: Done. Added check for Edge transport server role. Will keep the check for Edge transport server role within the `Invoke-AnalyzerSecurityAMSIConfigState` function as well.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
lusassl-msft: I have no good idea how to better handle this case. It's possible to have multiple and conflicting settings:
image
image

I don't think that we see this scenario that much and so would like to start with this simple logic to call out that we've detected multiple settings and so cannot clearly determine the state of the AMSI Exchange integration:
image

What do you think?

lusassl-msft: Should add this to the AMSI check description with a few sentence about how to check and delete conflicting settings.

dpaulson45: I think for this, we need to just return a list of all the setting overrides and loop through them and display them all with the Health Checker. Still call out the Multiple overrides detected, but just display it the critical information that we want for each setting.

lusassl-msft: Yeah, might confuse the user but should be no problem to add the information.

lusassl-msft: Re-worked the multiple override detection / output. It now looks like this:
image

lusassl-msft: Added with latest commit. Resolve this conversation now. Please re-open if something else should be added.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
```diff
@@ -0,0 +1,88 @@
...
+                $amsiState = Get-AMSIStatusFlag -AMSIParameters $amsiConfiguration.Parameters -ErrorAction Stop
+                $amsiOrgWideSetting = ($null -eq $amsiConfiguration.Server)
+            } elseif (($null -ne $amsiConfiguration) -and
+                ($amsiConfiguration.Count -gt 1)) {
+                Write-Verbose "$($amsiConfiguration.Count) overrides for the same component and section detected"
+                $amsiMultiConfigObject = @()
+                foreach ($amsiConfig in $amsiConfiguration) {
+                    $amsiState = Get-AMSIStatusFlag -AMSIParameters $amsiConfig.Parameters -ErrorAction Stop
+                    $amsiOrgWideSetting = ($null -eq $amsiConfig.Server)
+                    $amsiConfigTempCustomObject = [PSCustomObject]@{
```
lusassl-msft, line 51: @dpaulson45 I think about to put the custom object creation into another method to avoid this duplicate code. Make sense? What do you think?

dpaulson45: Yeah, this is going to need to change. I don't think it needs to be another method though.

This should be easy to do all within here.

lusassl-msft: Okay, going to improve this part of the code

lusassl-msft: Reduced the complexity of the check. We return a custom object with two properties if no setting override was detected. This is all we need to perform the output.

Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityAMSIConfigState.ps1
```diff
@@ -0,0 +1,90 @@
...
+                $additionalAMSIDisplayValue += "`r`n`t`t`tEnabled: {0}" -f $amsi.Enabled
+                if ($amsi.OrgWideSetting) {
+                    $additionalAMSIDisplayValue += "`r`n`t`t`tSetting applies to all servers of the organization"
+                } else {
+                    $additionalAMSIDisplayValue += "`r`n`t`t`tSetting applies to the following server(s) of the organization:"
+                    foreach ($server in $amsi.Server) {
+                        $additionalAMSIDisplayValue += "`r`n`t`t`t{0}" -f $server
+                    }
+                }
+            }
```
dpaulson45, line 72: We should maybe add these lines as well to the multiple overrides detected

```
                $additionalAMSIDisplayValue += "`r`n`t`tThis may pose a security risk to your servers"
                $additionalAMSIDisplayValue += "`r`n`t`tMore Information: https://aka.ms/AMSIExchange"
````

File: Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1 [hunk 10]
```diff
@@ -0,0 +1,76 @@
...
+            )
+
+            try {
+                Switch ($AMSIParameters.Split("=")[1]) {
+                    ("False") { return $false }
+                    ("True") { return $true }
+                    Default { return $null }
+                }
+            } catch {
+                throw
```
dpaulson45, line 25: Add a `Write-Verbose` in here to help determine where we are at if we do `throw` out of this. Maybe even add `$AMSIParameters` to the wording

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
```diff
@@ -0,0 +1,76 @@
...
+        try {
+            Write-Verbose "Trying to query AMSI configuration state"
+            $amsiConfiguration = Get-SettingOverride -ErrorAction Stop | Where-Object { ($_.ComponentName -eq "Cafe") -and ($_.SectionName -eq "HttpRequestFiltering") }
+            $amsiConfigurationQuerySuccessful = $true
+
+            if ($null -ne $amsiConfiguration) {
+                Write-Verbose "$($amsiConfiguration.Count) override(s) detected for AMSI configuration"
+                $amsiMultiConfigObject = @()
+                foreach ($amsiConfig in $amsiConfiguration) {
+                    $amsiState = Get-AMSIStatusFlag -AMSIParameters $amsiConfig.Parameters -ErrorAction Stop
```
dpaulson45, line 43: Since this has a possible `throw` in it and we are calling a few times. Maybe we should do a `try` `catch` here to handle this. That way we don't fail on all the possible results here.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
```diff
@@ -0,0 +1,76 @@
...
+            } else {
+                Write-Verbose "No setting override found that overrides AMSI configuration"
+                $amsiState = $true
+            }
+        } catch {
+            Write-Verbose "Unable to query AMSI configuration state"
+            Invoke-CatchActions
+        }
+    } end {
+        if ($amsiMultiConfigObject) {
```
dpaulson45, line 67: @bill-long  is there a better way to do this?

Diagnostics/HealthChecker/Tests/HealthChecker.Tests.ps1
```diff
@@ -239,6 +240,7 @@ Describe "Testing Health Checker by Mock Data Imports" {
             Assert-MockCalled Get-OrganizationConfig -Exactly 1
             Assert-MockCalled Get-HybridConfiguration -Exactly 1
             Assert-MockCalled Get-Service -Exactly 1
+            Assert-MockCalled Get-SettingOverride -Exactly 1
             Assert-MockCalled Get-ServerComponentState -Exactly 1
             Assert-MockCalled Test-ServiceHealth -Exactly 1
         }
```
dpaulson45, line 246: Also add actual testing results besides that we just called it. Maybe add it to "Checking Scenarios 1"

dpaulson45 (changes requested):

Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityAMSIConfigState.ps1
lusassl-msft: Fixed with the next commit.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
lusassl-msft: We now log the following before throwing an exception:

`Write-Verbose "Ran into an issue when calling Split method. Parameters passed: $AMSIParameters"`

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
lusassl-msft: Good finding. Wrapped the `Get-AmsiStatusFlag` into an `try` `catch` block.
We do the following within the `catch`:
```
Write-Verbose "Unable to process: $($amsiConfig.Parameters) to determine status flags"
$amsiState = "Unknown"
Invoke-CatchActions
```

Need to set the `$amsiState` to `Unknown` again to make sure that a value which was (possibly) set in a previous call (if the exception happens between the processing of multiple objects) is overwritten.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAMSIConfigurationState.ps1
bill-long: @dpaulson45 I don't see a better way to do it without null coalescing, which is only available in PS 7+.

Diagnostics/HealthChecker/Tests/HealthChecker.Tests.ps1
lusassl-msft: Added `AMSI Override` and `AMSI Enabled` test to `Checking Scenarios 1`. Please have a look.

docs/Diagnostics/HealthChecker/AMSIIntegration.md
```diff
@@ -0,0 +1,30 @@
+---
+title: AMSI Check
```
dpaulson45, line 2: Going to need to change this page now with the new github pages formatting

dpaulson45 (changes requested):

docs/Diagnostics/HealthChecker/AMSIIntegration.md
lusassl-msft: Done.

dpaulson45 (approved):

Issue #547, opened 2021-04-26
Migrate HealthChecker Docs to Here


dpaulson45:
**Describe The Request**
Need to migrate the current work we have in HealthChecker docs dev branch to here in to GitHub Pages.

PR #794, opened 2021-11-25
HealthChecker docs migrated from old repo

lusassl-msft:
- Migrated the Health Checker check description from the old repository
- Added two new check descriptions: EEMS and Download Domains


lusassl-msft:
#547

lusassl-msft:
@bill-long please have a look at the `mkdocs.yml`. I've configured the edit function to use the main branch instead of master (which is the default and doesn't work for the CSS-Exchange repository).

See: https://www.mkdocs.org/user-guide/configuration/

bill-long (approved):

PR #797, opened 2021-11-26
Make LoadBalancing feature language-independent

### LINKED ISSUE #581: [Issue] - Health Checker - Load Balancing Report doesn't work on Non-US systems
Health Checker Load Balancing doesn't work on Non-US systems. We use a static approach (counter name in English language) to query the required counters. Need to investigate if there is another way to query the required counters without relying on the name (like: https://powershellmagazine.com/2013/07/19/querying-performance-counters-from-powershell/).

lusassl-msft:
**Issue:**
The LoadBalancing report feature does not work on non-English OS.
Performance counters are language dependent. This update tries to address this problem by introducing the `Get-LocalizedPerformanceCounterNames` helper function which can map the English performance counter names to the localized name by using performance counter ids.

Resolve #581 

**Fix:**
We query the performance counter id and then mapping the id to the localized counter name which we finally use to run the `Get-Counter` cmdlet

**Validation:**
Validated on non-English lab system.
Validated also that the script works as expected on a pure English OS.


dpaulson45:
Need to think about this some more and see if there is a better way to handle this

lusassl-msft:
Not sure if there is a better way to do this via PowerShell as this is the documented way to retrieve these data via Registry Function (https://docs.microsoft.com/en-us/windows/win32/perfctrs/using-the-registry-functions-to-consume-counter-data).

Performance Data Helper (PDH) functions also use this method: https://docs.microsoft.com/en-us/windows/win32/perfctrs/using-the-pdh-functions-to-consume-counter-data

dpaulson45:
closing PR

PR #815, opened 2021-12-17
New check to validate if build was replaced by a newer one 

dpaulson45:
**Description:**
Added some additional logic in the Analyzer Engine to determine if the Exchange build number is running into a known issue and if we have a fix for that build. 

image

Replacing PR #812

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerKnownBuildIssues.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerKnownBuildIssues.ps1
@@ -1,0 +1,125 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+. $PSScriptRoot\Add-AnalyzedResultInformation.ps1
+. $PSScriptRoot\..\Helpers\Invoke-CatchActions.ps1
+
+Function Invoke-AnalyzerKnownBuildIssues {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
+        [ref]$AnalyzeResults,
+
+        [Parameter(Mandatory = $true)]
+        [string]$CurrentBuild,
+
+        [Parameter(Mandatory = $true)]
+        [object]$DisplayGroupingKey
+    )
+
+    Write-Verbose "Calling: $($MyInvocation.MyCommand)"
+
+    Function GetVersionFromString {
+        param(
+            [object]$VersionString
+        )
+        try {
+            return New-Object System.Version $VersionString -ErrorAction Stop
+        } catch {
+            Write-Verbose "Failed to convert '$VersionString' in $($MyInvocation.MyCommand)"
+            Invoke-CatchActions
+        }
+    }
+
+    Function GetKnownIssueInformation {
+        param(
+            [string]$Name,
+            [string]$Url
+        )
+
+        return [PSCustomObject]@{
+            Name = $Name
+            Url  = $Url
+        }
+    }
+
+    Function TestForKnownBuildIssues {
+        param(
+            [version]$CurrentVersion,
+            [hashtable]$KnownBuildIssuesToFixes,
+            [object]$InformationUrl
+        )
+        Write-Verbose "Calling: $($MyInvocation.MyCommand)"
+        Write-Verbose "Testing CurrentVersion $CurrentVersion"
+
+        if ($null -eq $Script:CachedKnownIssues) {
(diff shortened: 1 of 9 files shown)


Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerKnownBuildIssues.ps1
```diff
@@ -0,0 +1,125 @@
...
+            Write-Verbose "Testing Known Issue Build $knownIssue"
+
+            if ($null -ne $knownIssue -and
+                $CurrentVersion.Minor -eq $knownIssue.Minor) {
+
+                $resolvedBuild = GetVersionFromString $KnownBuildIssuesToFixes[$key]
+                Write-Verbose "Testing against possible resolved build number $resolvedBuild"
+                if (($null -eq $KnownBuildIssuesToFixes[$key] -and
+                    (-not($Script:CachedKnownIssues.Contains($InformationUrl)))) -or
+                    ($CurrentVersion -ge $knownIssue -and
```
lusassl-msft, line 70: I would add some more curved brackets here to make it easier to read:
```
if ((($null -eq $KnownBuildIssuesToFixes[$key]) -and
                    (-not($Script:CachedKnownIssues.Contains($InformationUrl)))) -or
                    (($CurrentVersion -ge $knownIssue) -and
                     ($CurrentVersion -lt $resolvedBuild))) {
```

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerKnownBuildIssues.ps1
```diff
@@ -0,0 +1,125 @@
...
+
+        if ($null -eq $Script:CachedKnownIssues) {
+            $Script:CachedKnownIssues = @()
+        }
+
+        foreach ($key in $KnownBuildIssuesToFixes.Keys) {
+            $knownIssue = GetVersionFromString $key
+            Write-Verbose "Testing Known Issue Build $knownIssue"
+
+            if ($null -ne $knownIssue -and
```
lusassl-msft, line 63: Nit: Same here (add brackets):

```
 if (($null -ne $knownIssue) -and
($CurrentVersion.Minor -eq $knownIssue.Minor)) {
```

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerKnownBuildIssues.ps1
```diff
@@ -0,0 +1,125 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+. $PSScriptRoot\Add-AnalyzedResultInformation.ps1
+. $PSScriptRoot\..\Helpers\Invoke-CatchActions.ps1
+
+Function Invoke-AnalyzerKnownBuildIssues {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
```
lusassl-msft, line 10: Let's add `[ValidateNotNullOrEmpty()]` to make sure that no `$null` or empty values are passed

bill-long: A Mandatory parameter already can't be null or empty.

image

lusassl-msft: Can you pass “ “ (with blank space) instead of “”? Does that work for required parameters?

dpaulson45: Could probably pass an empty value since it is string param. But what we are passing as a `ref` not sure how that is possible to be a problem.

bill-long: Yes you can pass a space. That satisfies the requirement of Mandatory. ValidateNotNullOrEmpty doesn't block that, either, because it's not null or empty, it's a space.

You can't pass an empty string, even when it's a string type (see above).

bill-long: image

lusassl-msft: Got it. So let’s do it without the validation.

lusassl-msft (approved):

PR #835, opened 2022-01-02
Optionally enable antimalware scanning when done

bill-long:
With this change, the script offers to enable malware scanning when it is complete. Note that we always prompt before enabling. This is so the user can verify the script actually succeeded before enabling.

The prompt can be disabled by passing -DontEnableAntiMalwareScanning.

### DIFF
--- a/Admin/Reset-ScanEngineVersion.ps1
+++ b/Admin/Reset-ScanEngineVersion.ps1
@@ -3,8 +3,10 @@
 
 #Requires -Version 3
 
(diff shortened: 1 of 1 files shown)


bill-long:
This is what the experience looks like with this change.

image

bill-long:
Do not merge this. This is still under discussion.

bill-long:
Closing for now. Will keep the branch alive in case we decide to proceed with this.

PR #836, opened 2022-01-02
Multiple fixes based on feedback

bill-long:
This PR contains the following fixes:

* Check if server has an impacted version starting with 22 before running. Can override with -Force.
* Check if server has mailbox role, as it is not needed to run on Edge servers. Can override with -Force.
* Verify updateprocess.exe is actually killed before proceeding. If not, script asks user to kill process and rerun script.
* Say exactly what we're stopping instead of just "stopping services".
* Don't attempt to update progress if values are null.

Behavior with the new version (I faked the role issue as I don't have an Edge role):

image

image

### DIFF
--- a/Admin/Reset-ScanEngineVersion.ps1
+++ b/Admin/Reset-ScanEngineVersion.ps1
@@ -17,7 +17,7 @@ begin {
         }
 
         function StopServicesAndProcesses {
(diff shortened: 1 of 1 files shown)


lusassl-msft (dismissed):

lusassl-msft (approved):

Batre-MSFT (approved):
Approved

PR #844, opened 2022-01-04
Updated FIP-FS issue for Failed agent paths and no agents

dpaulson45:
**Issue:**
Customer was reporting this issue:

```
Der Pfad "C:\Program Files\Microsoft\Exchange Server\V15\FIP-FS\Data\Engines\amd64\Microsoft\Bin" kann nicht gefunden werden, da er nicht vorhanden ist.
bei GetFolderFromExchangeInstallPath, C:\wartung\HealthChecker.ps1: Zeile 3966
bei Invoke-ScriptBlockHandler<Process>, C:\wartung\HealthChecker.ps1: Zeile 374
bei GetHighestScanEngineVersionNumber, C:\wartung\HealthChecker.ps1: Zeile 3985
bei Get-FIPFSScanEngineVersionState<Process>, C:\wartung\HealthChecker.ps1: Zeile 4005
bei Get-ExchangeInformation, C:\wartung\HealthChecker.ps1: Zeile 4764
bei Get-HealthCheckerExchangeServer, C:\wartung\HealthChecker.ps1: Zeile 6159
bei Main, C:\wartung\HealthChecker.ps1: Zeile 8157
bei <ScriptBlock>, C:\wartung\HealthChecker.ps1: Zeile 8195
bei <ScriptBlock>, <Keine Datei>: Zeile 1
```

Doesn't have the FIP-FS bin path, which can occur after removing the agents. After talking with Bill about this, we need to have an agent there for things to work properly. 

**Reason:**
Handled the failure to find the path to avoid errors bubbling up. 

**Fix:**
Did a `Test-Path` prior to the `Get-ChildItem`. If we failed to find the path, don't attempt to get the child items but still return a similar object in `GetFolderFromExchangeInstallPath`. If we have `Failed` set on the return object to `$true` provide that we couldn't find the path and return `$null` back. 

Updated the analyzer wording of the `$null` results to provide that this can be an issue and to report it as red still. 

**Validation:**
Pester testing and lab.


bill-long (dismissed):

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-FIPFSScanEngineVersionState.ps1
```diff
@@ -58,12 +61,17 @@ Function Get-FIPFSScanEngineVersionState {
...
                 if ($null -ne $scanEngineVersions) {
-                    [Int64]$highestScanEngineVersion = ($scanEngineVersions.Name | Measure-Object -Maximum).Maximum
+                    if ($scanEngineVersions.Failed) {
+                        Write-Verbose "Failed to find the scan engine directory"
+                        return $null
+                    } else {
+                        [Int64]$highestScanEngineVersion = ($scanEngineVersions.Name | Measure-Object -Maximum).Maximum
+                    }
                 } else {
                     Write-Verbose "No FIP-FS scan engine version(s) detected"
```
lusassl-msft, line 71: Changes made to the check are looking good to me. One thing to point out: 
I think we should change line 71 by either removing the `else` statement or changing the wording and adding `return $null`. 
We should only end up here if `GetHighestScanEngineVersionNumber` returns `$null`. This can be the case if, a file name + extension was passed to the `GetHighestScanEngineVersionNumber` method (we have a RegEx filter in place to check whether `$ExchangeSubDir` is a directory path or not and if it's not we simply return `$null`. The other case would be a catastrophic failure of the call and nothing's returned.

lusassl-msft (changes requested):

lusassl-msft (approved):

lusassl-msft (approved):

PR #866, opened 2022-01-16
DLT365Groupupgrade script add new condition &UI enhancements

hazemembaby:
Issue:
**Add MFA Auth, new condition for DL checker, include Solution steps for each check, feedback form at the nd of the script**

Reason:
**enhance the script for including more conditions and better proposing fixes for each condition**

Fix:
Short description of the fix
**Fixes will depend on each condition**

Validation:
Provide if applicable


M365/src/DLT365Groupsupgrade.ps1
```diff
@@ -87,8 +87,8 @@ Function Debugdirsync {
     $IsDirSynced = $Distgroup.IsDirSynced
     if ($IsDirSynced -eq $true) {
         $script:Conditionsfailed++
-        Write-Host "Distribution Group can't be upgraded because it's synchronized from on-premises!" -ForegroundColor Red
-        "Distribution Group can't be upgraded because it's synchronized from on-premises!" | Out-File $ExportPath\DlToO365GroupUpgradeChecksREPORT.txt -Append
+        Write-Host "Distribution Group can't be upgraded because it's synchronized from on-premises!`n" -ForegroundColor Red
```
bill-long, line 90: Consider creating an output function that writes to both the log and the host, so we don't have duplicate output lines everywhere. That can be a future work item.

bill-long (approved):

M365/src/DLT365Groupsupgrade.ps1
hazemembaby: Thank you for reporting that will work on it on next PR!

PR #912, opened 2022-02-25
Add SHA256 file hashes to release notes

lusassl-msft:
**Description:**
Write out the SHA256 checksum for each script on the release page. We must create the file hash post script signing to make sure that the hash is accurate. We simply read the `ScriptVersions.csv` file which was created during `Build.ps1` run and ~~overwrite~~ create the `ScriptVersions.txt` ~~which was also initially created during Build.ps1 run~~. We also add the file hash to the `ScriptVersions.csv`.

### DIFF
--- a/.build/BuildScriptVersions.ps1
+++ b/.build/BuildScriptVersions.ps1
@@ -16,14 +16,12 @@ if (Test-Path -Path $scriptVersionsCsv) {
     $versionsFileCSV = ConvertFrom-Csv -InputObject (Get-Content -Path $scriptVersionsCsv)
 
     # Generate final ScriptVersions.txt file (with SHA256 hash) for release description
-
     $versionFile = "$distFolder\ScriptVersions.txt"
     New-Item -Path $versionFile -ItemType File -Force | Out-Null
     "Script | Version | SHA256 Hash" | Out-File $versionFile -Append
     "-------|---------|------------" | Out-File $versionFile -Append
     foreach ($script in $versionsFileCSV) {
-        $fullFilePath = "$($distFolder)\$($script.File)"
-        $sha256Hash = $((Get-FileHash -Path $fullFilePath).Hash)
+        $sha256Hash = $((Get-FileHash -Path "$($distFolder)\$($script.File)").Hash)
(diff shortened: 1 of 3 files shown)


bill-long:
It's intended for the CSV and TXT files to have the same data. You can use the CSV for programmatic verification and the TXT file for visual. So I think it would be better to generate the SHA256 hash when building the CSV and add a new column to the CSV to hold those values. Then when converting the CSV to TXT we add the new column.

This way it's easy to verify the hash by script or other code - for example if we want our scripts that update themselves to start verifying hashes after downloading.

bill-long:
Oh, but it has to be _after_ signing. I guess we need to generate the CSV, sign, then rewrite the CSV with hashes, then convert to TXT.

bill-long:
Pushed 2 lines of code to add the hash to CSV.

lusassl-msft:
Looks good. Should do the job. Thanks @bill-long

.build/BuildScriptVersions.ps1
```diff
@@ -0,0 +1,34 @@
...
+    "Script | Version | SHA256 Hash" | Out-File $versionFile -Append
+    "-------|---------|------------" | Out-File $versionFile -Append
+    foreach ($script in $versionsFileCSV) {
+        $sha256Hash = $((Get-FileHash -Path "$($distFolder)\$($script.File)").Hash)
+        $script | Add-Member -MemberType NoteProperty -Name SHA256Hash -Value $sha256Hash
+        "$($script.File) | $($script.Version) | $sha256Hash" | Out-File $versionFile -Append
+        Write-Host ("File: '{0}' Version: '{1}' Hash: '{2}' added" -f $script.File, $script.Version, $sha256Hash)
+    }
+
+    $versionsFileCSV | Export-Csv -Path $scriptVersionsCsv
```
dpaulson45, line 30: This should include `-NoTypeInformation` like the original csv file.

bill-long: It already wasn't adding type information, maybe because these objects were PSObject because they were imported from CSV:

image

But I added it anyway.

dpaulson45 (approved):

Issue #920, opened 2022-03-07
[Issue] ExchangeMitigations.ps1 breaks Exchange 2013 CU23


viniciusferrao:
**Describe the issue**
After running `ExchangeMitigations.ps1` on an Exchange Server 2013 CU23 running on Windows Server 2012 (non-R2) the server is unusable.

There's some warning that IIS Rewrite Module 2.1 will break on IIS8 but 2.0 is no longer available, I'm not sure if this is related or not.

Server was unable to even run `Get-HealthReport` and ECP/OWA returned 503. After running `ExchangeMitigations.ps1` rollback, everything went to normal again.

**Expected behavior**
Don't break the server

**Script Output**
Unfortunately I've lost. But was everything OK. I do have the rollback instead:

```
PS C:\Users\Administrator.TESTDOMAIN> .\ExchangeMitigations.ps1 -WebSiteNames "Default Web Site" -RollbackAllMitigation
WARNING: Waiting for service 'Microsoft Exchange Unified Messaging (MSExchangeUM)' to start...
WARNING: Waiting for service 'Microsoft Exchange Unified Messaging (MSExchangeUM)' to start...
WARNING: Waiting for service 'Microsoft Exchange Unified Messaging (MSExchangeUM)' to start...
WARNING: Waiting for service 'Microsoft Exchange Unified Messaging Call Router (MSExchangeUMCR)' to start...

Status   Name               DisplayName
------   ----               -----------
Running  MSExchangeUM       Microsoft Exchange Unified Messaging
Running  MSExchangeUMCR     Microsoft Exchange Unified Messagin...
VERBOSE: Performing operation "Set-Item" on Target "-path \\EXCH2013\AppPools\MSExchangeECPAppPool".
VERBOSE: Status of MSExchangeECPAppPool

Value : Started

WARNING: Waiting for service 'Microsoft Exchange Health Manager (MSExchangeHM)' to start...
WARNING: Waiting for service 'Microsoft Exchange Health Manager (MSExchangeHM)' to start...
WARNING: Waiting for service 'Microsoft Exchange Health Manager (MSExchangeHM)' to start...
Running  MSExchangeHM       Microsoft Exchange Health Manager
VERBOSE: Performing operation "Set-Item" on Target "-path \\EXCH2013\AppPools\MSExchangeOABAppPool".
VERBOSE: Status of MSExchangeOABAppPool

Value : Started
```

**Additional context**
[...]

PR #923, opened 2022-03-08
Remove obsolete scripts

bill-long:
To address #920 and make it harder to accidentally run obsolete scripts, this PR:

* Removes the scripts from source.
* Changes the old docs page (these were written before we moved to the new docs, so we left it alone to avoid breaking links) to clearly call out that these are obsolete and point to the final release that included them.

Anything else we should do here? Should we also remove Test-ProxyLogon?


dpaulson45:
Would be good to keep Test-ProxyLogon. It doesn't break anything if someone uses it. Good to keep around as an example source code for the next time we need to do something similar. 

Are we keeping EOMT since that appears to only do actions if on X build?

bill-long:
Yeah, we are keeping EOMT.

lusassl-msft (dismissed):
I'm okay with this change. Nothing to add/remove from my side.

dbacon247:
EOMT won't break anything - although running it on the current builds of Exchange is not necessary it's only necessary for builds prior to the may 2021 CU if I recall - should be documented in the read me/help files. 

ExchangeMitgations.ps1 on the other hand did more drastic remediations such as turning of critical services in Exchange (also throughly documented) so that those who ran ExchangeMitigations.ps1 could roll back those changes and EOMT didn't account for those mitigations.

dpaulson45:
@dbacon247 does that mean you approve this change? Waiting on you to approve it before merge.

dbacon247 (approved):

dpaulson45 (approved):

PR #924, opened 2022-03-08
Improved certificate unit test

lusassl-msft:
**Description:**
Added Mock for `Get-Date` call to make sure that the same number of days is returned via `Get-ExchangeServerCertificates` call.


lusassl-msft:
**Note:**
There is an outage affecting the build pipelines:
https://status.dev.azure.com/_event/293762602

PR #929, opened 2022-03-10
service host crash issue after installing March SU

Batre-MSFT:
service host crash issue after installing March SU

**Issue:**

**Reason:**

**Fix:**
Short description of the fix

**Validation:**
Provide if applicable


Batre-MSFT:
@dpaulson45 @lusassl-msft please approev/commit

bill-long (approved):

PR #949, opened 2022-03-24
New function Get-ExchangeConnectors

### LINKED ISSUE #919: [Health Checker] Check Receive/SendConnector TlsCertificateName in hybrid environments
**Describe the issue**
A misconfigured Send/ReceiveConnector can lead to mail flow problem. Especially in hybrid environments. We should query the `TlsCertificateName` and compare it against the existing certificates on the server. The correct format is: `$tlscertificatename = "$($cert.Issuer)$($cert.Subject)"`.

lusassl-msft:
**Issue:**
Resolve #919 

**Description:**
New function to query Exchange Send-/Receive Connectors. The function is general and can easily be extended (e.g. other use cases).

How does it work? 
`Get-ExchangeConnectors` contains the core functionality of this new function. 

- `ExchangeConnectorObjectFactory` generates custom objects with the same set of properties for send and receive connectors
- `NormalizeTlsCertificateName` is used to check and normalize the `TlsCertificateName` which might be configured for the send-/receive connector
- `FindMatchingExchangeCertificate` is used to find the matching certificated on the Exchange server

`ExchangeConnectorObjectFactory` is a general function and can be easily adjusted for other connector related scenarios. Just add the properties that you need.

~~Do not merge yet. Still in development. Feel free to review.~~

**Validation:**
Lab / Pester


Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerHybridInformation.ps1
```diff
@@ -150,5 +150,122 @@ Function Invoke-AnalyzerHybridInformation {
...
+
+                    $connectorCertificateMatchesHybridCertificateWritingType = "Yellow"
+                    if (($connector.TlsCertificateSet) -and
+                        (-not([System.String]::IsNullOrEmpty($exchangeInformation.GetHybridConfiguration.TlsCertificateName))) -and
+                        ($connector.TlsCertificateName -eq $exchangeInformation.GetHybridConfiguration.TlsCertificateName)) {
+                        $connectorCertificateMatchesHybridCertificate = $true
+                        $connectorCertificateMatchesHybridCertificateWritingType = "Green"
+                    }
+
+                    $AnalyzeResults | Add-AnalyzedResultInformation -Name "Certificate Matches Hybrid Certificate" -Details $connectorCertificateMatchesHybridCertificate `
```
dpaulson45, line 251: `$connectorCertificateMatchesHybridCertificate` should be set as `$false` to have something come up. Otherwise by default it will be blank.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeConnectors.ps1
```diff
@@ -0,0 +1,219 @@
...
+                }
+
+                if (($issuerIndex -ne -1) -and
+                    ($subjectIndex -ne -1)) {
+                    if ($expectedTlsCertificateNameDetected) {
+                        $issuer = $TlsCertificateName.Substring(($issuerIndex + 3), ($subjectIndex - 3))
+                        $subject = $TlsCertificateName.Substring($subjectIndex + 3)
+                    } else {
+                        $issuer  = $TlsCertificateName.Substring($issuerIndex, $subjectIndex)
+                        $subject = $TlsCertificateName.Substring($subjectIndex)
```
dpaulson45, line 108: If `$issuerIndex` or `$subjectIndex` is set to a `-1` value this will throw an error. 

image

dpaulson45 (changes requested):

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeConnectors.ps1
lusassl-msft: Yep, therefore I've placed it inside this if statement (or have I misunderstood something?):
image

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerHybridInformation.ps1
lusassl-msft: You're right. Fixed.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeConnectors.ps1
dpaulson45: That if statement requires both `$issuerIndex` AND `$subjectIndex` to not equal `-1`. What about when only 1 of those variables are set to `-1`?

lusassl-msft: We don't have enough information available when only one of both has a value other than -1. In this case `NormalizeTlsCertificateName` doesn't return anything. However, this case should not occur in normal circumstances as it's not possible to set any other value than `$null` or `(<i>).*(<s>).*` for `TlsCertificateName` via PowerShell.  It's more a failsafe and we normally should not end up here.

dpaulson45: Right, and what I am saying is that you will cause an error to occur here if one of the values are `-1`. Thus, the `else` statement wouldn't really execute the way you are wanting.

lusassl-msft: Ah, I know what you mean. The idea of the `else `statement is to handle this particular case (which should also not occur - just a failsafe): 

image
`$tlscertificatename = "$($cert.Issuer)$($cert.Subject)"` (without <I> and <S>)

lusassl-msft: Added a comment to the script as this is by design.

dpaulson45:
@lusassl-msft the new CodeFormatter.ps1 script also requires a blank space at the end of the .md files as well.

Let's also discuss the rest of this here today.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Tests/Get-ExchangeConnectors.Tests.ps1
```diff
@@ -0,0 +1,263 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSAvoidUsingInvokeExpression', '', Justification = 'Pester testing file')]
```
dpaulson45, line 4: Do you need these overrides? Not sure if the rule would hit on this file.

dpaulson45 (changes requested):

dpaulson45 (approved):

PR #951, opened 2022-03-30
Exchange Log Collector - Removed legacy code and handled new logging

dpaulson45:
**Reason:**
Need to remove the old legacy code like `Write-VerboseWriter` and `Write-ScriptMethodVerboseWriter` where it was used instead of the override option that we now have with `Write-Verbose` and still being able to call the native `Write-Verbose` within the override method. 

**Fix:**
Updated code enough to remove the old functions and be able to handle logging while inside of a `Invoke-Command` to be returned on the pipeline to give you better error context. 

**Validation:**
Tested within lab with multiple servers and single server.


bill-long (approved):
This looks fine for now. Compress-Folder and New-Folder (does that even need to exist?) should be looked at in a future PR. They look like old code that could use some work.

dpaulson45:
> This looks fine for now. Compress-Folder and New-Folder (does that even need to exist?) should be looked at in a future PR. They look like old code that could use some work.

Yep, will optimize that code here in a different PR.

New-Folder still needs to exist as it is called a few different times and this way it will always do the Test-Path prior to attempting to create the folder to avoid errors from occurring.

bill-long:
> New-Folder still needs to exist as it is called a few different times and this way it will always do the Test-Path prior to attempting to create the folder to avoid errors from occurring.

That's what -Force does.

image

It either creates the folder (and every other folder in the path you specified that doesn't already exist), or returns the folder if it already exists.

dpaulson45:
> > New-Folder still needs to exist as it is called a few different times and this way it will always do the Test-Path prior to attempting to create the folder to avoid errors from occurring.
> 
> That's what -Force does.
> 
> image
> 
> It either creates the folder (and every other folder in the path you specified that doesn't already exist), or returns the folder if it already exists.

Was not aware of the Force switch.

PR #953, opened 2022-04-11
Update Invoke-EnableExtraTracing.ps1

iserrano76:
**Issue:**
I tried to run the VSStester in an environment with a non-English language in the OS and it fails 

**Reason:**
To allow the script in any language.

**Fix:**
Short description of the fix
VSStester use a localized string in the comparison.
If we use the LASTEXITCODE variable instead of the result and we verify that we receive 0 as result should work for any language.

**Validation:**


bill-long:
Thanks for the PR! There's a small issue with the formatting. In your local clone please run:

`.build\CodeFormatter.ps1 -Save`

That should highlight some issues and auto-fix them if it can. Run the same command a second time and hopefully it's clean. At that point, commit the formatting fixes and push to this branch.

Also you'll need to sign the CLA before we can merge.

bill-long:
Looks good now. Thank you!

PR #963, opened 2022-04-15
Update Get-EASMailboxLogs.ps1

DKhrebin:
Add a parameter to the Get-EASMailboxLogs script to enable/disable the EnableMailboxLoggingVerboseMode setting in the EAS web.config file on the current server.

**Reason:**
In a number of scenarios, we need detailed logging of all information transmitted within the framework of the EAC request, including all the contents of the letter, attachments, etc.

**Fix:**
Added code to change EnableMailboxLoggingVerboseMode value in EAS web.config


Admin/Get-EASMailboxLogs.ps1
```diff
@@ -6,9 +6,28 @@ param(
     [Parameter(Mandatory = $true)] [string[]] $Mailbox = "",
     [ValidateScript({ Test-Path $_ })]
     [Parameter(Mandatory = $true)] [string] $OutputPath = (Get-Location).Path,
-    [Parameter(Mandatory = $false)] [int] $Interval = 30
+    [Parameter(Mandatory = $false)] [int] $Interval = 30,
+    [Parameter(Mandatory = $false)] [string] $EnableMailboxLoggingVerboseMode = ""
```
bill-long, line 10: Seems odd to have a string parameter that will be interpreted as a bool. Why not just make the parameter bool or switch? I think switch would be more natural here.

Edit: See https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters

image

bill-long (changes requested):

Admin/Get-EASMailboxLogs.ps1
DKhrebin: Bill, point is: when you use Boolean value it is always true or false (be default).
It means that we need somehow understand if False was set by Admin or by Power Shell. As a result, we can override web.config True setting, set in other debugging session.
In other words, I prefer not to touch web.config until I see that False was obviously set by Admin.

bill-long: I see. So either $true or $false could lead to a config change, so we need a third value. Hmm, can we have nullable bool in PowerShell, like `bool?` in C#?

Apparently not. This StackOverflow answer suggests using a ValidateSet with 3 values, true, false, or null: https://stackoverflow.com/questions/51207288/is-it-possible-to-have-nullablebool-in-a-bool-in-powershell

Another option would be two separate switches: -EnableMailboxLoggingVerboseMode and -DisableMailboxLoggingVerboseMode, and make them mutually exclusive.

Admin/Get-EASMailboxLogs.ps1
```diff
@@ -6,9 +6,28 @@ param(
...
+    default { Write-Warning "Wrong value $EnableMailboxLoggingVerboseMode in EnableMailboxLoggingVerboseMode will be ignored"; Write-Warning "Press any key to continue ..."; $host.UI.RawUI.ReadKey("NoEcho,IncludeKeyDown") | Out-Null }
+}
+
+#Override EnableMailboxLoggingVerboseMode key's value with EnableVerboseLogging
+If ($NeedChangeConfig) {
+    [xml]$web = Get-Content $env:ExchangeInstallPath"ClientAccess\Sync\web.config"
+    $web.SelectSingleNode('//add[@key="EnableMailboxLoggingVerboseMode"]').Value=$EnableVerboseLogging
+    #if ($EnableVerboseLogging) { $web.SelectSingleNode('//add[@key="EnableMailboxLoggingVerboseMode"]').Value="true"}
+    #else {$web.SelectSingleNode('//add[@key="EnableMailboxLoggingVerboseMode"]').Value = "false"}
+    $web.Save($env:ExchangeInstallPath+"ClientAccess\Sync\web.config")
```
dpaulson45, line 28: The original `web.config` file should be saved somewhere else when we process this. There should also be a `ShouldProcess` on `$web.Save()`

dpaulson45:
The Docs page should also be updated in this pull request of the new parameter(s) once this is ready to be merged.

dpaulson45:
@DKhrebin are you going to be able to address the above comments?

DKhrebin:
Hi David,

Yes. Thinking about -WhatIf output as default outcome for it is not covering changes.

Dmitry Khrebin
Exchange Escalation Engineer
EMEA
Customer Services and Support
Office: +351 (21) 0605382
Mobiile: +351 (92) 797-77-21
Working hours: 09:00-18:00 (UTC +00:00) Mon-Fri
[Microsoft Logo]
Backup | Alexey Bogomolov ***@***.******@***.***>) 7 (499) 9269382
Team Manager | Yolande Dupoy ***@***.******@***.***>) +33 (1) 57753129

lusassl-msft:
@DKhrebin please run `.build\CodeFormatter.ps1 -Save` and commit again. There is a code-formatting issue: `WARNING: Keyword If at position 937 in D:\a\1\s\Admin\Get-EASMailboxLogs.ps1 is not lowercase.`

DKhrebin:
@lusassl-msft done (empty string removed)

dpaulson45:
@DKhrebin, you created a new branch. Git is case sensitive. This branch is lower case `dkhrebin` under your remote location of `Dkhrebin`

DKhrebin:
@dpaulson45 Fixed.

PS C:\GitHub\Css-Exchange\CSS-Exchange> git push Dkhrebin HEAD:dkhrebin 
Everything up-to-date

Admin/Get-EASMailboxLogs.ps1
```diff
@@ -2,13 +2,34 @@
...
+#parse $EnableMailboxLoggingVerboseMode and convert it boolean value(s)
+$NeedChangeConfig = $false
+switch ($EnableMailboxLoggingVerboseMode) {
+    $true { $NeedChangeConfig = $true; $EnableVerboseLogging = "true" }
+    $false { $NeedChangeConfig = $true; $EnableVerboseLogging = "false" }
+}
+
+#Override EnableMailboxLoggingVerboseMode key's value with EnableVerboseLogging
+if ($NeedChangeConfig) {
+    if ($PSCmdlet.ShouldProcess("Set EnableMailboxLoggingVerboseMode attribute to $EnableVerboseLogging in $env:ExchangeInstallPath" + "ClientAccess\Sync\web.config", 'TARGET', 'OPERATION')) {
```
dpaulson45, line 23:
```suggestion
switch ($EnableMailboxLoggingVerboseMode) {
    $true { $EnableVerboseLogging = "true" }
    $false { $EnableVerboseLogging = "false" }
    default { $EnableVerboseLogging = $null }
}

#Override EnableMailboxLoggingVerboseMode key's value with EnableVerboseLogging
if ($null -ne $EnableVerboseLogging) {
    if ($PSCmdlet.ShouldProcess("Set EnableMailboxLoggingVerboseMode attribute to $EnableVerboseLogging in $env:ExchangeInstallPath" + "ClientAccess\Sync\web.config", 'TARGET', 'OPERATION')) {
```

Admin/Get-EASMailboxLogs.ps1
```diff
@@ -2,13 +2,34 @@
...
+        [xml]$web = Get-Content $env:ExchangeInstallPath"ClientAccess\Sync\web.config"
+        Copy-Item $env:ExchangeInstallPath"ClientAccess\Sync\web.config" -Destination $env:ExchangeInstallPath"ClientAccess\Sync\web.config.bak"
```
dpaulson45, line 25: should be placed inside of a `try` `catch` block

Admin/Get-EASMailboxLogs.ps1
```diff
@@ -2,13 +2,34 @@
...
+        $web.SelectSingleNode('//add[@key="EnableMailboxLoggingVerboseMode"]').Value = $EnableVerboseLogging
+        $web.Save($env:ExchangeInstallPath + "ClientAccess\Sync\web.config")
+    } else {
+        break
```
dpaulson45, line 29: why is the `break` here? I think this needs to be removed.

dpaulson45 (changes requested):

Admin/Get-EASMailboxLogs.ps1
DKhrebin: To exit script without enabling logging.

dpaulson45: Then we should use `exit` instead of break.

Admin/Get-EASMailboxLogs.ps1
```diff
@@ -2,13 +2,39 @@
...
+    if ($PSCmdlet.ShouldProcess("Set EnableMailboxLoggingVerboseMode attribute to $EnableVerboseLogging in $env:ExchangeInstallPath" + "ClientAccess\Sync\web.config", 'TARGET', 'OPERATION')) {
+        try {
+            [xml]$web = Get-Content $env:ExchangeInstallPath"ClientAccess\Sync\web.config"
+        } catch {
+            Write-Error ("Failed to read $env:ExchangeInstallPath" + "ClientAccess\Sync\web.config")
+            exit
+        }
+        Copy-Item $env:ExchangeInstallPath"ClientAccess\Sync\web.config" -Destination $env:ExchangeInstallPath"ClientAccess\Sync\web.config.bak"
+        $web.SelectSingleNode('//add[@key="EnableMailboxLoggingVerboseMode"]').Value = $EnableVerboseLogging
+        $web.Save($env:ExchangeInstallPath + "ClientAccess\Sync\web.config")
```
dpaulson45, line 32: Include all of this in the `try` `catch` block

Admin/Get-EASMailboxLogs.ps1
```diff
@@ -2,13 +2,39 @@
...
+    default { $EnableVerboseLogging = $null }
+}
+
+#Override EnableMailboxLoggingVerboseMode key's value with EnableVerboseLogging
+if ($null -ne $EnableVerboseLogging) {
+    if ($PSCmdlet.ShouldProcess("Set EnableMailboxLoggingVerboseMode attribute to $EnableVerboseLogging in $env:ExchangeInstallPath" + "ClientAccess\Sync\web.config", 'TARGET', 'OPERATION')) {
+        try {
+            [xml]$web = Get-Content $env:ExchangeInstallPath"ClientAccess\Sync\web.config"
+        } catch {
+            Write-Error ("Failed to read $env:ExchangeInstallPath" + "ClientAccess\Sync\web.config")
```
dpaulson45, line 27: Update `Write-Error` to include exception and change wording once all of the other code is added to `try` block.

Admin/Get-EASMailboxLogs.ps1
DKhrebin: Separate block for each call? To have clear error explanation?

Admin/Get-EASMailboxLogs.ps1
DKhrebin: @dpaulson45  We are doing 3 operations. Read, Copy and Write to existing file. Theoretically we can have problem on any of those operations. Due funny access rights to corresponding directory. Should I put all issues in one bin or separate to have clear error statement?

bill-long (approved):

dpaulson45 (approved):

DKhrebin:
I guess docs also need to be updated, right?

dpaulson45:
yeah, you will need to do that in a different PR however.

PR #965, opened 2022-04-18
TLS Output Display Changes

### LINKED ISSUE #538: HealthChecker - Add data collection for SchUseStrongCrypto of the registry
This is currently added to the script, but a warning is present. Need to check to see how we can 'fix' or address this as we are getting inconsistent results in the labs. 
 
-----------------------------------------------------

Running into some odd issues where some applications use .NET TLS version by running this command: 

```
[System.Net.ServicePointManager]::SecurityProtocol
```

In order to call this out a little better, throw a warning if the above contains `TLS` or `TLS11` when we have them disabled.

https://docs.microsoft.com/en-us/dotnet/framework/network-programming/tls#schusestrongcrypto
https://techcommunity.microsoft.com/t5/exchange-team-blog/exchange-server-tls-guidance-part-3-turning-off-tls-1-0-1-1/ba-p/607898

```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\.NETFramework\v2.0.50727]
"SystemDefaultTlsVersions"=dword:00000001
"SchUseStrongCrypto"=dword:00000001

[HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\.NETFramework\v4.0.30319]
"SystemDefaultTlsVersions"=dword:00000001
"SchUseStrongCrypto"=dword:00000001

[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\.NETFramework\v2.0.50727]
"SystemDefaultTlsVersions"=dword:00000001
"SchUseStrongCrypto"=dword:00000001

[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\.NETFramework\v4.0.30319]
"SystemDefaultTlsVersions"=dword:00000001
"SchUseStrongCrypto"=dword:00000001
```

Migrated From: https://github.com/dpaulson45/HealthChecker/issues/482

### LINKED ISSUE #849: [Feature Request] HealthChecker should display available TLS Cipher Suites
**Describe the work**
The following feature request was received:
It would be helpful (especially for troubleshooting purposes) to display the available TLS Cipher Suites for the computer against which the script is executed. The requested information is available via `Get-TlsCipherSuite | Format-Table -Property CipherSuite, Name` (https://docs.microsoft.com/en-us/powershell/module/tls/get-tlsciphersuite?view=windowsserver2022-ps). 
The cmdlet is available starting with Windows Server 2016 +.

### LINKED ISSUE #918: Health Checker - Improve TLS Output Display
**Describe the issue**
Currently, the TLS settings are just the raw values of all the registry settings. Then, we only display an issue if it is misconfigured. With EXO going through the process of finally disabling TLS 1.0/1.1, it would be better if we display if TLS 1.0/1.1 are fully disabled or enabled. 

Also, should improve the output to not be as complicated. 

**Additional context**
When addressing this issue, please include the issues:

#849 
#538

dpaulson45:
**Issue:**
The TLS output information was mostly in Grey or default text which made it hard to determine if there was a problem. It was also in a list format vs a table format to make things easier to compare against the different TLS versions. 

image

**Reason:**

Due to the list view, decided to use the `Out-Columns` feature of the script to display the information in an easier to read view. Also added in the TLS Cipher Suites to be collected. 

image

**Fix:**
- Moved TLS Setting to be under Security Settings vs the LmCompatibilityLevel
- Changed TLS to use Out-Columns instead of being in a list view
- Included NET Frame TLS settings that are related to be in an Out-Column view
- Included both versions of .NET registry locations (likely not required but can be useful if 3rd party is involved)
- Included TLS Cipher Suites, which is also in the Out-Columns view
- Updated the TLS Docs on the check and provided the configurations states 
- Updated HTML Report to be able to include Out-Column type display objects
- Updated HTML Report to also do the color coding the same as on screen display
- Updated Pester for the new logic in place

Resolved #849 
Resolved #538 
Resolved #918 

**Validation:**
Tested in lab and with Pester


bill-long (approved):

bill-long:
One thing to be aware of - I'm seeing a lot of backtick line continuations. Style guide advises against this:

https://github.com/PoshCode/PowerShellPracticeAndStyle/blob/master/Style-Guide/Readability.md

Because of how the results are being piped to the function, I'm not clear what this would look like if we splatted the parameters. Backticks might actually be the better option here? But it's something to keep in mind.

lusassl-msft (approved):

PR #992, opened 2022-04-26
ExchangeConnector unit test adjusted

lusassl-msft:
**Issue:**
Fixed the `CertificateLifetimeInfo` validation in `Get-ExchangeConnector` unit test.


dpaulson45:
Still failing locally, this is the problem. Works either on my machine or in the build system but not both for some reason. I am guessing it has to do with the timezone of where it is running. 

Do we need to test the lifetime information here? We are testing it in the other pester test. 

image

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Tests/Get-ExchangeConnectors.Tests.ps1
```diff
@@ -130,11 +130,13 @@ Describe "Testing Get-ExchangeConnectors.ps1" {
             ($results[5].CertificateDetails.CertificateLifetimeInfo).Count | Should -Be 2
             foreach ($key in ($results[5].CertificateDetails.CertificateLifetimeInfo).keys) {
                 if ($key -eq "E267D459A0FB53D0EF225C11FAC062D522648C09") {
-                    ($results[5].CertificateDetails.CertificateLifetimeInfo)[$key] | Should -Be 1678
+                    $testDays = ([System.Convert]::ToDateTime($multipleMatchingExchangeCertificates[3].CertificateObject.NotAfter, [System.Globalization.DateTimeFormatInfo]::InvariantInfo) - (Get-Date)).Days
```
dpaulson45, line 133: While this does work, not sure if this is the best way to go about the test.

dpaulson45: Would this be a test case that we need to have? We could hard code the UTC time here, similar to what we had before, that might be better.

dpaulson45 (dismissed):

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Tests/Get-ExchangeConnectors.Tests.ps1
lusassl-msft: I've switched to using the `BaseUtcOffset` (https://docs.microsoft.com/en-us/dotnet/standard/datetime/converting-between-datetime-and-offset) to make sure that the `DateTime` that is being used in the unit test is the same on all computers (independent of the local configured time zone). I've added a `ConvertTimeToUtcHelper` function to perform the operations.

dpaulson45:
Check out this possible resolution: https://github.com/microsoft/CSS-Exchange/commit/3bcefa8909ea7a35812c6a06d3de3f7f571613c8 this is likely the best option to address the issue.

lusassl-msft:
> Check out this possible resolution: [3bcefa8](https://github.com/microsoft/CSS-Exchange/commit/3bcefa8909ea7a35812c6a06d3de3f7f571613c8) this is likely the best option to address the issue.

That’s exactly what I’ve tried. It worked okay in PST time zone but not in German time zone. I decided for this reason to go with the BaseUtcOffset to do the calculation to UTC.

lusassl-msft:
@dpaulson45 Made a few more adjustments and it should work now across time zones (validated in three different TZ). Please have a look to close this long runner timely. Thanks.

dpaulson45:
@lusassl-msft it is still failing locally.


image

dpaulson45:
At this point, is it worth it to have the days worth be tested? The amount of time putting into this is a bit much for the value add that we would be getting out of this.

lusassl-msft:
Strange. But yes, we don't need to do the lifetime check for the `Get-ExchangeConnectors.Tests.ps1`. Does the `Get-ExchangeServerCertificates.Tests.ps1` works correctly for you?

dpaulson45:
> Strange. But yes, we don't need to do the lifetime check for the `Get-ExchangeConnectors.Tests.ps1`. Does the `Get-ExchangeServerCertificates.Tests.ps1` works correctly for you?

No, that one fails as well.

dpaulson45 (approved):

PR #1002, opened 2022-05-03
Update Get-ExchangeBuildVersionInformation.ps1

alainQtec:
**Issue:**
When I ran Get-ExchangeBuildVersionInformation it throws bunch of errors, Instead of returning the version.
It should return zero if not found. or any version greater than zero if The app is Installed.

**Reason:**
Erroring when parameter is missing or $null

**Fix:**
Prevent throwing Runtime Exceptions
By making parameter Mandatory, and add try-catch blocks.

**Validation:**
Provide if applicable

### DIFF
--- a/Setup/Shared/Get-FileInformation.ps1
+++ b/Setup/Shared/Get-FileInformation.ps1
@@ -14,7 +14,6 @@ Function Get-FileInformation {
         $revNumber = [string]::Empty
     }
     process {
-
         if (-not($installerCOM) -and
             -not($AllowFileSubjectOnly)) {
             throw "Failed to create 'WindowsInstaller.Installer' COM object. This can lead to issues with validation of the script."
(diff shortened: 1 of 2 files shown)


bill-long:
Thank you for the PR!

Do we know what scenario leads to the function being called with no AdminDisplayVersion? It would be nice to fix the root cause if possible.

Shared/Get-ExchangeBuildVersionInformation.ps1
```diff
@@ -4,44 +4,52 @@
...
+        Write-Verbose "Passed $([string]$AdminDisplayVersion)"
+        $AdminDisplayVersion = [string]$AdminDisplayVersion
         $exchangeMajorVersion = [string]::Empty
         [int]$major = 0
         [int]$minor = 0
         [int]$build = 0
         [int]$revision = 0
         $product = $null
         [double]$buildVersion = 0.0
+        $oldErrorAction = $ErrorActionPreference
```
dpaulson45, line 22: Cannot have the code written this way inside of `Shared`. 

This will break the logic in place for capturing all the errors that do occur and are counted.

Setup/Shared/Get-FileInformation.ps1
```diff
@@ -34,19 +33,26 @@ Function Get-FileInformation {
...
+                    $shellApplication = New-Object -ComObject "Shell.Application"
+                } catch {
+                    "Failed to create 'Shell.Application' COM Object. This can lead to issues with validation of the script." | Receive-Output -ForegroundColor Red
+                    break
+                }
+                # One $fileItem Example:
+                # $localPackageChildItems = Get-ChildItem -Path "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer" -Recurse | Where-Object { $_.Property -eq "LocalPackage" };
+                # $File = [IO.FileInfo]$localPackageChildItems[1].GetValue("LocalPackage");
+                # $fileItem = Get-Item $File
+                $fileItem = Get-Item $File
```
dpaulson45, line 47: What is this code about? The PR is regarding Get-ExchangeBuildVersionInformation.ps1 and doesn't provide anything about what we are fixing in here. 

Secondly, setup code does NOT call Get-ExchangeBuildVersionInformation. These should not be in the same PR anyways.

dpaulson45 (changes requested):

bill-long:
@alainQtec Can you explain what scenario you're trying to address with this PR?

The shared Get-ExchangeVersionInformation.ps1 script is called by two scripts that we release from this repo:

* ExchangeLogCollector.ps1
* HealthChecker.ps1

To test this change, I started by taking our current `main` branch and inserting a `throw` at the top of Get-ExchangeVersionInformation:

image

Then I ran both scripts that use this script.

ExchangeLogCollector:

image

HealthChecker:

image

Then, I cloned your fork and switched to the error-handling branch, and inserted a `throw` at the top once again, but inside the `try`:

image

ExchangeLogCollector:

image

HealthChecker:

image

In the case of ExchangeLogCollector, this change lets it continue collecting data even if we failed to get a version number. That might have some value if we're running into scenarios where we can't get version numbers, though I would question whether this should be handled by Get-ExchangeVersionInformation rather than the caller. If the caller wants to catch and deal with errors, they can do so.

In the case of HealthChecker, we more clearly see the effect of changing this function's behavior without touching the caller. HealthChecker fails because the returned value is not the expected shape. Adding the try/catch in the function alone has no value here.

Right now, I think letting the calling script decide whether to terminate when Get-ExchangeVersionInformation fails is probably the right approach, and no change to Get-ExchangeVersionInformation.ps1 is needed. However, my mind could be changed if you can provide more context around what scenario you're trying to address with these changes.

alainQtec:
You are right; let the caller handle the errors. I had no idea how badly I was messing things up! @bill-long, thank you for clarifying.

PR #1018, opened 2022-05-11
Check CU has fixed CVE-2021-1730

iserrano76:
**Issue:**
CVE-2021-1730 was fixed in CU12(2019) or CU23(2016) but the script shows an alert with those versions and newer.

**Reason:**
Avoid confusions.

**Fix:**
include the new version in the verification of the vulnerability.

**Validation:**

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityCve-2021-1730.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityCve-2021-1730.ps1
@@ -21,9 +21,11 @@ Function Invoke-AnalyzerSecurityCve-2021-1730 {
     #Workaround: N/A
 
     if (((($SecurityObject.MajorVersion -eq [HealthChecker.ExchangeMajorVersion]::Exchange2016) -and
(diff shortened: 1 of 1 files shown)


dpaulson45:
@iserrano76 this is not "fixed" with the latest CUs. We are still required to check the configuration as the CVE provides.

iserrano76:
In the description of the CU12 shows:
Issues that this cumulative update fixes
[5012787] Download domains created per CVE-2021-1730 don’t support ADFS authentication in OWA
(https://support.microsoft.com/en-us/topic/cumulative-update-12-for-exchange-server-2019-kb5011156-6a4e598a-876c-4ff1-9cfa-f7b87246f1d8?msclkid=1b31d3afd13e11ec952bc3dbc7e84795)

And in the following link appear as the solution:
https://support.microsoft.com/en-us/topic/download-domains-created-per-cve-2021-1730-don-t-support-adfs-authentication-in-owa-kb5012787-6bc473e2-ed85-497e-b6df-4341acc28617

Is it still necessary the configuration?

dpaulson45:
Thanks for the PR @iserrano76 we love to see this type of contribution. 

However, we can't accept this as we still require the configuration settings to be checked to protect against CVE-2021-1730.

PR #1022, opened 2022-05-11
Get Default Location of Event Logs from Registry

### LINKED ISSUE #785: [Issue] Exchange Log Collector - Default location of Event Logs
**Describe the issue**
Customer moved the default location of the Event Logs to a different drive. The way the script collects the data is based off the default location and need to determine the correct location on the fly. 

**Expected behavior**
Collect the data correctly based off where the events are being logged at. 

**Additional context**
Should be able to determine the location of the event logs from here in the registry. However, need to figure out where the rest of the Exchange Event Logs would be kept at. 

image

dpaulson45:
**Describe the issue**
Customer moved the default location of the Event Logs to a different drive. The way the script collects the data is based off the default location and need to determine the correct location on the fly. 

**Fix**
Get the appropriate location from the registry if possible. Otherwise, default back to the old logic of the hard coded path. 

Didn't find a registry location for the other Exchange Logs that we are looking for, however, doesn't mean that they aren't there. For this time only worried about these logs as these were the only ones that we have seen moved to a different location before.

Resolved #785

### DIFF
--- a/Diagnostics/ExchangeLogCollector/RemoteScriptBlock/IO/Save-WindowsEventLogs.ps1
+++ b/Diagnostics/ExchangeLogCollector/RemoteScriptBlock/IO/Save-WindowsEventLogs.ps1
@@ -3,6 +3,7 @@
 
 . $PSScriptRoot\Copy-BulkItems.ps1
 . $PSScriptRoot\Save-DataInfoToFile.ps1
+. $PSScriptRoot\..\..\..\..\Shared\Get-RemoteRegistryValue.ps1
 Function Save-WindowsEventLogs {
 
     Write-Verbose("Function Enter: Save-WindowsEventLogs")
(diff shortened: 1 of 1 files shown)


Diagnostics/ExchangeLogCollector/RemoteScriptBlock/IO/Save-WindowsEventLogs.ps1
```diff
@@ -14,15 +15,52 @@ Function Save-WindowsEventLogs {
...
+
+        if ($null -eq $applicationLogs) { $applicationLogs = "$rootLogPath\Application.evtx" }
+
+        if ($null -eq $systemLogs) { $systemLogs = "$rootLogPath\System.evtx" }
+
+        if ($null -eq $managementLogs) { $managementLogs = "$rootLogPath\MSExchange Management.evtx" }
+
+        $logs += $applicationLogs
+        $logs += $systemLogs
+        $logs += $managementLogs
```
bill-long, line 49: The repetition here is bugging me. What about:

```powershell
foreach ($logName in @("Application", "System", "MSExchange Management")) {
  $regValue = Get-RemoteRegistryValue "$baseRegistryLocation$logName"
  $logs += ($null -ne $regValue ? $regValue : "$rootLogPath\$logName.evtx")
}
```

dpaulson45: would have to change it to an `if` statement not a `?`. Only available in PowerShell Core.

dpaulson45: how does this look now?

bill-long (approved):

Diagnostics/ExchangeLogCollector/RemoteScriptBlock/IO/Save-WindowsEventLogs.ps1
bill-long: Better.

Issue #1030, opened 2022-05-12
[Issue] VSSTester.ps1 does not check for updates when starting


worldsdream:
When using SetupAssist and HealthChecker script, they both do search for updates when starting up. Is it true that this does NOT happen with VSSTester.ps1? Or is there something wrong with my machine?

If so, can this be added, please? It should be added to all the scripts...

dpaulson45:
VSSTester.ps1 does not have the code to check for updates. We only have that for scripts that have frequent updates. Like the SetupAssist and HealthChecker.

worldsdream:
It would be great if all these scripts had such functionality and checked for updates. Perhaps one day, this can be added if this is not a big request to ask for.

dpaulson45:
It isn't a big ask, due to the way we wrote out the shared code to be easily added to scripts, but it likely wouldn't need to be added to all the scripts. That is where we would need to decide that, as it really isn't needed for all the scripts. 

For example, Clear-MailboxPermissions.ps1 or Find-AmbiguousSids.ps1 are pretty simple and easy scripts that haven't been modified since they have been added to the repository. They are also niche type scripts that you should only be using once in a long while. This would be similar to VSSTester, you should only be using it as a testing tool when you have backup issues. Typically, then we would suspect that the script would get removed or deleted after so long time has passed after resolving their issue. Then if they need to use the script again, they will want to download it again. 

Do you have a scenario in mind that could help explain why VSSTester for example would benefit from the update?

@bill-long  thoughts on this?

worldsdream:
VSSTester didn't work with non-English systems and an update was released for it. Every time the engineer was opening the file, he got an error and was troubleshooting why this happens. Without knowing that there was an update that fixes the bug.

This is just one simple reason. But a reason that the script did not work eventually which is a bummer at that time.

Most don't know that there is a Github page and that there are fixes pushed to these scripts.

bill-long:
There are a few interesting points here.

* Whether a script checks for updates is not really visible to the user unless there's an update. Maybe this isn't always the right approach. We might want to consider a parameter that makes the update check provide more feedback. "Latest version is nn.nn, no update needed", etc, just to indicate to the user that a check happened. We already have another issue opened about making auto update failures more visible, at least as an option, so we can tackle that at the same time.
* We might want to add additional behavior options to the shared auto-update code. Right now, your choices are update, or don't. But maybe for some scripts, we just want to tell the user that an update is available, and here's where they can download it.
[...]

PR #1052, opened 2022-05-17
Break up autoupdate code to improve flexibility

### LINKED ISSUE #1028: [Issue] - Health Checker - No way to see auto update failures
Currently, all auto update failures are silent, because we expect these to be common. However, we need a way to have -Verbose logging or something similar when a user does wish to identify and troubleshoot an auto-update failure.

bill-long:
Fixes #1028 and #1030.

Run HealthChecker with -Verbose switch to see update errors.

VSSTester now notifies of an update but leaves it to the user to download. We can always change it based on feedback, but for now this serves as an example of a less aggressive approach to the version check.


dpaulson45 (approved):

dpaulson45:
Actually, in this PR should we do the update for all the scripts as the issue provided?

bill-long:
> Actually, in this PR should we do the update for all the scripts as the issue provided?

Nah, we need to think more about which scripts should have which update behavior.

PR #1095, opened 2022-06-05
New Feature Request : IIS Http  Security Headers

KarincaMurat:
**Issue:**
The script is detecting TLS and more configuration about IIS,. There are no details for IIS Security Headers. 

**Reason:**
Microsoft announced some of articles about EXchange and IIS as below,
https://techcommunity.microsoft.com/t5/iis-support-blog/remove-unwanted-http-response-headers/ba-p/369710
But in the script there is no reviewing metric for http header details 

**Fix:**
could you please add reviewing http header settings for following setting as in best practices 
X Frame Options
--
X XSS Protection
--
X Content Type Options
--
X Permitted Cross Domain Policies
--
Strict Transport Security
--
Content Security Policy
--
Referrer Policy
--
Feature Policy
--
Expect CT
--

**Validation:**
N/A


dpaulson45:
Closing PR as there are no changes.

@bill-long we need to figure out how this PR was able to be created as I don't think contributors should be able to create a PR from branches within our own repository.

PR #1097, opened 2022-06-07
Update Test-ExchAVExclusions.ps1 to support Edge Server

iserrano76:
**Issue:**
If you run the script in an edge server, and error apeears

**Reason:**
Avoid errors in the execution

**Fix:**
Do not check database paths if it is an Edge server

**Validation:**
Provide if applicable

### DIFF
--- a/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
+++ b/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
@@ -91,10 +91,23 @@ $BaseFolders.Add((Join-Path $env:SystemDrive '\inetpub\temp\IIS Temporary Compre
 $BaseFolders.Add((Join-Path $env:SystemRoot '\Microsoft.NET\Framework64\v4.0.30319\Temporary ASP.NET Files').tolower())
 $BaseFolders.Add((Join-Path $env:SystemRoot '\System32\Inetsrv').tolower())
 
-# Add all database folder paths
-foreach ($Entry in (Get-MailboxDatabase -Server $Env:COMPUTERNAME)) {
-    $BaseFolders.Add((Split-Path $Entry.EdbFilePath -Parent).tolower())
-    $BaseFolders.Add(($Entry.LogFolderPath.pathname.tolower()))
+
+$CmdAvailable = Get-Command Get-ExchangeServer -ErrorAction SilentlyContinue
+if ( -not $CmdAvailable) {
+    try {
+        Add-PSSnapin Microsoft.Exchange.Management.PowerShell.SnapIn
+        # Checking if it is a mailbox server to avoid errors if it is an Edge server.
+        if ((Get-ExchangeServer $env:COMPUTERNAME).IsMailboxServer) {
+            # Add all database folder paths
+            foreach ($Entry in (Get-MailboxDatabase -Server $Env:COMPUTERNAME)) {
+                $BaseFolders.Add((Split-Path $Entry.EdbFilePath -Parent).tolower())
+                $BaseFolders.Add(($Entry.LogFolderPath.pathname.tolower()))
+            }
+        }
+    } catch {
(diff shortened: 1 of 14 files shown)


Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -91,10 +91,23 @@ $BaseFolders.Add((Join-Path $env:SystemDrive '\inetpub\temp\IIS Temporary Compre
...
 
-# Add all database folder paths
-foreach ($Entry in (Get-MailboxDatabase -Server $Env:COMPUTERNAME)) {
-    $BaseFolders.Add((Split-Path $Entry.EdbFilePath -Parent).tolower())
-    $BaseFolders.Add(($Entry.LogFolderPath.pathname.tolower()))
+
+$CmdAvailable = Get-Command Get-ExchangeServer -ErrorAction SilentlyContinue
+if ( -not $CmdAvailable) {
+    try {
+        Add-PSSnapin Microsoft.Exchange.Management.PowerShell.SnapIn
```
dpaulson45, line 98: We should not be directly adding in the PowerShell SnapIn like this, as we bypass RBAC. 

Use the `Confirm-ExchangeShell` from the Shared location.

dpaulson45 (changes requested):

dpaulson45 (approved):

iserrano76:
> Just noticed this actually, do we need the `Confirm-ExchangeShell`? The reason I ask is because if we aren't already in EMS when running the script, the `$env:ExchangeInstallPath` might not even be set yet.

We need Management shell to get the database paths if it is a mailbox server, but we are not checking if it is an Exchange server.
We could add a verification if env:ExchangeInstallPath exists and if not stop the script.

dpaulson45:
> > Just noticed this actually, do we need the `Confirm-ExchangeShell`? The reason I ask is because if we aren't already in EMS when running the script, the `$env:ExchangeInstallPath` might not even be set yet.
> 
> We need Management shell to get the database paths if it is a mailbox server, but we are not checking if it is an Exchange server. We could add a verification if env:ExchangeInstallPath exists and if not stop the script.

Need to move `Confirm-ExchangeShell` to prior to using `$env:ExchangeInstallPath` and then do the following:

```
$exchangeShell = Confirm-ExchangeShell -Identity $env:computerName
if (-not($exchangeShell.ShellLoaded)) {
Write-Warning "Failed to load Exchange Shell Module..."
exit
}

if ($null -ne $env:ExchangeInstallPath -and
(Test-Path $env:ExchangeInstallPath)) {
Write-Warning "Failed to find Exchange Install Path" 
}
```

I would be best to not use `$env:ExchangeInstallPath` and use the registry location, but that should be done in a different PR.

iserrano76 (commented):
Do you have any function to check the installation path in registry?

dpaulson45:
> Do you have any function to check the installation path in registry?

No, nothing in Shared. However, you can just move do what I provided above, then you can open up a different PR if you also want to change the `$env:ExchangeInstallPath` value.

dpaulson45:
@iserrano76 this is not what I provided. It shouldn't have been a second PR to merge the other changes that I was talking about. Part of this doesn't have anything to do with main point of this PR, which is to get this script to support with Edge Servers. 

Please remove the last commit you did and ONLY do what is provided in this comment: 

https://github.com/microsoft/CSS-Exchange/pull/1097#issuecomment-1148878545

dpaulson45:
This commit needs to be dropped 

e24f632b7528535f2934d9474063e23ec42aaf1c

Otherwise, we will have regressions occurring.

dpaulson45:
Thanks! Waiting on Bill to see if we want to merge this PR this way or have you rebase. There are a lot of commits with revert and everything like that which I don't care to see.

dpaulson45 (approved):

PR #1118, opened 2022-07-01
Added missing exclusions based on the role

iserrano76:
**Issue:**
Previous version do not include all paths includes in the following article:
https://docs.microsoft.com/en-us/Exchange/antispam-and-antimalware/windows-antivirus-software?view=exchserver-2019
Additionally, it does not differentiate the paths depending on the role as the article explain

**Reason:**
To have an accurate verification of all the paths included in the article

**Fix:**
Get the role by Get-ExchangeServer and define the folders to exclude with Get-FrontEndTransportService, Get-TransportService and Get-TransportService.

**Validation:**
Test it in lab.


Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -65,60 +65,152 @@ $LogFile = "ExchAvExclusions.log"
...
+
+# Check the installation path
+if (-not ( Test-Path $($serverExchangeInstallDirectory.MsiInstallPath) -PathType Container) ) {
+    Write-Warning "Failed to find the Exchage instalation Path"
+    exit
+}
+
+# Check Exchange is 2016 or 2019
+if ( -not ( $($serverExchangeInstallDirectory.MsiProductMajor) -eq 15 -and `
+        ( $($serverExchangeInstallDirectory.MsiProductMinor) -eq 1 -or $($serverExchangeInstallDirectory.MsiProductMinor) -eq 2 ) ) ) {
```
dpaulson45, line 84: Why can't this script be run against 2013? We still need to support that till April 2023.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -65,60 +65,152 @@ $LogFile = "ExchAvExclusions.log"
...
+
+    $BaseFolders.Add((Join-Path $ExchangePath '\Mailbox').tolower())
+
     # Add all database folder paths
     foreach ($Entry in (Get-MailboxDatabase -Server $Env:COMPUTERNAME)) {
         $BaseFolders.Add((Split-Path $Entry.EdbFilePath -Parent).tolower())
         $BaseFolders.Add(($Entry.LogFolderPath.pathname.tolower()))
     }
+
+    $fetsLogs = Get-FrontEndTransportService $($env:COMPUTERNAME) | Select-Object ConnectivityLogPath, `
```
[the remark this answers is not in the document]
iserrano76: This script was based in the following link from the comments:
https://docs.microsoft.com/en-us/Exchange/antispam-and-antimalware/windows-antivirus-software?view=exchserver-2019
And it is only for 2016/2019.
The doc for 2013 is:
https://docs.microsoft.com/en-us/exchange/anti-virus-software-in-the-operating-system-on-exchange-servers-exchange-2013-help
and it has many differences. That was the main reason because I thought it is better to control the version to 2016/9.
Because the original is based on the 2016/2019, I think is better to keep on that way and in a new version handle all the 2013 paths.
What do you think?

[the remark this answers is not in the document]
iserrano76: I do not test it because the documentation on the comments was for 2016/2019 but looking at 2013 doc, it should work partially but will not verify some paths. 
For example, 2013 has client access role with some specific paths that are not controlled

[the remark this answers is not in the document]
dpaulson45: If right now the script does work with 2013, then we need to keep the support for 2013. We don't want it removed. till we have the additional logic in place for 2013 as well.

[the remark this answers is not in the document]
iserrano76: if we remove:
if ( -not ( $($serverExchangeInstallDirectory.MsiProductMajor) -eq 15 -and `
        ( $($serverExchangeInstallDirectory.MsiProductMinor) -eq 1 -or $($serverExchangeInstallDirectory.MsiProductMinor) -eq 2 ) ) ) {

it will work in the same way that before the modification with 2013, in fact I think we will include more exclusion that before because the original version does not include several paths 
I just included the control of the version to be accurate with the documentation that the script follow.

In any case, I am working in a new version that works with Exchange 2013.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -65,57 +65,181 @@ $LogFile = "ExchAvExclusions.log"
...
+    exit
+}
+
+# Check the installation path
+if (-not ( Test-Path $($serverExchangeInstallDirectory.MsiInstallPath) -PathType Container) ) {
+    Write-Warning "Failed to find the Exchage instalation Path"
+    exit
+}
+
+# Check Exchange is 2016, 2016 or 2019
```
dpaulson45, line 82:
```suggestion
# Check Exchange is 2013, 2016 or 2019
```

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -65,57 +65,181 @@ $LogFile = "ExchAvExclusions.log"
...
+        }
+    }
+}
+
+if ((Get-ExchangeServer $env:COMPUTERNAME).IsUnifiedMessagingServer) {
+    $BaseFolders.Add((Join-Path $ExchangePath '\UnifiedMessaging\Grammars'))
+    $BaseFolders.Add((Join-Path $ExchangePath '\UnifiedMessaging\Prompts'))
+    $BaseFolders.Add((Join-Path $ExchangePath '\UnifiedMessaging\Temp'))
+    $BaseFolders.Add((Join-Path $ExchangePath '\UnifiedMessaging\Voicemail'))
+}
```
dpaulson45, line 174: Don't believe these exist in Exchange 2019.

dpaulson45 (changes requested):

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
iserrano76: it does not and it is not a IsUnifiedMessagingServer it is not going to go into, but I think it exists as a property

dpaulson45: I actually missed the `IsUnifiedMessagingServer` piece

dpaulson45 (approved):

PR #1128, opened 2022-07-08
Forced eicar detection

iserrano76:
**Issue:**
If we just create the Eicar file, could be possible that in only one
minute is not detected by the AV.
Added the execution of the Eicar to force detection and extend the time
from one minutes to 5 as the comment said.

**Reason:**
Avoid to give correct configuration when the AV has not time to detect eicar

**Fix:**
Force the execution of the eicar and expand the time to detect.

**Validation:**
Tested in lab

### DIFF
--- a/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
+++ b/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
@@ -139,9 +139,11 @@ foreach ($path in $BaseFolders) {
 Write-SimpleLogfile -String "Creating EICAR Files" -name $LogFile -OutHost
 
 # Create the EICAR file in each path
+$EicarFileName = "eicar.bat"
+
 foreach ($Folder in $FolderList) {
 
-    [string] $FilePath = (Join-Path $Folder eicar.bat)
+    [string] $FilePath = (Join-Path $Folder $EicarFileName)
     Write-SimpleLogfile -String ("Creating EICAR file " + $FilePath) -name $LogFile
 
     #Base64 of Eicar string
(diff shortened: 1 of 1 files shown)


Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -141,7 +141,7 @@ Write-SimpleLogfile -String "Creating EICAR Files" -name $LogFile -OutHost
 # Create the EICAR file in each path
 foreach ($Folder in $FolderList) {
 
-    [string] $FilePath = (Join-Path $Folder eicar.com)
+    [string] $FilePath = (Join-Path $Folder eicar.bat)
```
dpaulson45, line 144: Change to File name being a variable. Do this outside of the loop. 

$EicarFileName = "eicar.com" 


Secondly, why are we changing the file to be .bat vs .com? Most articles I have found do .com or .com.txt not .bat.

dpaulson45: seen a few Microsoft docs that provide that we can do this with a txt file as well. That might be even a better option?

iserrano76: I agree with you, and I tested before, but the issue is that we have the txt extension in the exclusion list and if AV was correctly configured is not going to be detected:
https://docs.microsoft.com/en-us/Exchange/antispam-and-antimalware/windows-antivirus-software?view=exchserver-2019#file-name-extension-exclusions

dpaulson45: Is that the point? If it is correctly configured, to not get picked up?

iserrano76: but in that case we can not be sure if it was for the path or for the extension and this script only test the paths

dpaulson45: If that is the case, then we should drop multiple file extensions to the path. Do that based off a couple of the extensions that we commonly need and then one we don't.

iserrano76: I was thinking to create another function that return the extensions that we use, because it depends on the role, and we can cover all the extensions correctly.
Just to extend this script to extensions and not only paths.
But to be honest I thought it is better in another PR.
Let me know your thoughts.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -194,7 +196,7 @@ foreach ($Folder in $FolderList) {
     }
     # If the file doesn't exist Add that to the bad folder list -- means the folder is being scanned
     else {
-        Write-SimpleLogfile -String ("[FAIL] - Possible AV Scanning: " + $FilePath) -name $LogFile -OutHost
+        Write-SimpleLogfile -String ("[FAIL] - Possible AV Scanning: " + $Folder) -name $LogFile -OutHost
```
dpaulson45, line 199: Shouldn't this be the full path?

iserrano76: the fullpath shows:
[7/11/2022 11:55:50 AM] - [FAIL] - Possible AV Scanning: c:\program files\microsoft\exchange server\v15\fip-fs\eicar.bat
And folder shows:
[7/11/2022 12:09:57 PM] - [FAIL] - Possible AV Scanning: c:\program files\microsoft\exchange server\v15\fip-fs

I think it is better to show only the folder that is not configured and not including the file that we are using to test it because it doesn't care to the user.

PR #1177, opened 2022-08-17
Fix for Extended Protection Management Server usage and preparation work for PSSession usage

lusassl-msft:
**Issue:**
The script fails if it was executed on a management server in the following scenarios:

- Executed via EMS on a management server
- Executed via Exchange PowerShell session (`New-PSSession -ConfigurationName Microsoft.Exchange -ConnectionUri http://e2k16-1/PowerShell | Import-PSSession`)

**Reason:**
- Scenario 1) wasn't properly handled in the `ExchangeExtendedProtectionManagement.ps1`.
- Scenario 2) wasn't properly handled in the `Confirm-ExchangeShell.ps1`

**Fix:**
- Scenario 1): `IgnoreToolsIdentity $true` was added
- Scenario 2): Logic added to `Confirm-ExchangeShell.ps1` to check for an active Exchange PowerShell session

We have not yet validated scenario 2) in the `ExchangeExtendedProtectionManagement.ps1` script. Therefore, it's not yet supported and may be added at a later date.

**Validation:**
Lab

### DIFF
--- a/Security/src/ExchangeExtendedProtectionManagement/ExchangeExtendedProtectionManagement.ps1
+++ b/Security/src/ExchangeExtendedProtectionManagement/ExchangeExtendedProtectionManagement.ps1
@@ -84,7 +84,7 @@ begin {
             $Script:SkipEWS = $false
         }
 
-        if (-not((Confirm-ExchangeShell -Identity $env:COMPUTERNAME).ShellLoaded)) {
+        if (-not((Confirm-ExchangeShell -Identity $env:COMPUTERNAME -IgnoreToolsIdentity $true).ShellLoaded)) {
             Write-Warning "Failed to load the Exchange Management Shell. Start the script using the Exchange Management Shell."
             exit
         }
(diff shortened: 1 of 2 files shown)


dpaulson45:
I don't understand what we are trying to address in `Confirm-ExchangeShell` here in this PR. 

image


Realistically, shouldn't even need to use `$IgnoreToolsIdentity`, we should just detect if we are on a tools box or not and decide to ignore it or not.

dpaulson45:
Based off the commit dc4b90df02f12655c6a1d665e9f4382bc774b982, looks like it was a bad use of a parameter. 

We should always Ignore the Identity if what is passed is what is on a tools box name or the name of the computer that we are on for Remote Shell.

dpaulson45:
@lusassl-msft checkout #1178 as this should handle everything that we need. @bill-long you can confirm as well.

lusassl-msft:
Will be worked on in #1178

PR #1178, opened 2022-08-18
Fix Confirm-ExchangeShell for handling on Remote Shell and Tool boxes

dpaulson45:
**Issue:**
By default, we don't always pass the Identity of an Exchange Server to `Confirm-ExchangeShell` because the script doesn't require it to target a single Exchange server. Because of this, we want to be able to support running on a tools box or a remote shell session by default within `Confirm-ExchangeShell`

**Fix:**
By default, we should be able to support `Confirm-ExchangeShell` when on a non-Exchange server and we pass the Identity of the local computer. If the Identity passed is the local computer name and is a non-Exchange servers, remove the Identity from the mix. 

**Validation:**
Lab tested. 

Scenario 1: 

- Single session exists
- Session uses short name
- Session is Imported
- From tools box
- Passed Identity with None, FQDN for an Exchange Server, FQDN for local server

Results: Expected

image

Scenario 2: 

- From tools box
- Running from PowerShell before session is created 

Results: Load Shell and Removed Identity

image

Scenario 3:

- Single session exists 
- Session uses FQDN 
- Session is Imported
- On tools box

Result: Uses the current session as expected 
image

Scenario 4: 

- No session is used or active
- On Remote Computer

Result: Failed as expected

image

Scenario 5:

- Single session exists
- Session uses short name
- Session is Imported
- On Remote Computer

Result: Works as expected

image

### DIFF
--- a/Shared/Confirm-ExchangeShell.ps1
+++ b/Shared/Confirm-ExchangeShell.ps1
@@ -7,15 +7,12 @@
 function Confirm-ExchangeShell {
     [CmdletBinding()]
     param(
-        [Parameter(Mandatory = $true)]
+        [Parameter(Mandatory = $false)]
         [ValidateNotNullOrEmpty()]
-        [string]$Identity,
+        [string]$Identity = $env:COMPUTERNAME,
 
         [Parameter(Mandatory = $false)]
         [bool]$LoadExchangeShell = $true,
-
-        [Parameter(Mandatory = $false)]
-        [bool]$IgnoreToolsIdentity = $false,
 
         [Parameter(Mandatory = $false)]
         [scriptblock]$CatchActionFunction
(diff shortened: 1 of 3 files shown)


lusassl-msft:
Looks good so far. Will do some tests. However, as suggested by @bill-long , we should add a parameter to explicitly block PSSession usage (as the returned objects can be different vs. EMS). As PSSession was not tested for `ExchangeExtendedProtectionManagement.ps1`, it should be blocked in this script by the help of a parameter.

dpaulson45:
> Looks good so far. Will do some tests. However, as suggested by @bill-long , we should add a parameter to explicitly block PSSession usage (as the PSSession returned objects can be different vs. EMS). As PSSession was not tested for `ExchangeExtendedProtectionManagement.ps1`, it should be blocked in this script by the help of a parameter.


This can be done in a different parameter, and is the responsibility of the caller, in my opinion. We provide the results stating if this is a Remote Shell or not. The caller should block it there.

bill-long:
None of the above tests were in EMS, but they show different results for RemoteShell. The last one shows True and the others show False. It doesn't seem like RemoteShell is a reliable indicator of whether we have the types loaded.

bill-long:
After our voice discussion, I think we landed on "this is fine for now". Approving.

bill-long (dismissed):

Shared/Confirm-ExchangeShell.ps1
```diff
@@ -102,7 +102,9 @@ function Confirm-ExchangeShell {
 
                     Write-Verbose "Imported Module. Trying Get-Exchange Server Again"
                     try {
-                        Get-ExchangeServer @params | Out-Null
+                        $isEMS = Get-ExchangeServer @params |
+                            Select-Object -First 1 |
+                            ForEach-Object { if ($_.GetType().Name -eq "ExchangeServer") { return $true } return $false }
```
bill-long, line 107: Isn't this an expensive way to do this due to the number of Exchange servers? What if we did something like Get-OrganizationConfig instead?

bill-long (approved):

PR #1179, opened 2022-08-18
Add Write-Progess to display the script is actively working

dpaulson45:
**Issue:**
The script at times will stop displaying things to screen but is actively working.

**Fix:**
Use `Write-Progress` to show that the script is still working and doing things when we don't display everything to the host. 

**Validation:**
Lab tested 

image


Security/src/ExchangeExtendedProtectionManagement/ExchangeExtendedProtectionManagement.ps1
```diff
@@ -323,11 +323,12 @@ begin {
                     $totalCount = $ExchangeServers.Count
                     $progressParams = @{
                         Activity        = "Prerequisites Check"
-                        Status          = [string]::Empty
+                        Status          = "Running Get-OutlookAnywhere...."
```
bill-long, line 326:
```suggestion
                        Status          = "Running Get-OutlookAnywhere"
```

bill-long (changes requested):

Security/src/ExchangeExtendedProtectionManagement/ExchangeExtendedProtectionManagement.ps1
dpaulson45: Don't like the `....` to show that it might take a while?

bill-long: Adding dots doesn't really communicate that. It just looks bad. If you want to tell the user it might take a while, actually say that in the text.

bill-long (changes requested):
Since we have two progress bars, and the second one is child progress of the first, we should give the first one an Id and then specify it as the ParentId on the child.

bill-long:
Have we considered doing something like this so it shows progress during the long operation:

image

dpaulson45:
> Have we considered doing something like this so it shows progress during the long operation:
> 
> image

Hmmm... let me try something

dpaulson45:
> Since we have two progress bars, and the second one is child progress of the first, we should give the first one an Id and then specify it as the ParentId on the child.

Done.

image

bill-long:
> Hmmm... let me try something

If the foreach emits the objects you can still have the same statement collect the results:

image

dpaulson45:
> > Hmmm... let me try something
> 
> If the foreach emits the objects you can still have the same statement collect the results:
> 
> image

got it to look like this:

image

bill-long (approved):

PR #1213, opened 2022-09-01
Mermaid flowchart diagram is broken

lusassl-msft:
**Issue:**
The mermaid flowchart which describes the SSL Offloading and SSL Bridging workflow was broken (Extended Protection documentation).

### DIFF
--- a/docs/Security/Extended-Protection.md
+++ b/docs/Security/Extended-Protection.md
@@ -26,21 +26,21 @@ While Extended Protection can be enabled on each virtual directory manually, we 
 
 Example:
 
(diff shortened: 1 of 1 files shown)


lusassl-msft:
Looks like mermaid isn't supported by default by MkDocs. I think we need a plugin like this to make it work: https://github.com/fralau/mkdocs-mermaid2-plugin

lusassl-msft:
@bill-long can we make this work in MkDocs by using the plugin mentioned or so?

bill-long (approved):

PR #1246, opened 2022-09-20
Fix to properly detect Azure certificates issued by MS-Organization-P2P-Access [YYYY] 

### LINKED ISSUE #1245: HealthChecker doesn't skip the MS-Organization-P2P-Access [YYYY] Azure certificate
**Describe the issue:**
The `MS-Organization-P2P-Access [YYYY]` certificate is still returned by HealthChecker (it should be skipped due to its short life cycle).

Originally addressed in #1119 

**Possible root cause:**
I guess the certificate is not self-signed (that's what we are currently looking for). Need to confirm with finder and if it's not self-signed, switch the flag to IsSelfSigned $false to properly detect and exclude the certificate.

More context:
Self-signed should be the Root certificate in this case (Subject & Issuer: `MS-Organization-P2P-Access [YYYY]`). However, in this case, the certificates with the short life cycle have a GUID as subject and Issuer is `MS-Organization-P2P-Access [YYYY]`). So they are not self-signed.

lusassl-msft:
**Issue:**
The certificate with the short life cycle of 24 hours wasn't properly excluded.
Resolve #1245 

**Reason:**
It's not a self-signed certificate. It's issued by the `MS-Organization-P2P-Access [YYYY]` CA

**Fix:**
Flipped isSelfSigned from `$true` to `$false`

### DIFF
--- a/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeServerCertificates.ps1
+++ b/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeServerCertificates.ps1
@@ -59,7 +59,7 @@ function Get-ExchangeServerCertificates {
         # These certificates can be deleted. The Azure VM Agent re-creates certificates if needed.
         # https://docs.microsoft.com/azure/virtual-machines/extensions/features-windows
         $certificatesToExclude = @(
-            NewCertificateExclusionEntry "CN=MS-Organization-P2P-Access \[[12][0-9]{3}\]$" $true
+            NewCertificateExclusionEntry "CN=MS-Organization-P2P-Access \[[12][0-9]{3}\]$" $false
             NewCertificateExclusionEntry "DC=Windows Azure CRP Certificate Generator" $true
         )
         Write-Verbose "Trying to receive certificates from Exchange server: $($Script:Server)"


Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeServerCertificates.ps1
```diff
@@ -59,7 +59,7 @@ function Get-ExchangeServerCertificates {
         # These certificates can be deleted. The Azure VM Agent re-creates certificates if needed.
         # https://docs.microsoft.com/azure/virtual-machines/extensions/features-windows
         $certificatesToExclude = @(
-            NewCertificateExclusionEntry "CN=MS-Organization-P2P-Access \[[12][0-9]{3}\]$" $true
+            NewCertificateExclusionEntry "CN=MS-Organization-P2P-Access \[[12][0-9]{3}\]$" $false
```
dpaulson45, line 62: Did this all of a sudden change?

lusassl-msft: I assume that it has not yet worked. Just found the certificate on my local machine and it is indeed not self-signed. The fix should do the job of properly detecting and excluding the certificate.

dpaulson45 (approved):

PR #1254, opened 2022-09-26
Return TLS cipher suites on OS older than Windows Server 2016

### LINKED ISSUE #1141: [Work Item] HealthChecker - Make the TLS Cipher Suites check available on Windows < 2016
**Describe the work**
We should check if it's possible to query the TLS Cipher Suites information from the registry (on machines where the `Get-TLSCipherSuite` cmdlet isn't available).

Registry key: `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Cryptography\Configuration\Local\SSL\00010002`
Value: `Functions`

#849

lusassl-msft:
**Issue:**
`Get-TlsCipherSuite` isn't available on OS older than Windows Server 2016 and so, no TLS cipher suites were displayed. This PR adds an improvement to query the cipher suites in these cases directly via registry. It also handles cases in which the TLS cipher suite order was configured via GPO and so takes precedence over the local configuration.

**Fix:**
Resolve #1141 
Query TLS cipher suites on legacy OS via registry call.

**Validation:**
Lab / Unit test


Shared/TLS/Get-TlsCipherSuiteInformation.ps1
```diff
@@ -0,0 +1,62 @@
...
+    begin {
+        Write-Verbose "Calling: $($MyInvocation.MyCommand)"
+    }
+    process {
+        try {
+            # 'Get-TlsCipherSuite' takes account of the cipher suites which are configured by the help of GPO.
+            # No need to query the ciphers defined via GPO if this call is successful.
+            Write-Verbose "Trying to query TlsCipherSuites via 'Get-TlsCipherSuite'"
+            $tlsCipherSuites = Get-TlsCipherSuite -ErrorAction Stop
+        } catch {
```
dpaulson45, line 21: This isn't getting the information remotely. This needs to be addressed.

Shared/TLS/Get-TlsCipherSuiteInformation.ps1
```diff
@@ -0,0 +1,62 @@
...
+                    ValueType           = "MultiString"
+                    CatchActionFunction = $CatchActionFunction
+                }
+
+                $tlsCipherSuites = Get-RemoteRegistryValue @tlsRegistryParams
+            }
+        }
+    }
+    end {
+        return $tlsCipherSuites
```
dpaulson45, line 60: Should maybe think about returning a standard object that works for both `Get-TlsCipherSuite` and the values from the registry.

Shared/TLS/Get-TlsCipherSuiteInformation.ps1
```diff
@@ -0,0 +1,62 @@
...
+            $tlsCipherSuites = Get-TlsCipherSuite -ErrorAction Stop
+        } catch {
+            # If we can't get the ciphers via cmdlet, we need to query them via registry call and need to check
+            # if ciphers suites are defined via GPO as well. If there are some, these take precedence over what
+            # is in the default location.
+            Write-Verbose "Failed to query TlsCipherSuites via 'Get-TlsCipherSuite' fallback to registry"
+
+            if ($null -ne $CatchActionFunction) {
+                & $CatchActionFunction
+            }
```
dpaulson45, line 29: Should use `Invoke-CatchActionError` from share

dpaulson45 (changes requested):

Shared/TLS/Get-TlsCipherSuiteInformation.ps1
lusassl-msft: Oops, you're right. Good catch.

Shared/TLS/Get-TlsCipherSuiteInformation.ps1
lusassl-msft: Removed this part as I changed the logic.

Shared/TLS/Get-TlsCipherSuiteInformation.ps1
lusassl-msft: We now return a standard object with the information that we need. Additional properties can be added on a need to use base.

Shared/TLS/Get-TlsCipherSuiteInformation.ps1
lusassl-msft: This call is now performed via `Invoke-ScriptBlockHandler`

dpaulson45 (approved):

PR #1274, opened 2022-10-06
Match the EEMS rule, which does not have a space

bill-long:
When a server has the mitigation automatically applied by EEMS, there is no space in the rule. Therefore, the scripts conclude that the server is not properly mitigated.


diganr (approved):
Approved

PR #1288, opened 2022-10-12
Fix false/positive in proxy detection

### LINKED ISSUE #1091: [HealthChecker] - Improve HTTP Proxy check
**Describe the improvement**
Received feedback from multiple customers regarding the HTTP proxy check. They stated that the check is inaccurate in some scenarios:

- WinHTTP proxy bypass list has wildcard entries. Wildcards are [supported here](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ms761351(v=vs.85)), but not via Set-ExchangeServer.
- Proxy server set via [Set-ExchangeServer](https://docs.microsoft.com/en-us/powershell/module/exchange/set-exchangeserver?view=exchange-ps) can be set with or without the protocol `https://` or `http://`. This fact also causes a warning because the proxy used by WinHTTP must be set without the protocol part.

Our documentation regarding this topic is also inconsistent: 

[Set-ExchangeServer example with protocol part](https://docs.microsoft.com/bg-bg/exchange/troubleshoot/hybrid-configuration-wizard-errors/the-underlying-connection-was-closed-error#resolution)
[Set-ExchangeServer example without protocol part](https://docs.microsoft.com/en-us/exchange/exchange-emergency-mitigation-service?view=exchserver-2019#connectivity)

**Possible solution**
- Strip the protocol part from the proxy server returned via `Get-ExchangeServer` OR
- Use `Authority` instead of `AbsoluteUri`: 
image

- Check for wildcard entries in WinHttp Proxy Bypass list (`[System.Management.Automation.WildcardPattern]::ContainsWildcardCharacters()`) and consider this when comparing the entries against the bypass list returned via `Get-ExchangeServer`.  --> Looks like we don't compare the `ByPassList` yet. This might be useful if something similar is planned in the future.

Also, exclude the proxy check for Edge Transport servers as there is no need to have a proxy server configured via `Set-ExchangeServer` (FIP-FS uses a different proxy configuration, EEMS isn't available, no client connections...)

lusassl-msft:
**Issue:**
We show a warning if the proxy server which has been set via `Set-ExchangeServer` contains the protocol part (http or https). Proxy server set via WinHTTP is always without protocol part. At the moment, we're flagging this as a proxy mismatch between Exchange and WinHTTP.

**Reason:**
This fix makes sure that we compare both values without using the protocol part. We achieve this by using the `Authority` property which contains the proxy server without the protocol part. We also exclude the proxy check from servers who are running the Edge Transport role as there is no need to set a proxy server via `Set-ExchangeServer` on these machines.

**Fix:**
Resolve #1091 

**Validation:**
Lab (E15 + E16)

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerOsInformation.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerOsInformation.ps1
@@ -254,7 +254,8 @@ function Invoke-AnalyzerOsInformation {
     $displayWriteType = "Grey"
     $displayValue = $osInformation.NetworkInformation.HttpProxy.ProxyAddress
 
-    if ($osInformation.NetworkInformation.HttpProxy.ProxyAddress -ne "None") {
+    if (($osInformation.NetworkInformation.HttpProxy.ProxyAddress -ne "None") -and
+        ($exchangeInformation.BuildInformation.ServerRole -ne [HealthChecker.ExchangeServerRole]::Edge)) {
         $displayValue = "$($osInformation.NetworkInformation.HttpProxy.ProxyAddress) --- Warning this can cause client connectivity issues."
         $displayWriteType = "Yellow"
     }
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerOsInformation.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerOsInformation.ps1
@@ -276,8 +276,9 @@ function Invoke-AnalyzerOsInformation {
         Add-AnalyzedResultInformation @params
     }
 
-    if ($osInformation.NetworkInformation.HttpProxy.ProxyAddress -ne "None" -and
-        $osInformation.NetworkInformation.HttpProxy.ProxyAddress -ne $exchangeInformation.GetExchangeServer.InternetWebProxy) {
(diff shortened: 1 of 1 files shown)


Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerOsInformation.ps1
```diff
@@ -276,8 +276,9 @@ function Invoke-AnalyzerOsInformation {
         Add-AnalyzedResultInformation @params
     }
 
-    if ($osInformation.NetworkInformation.HttpProxy.ProxyAddress -ne "None" -and
-        $osInformation.NetworkInformation.HttpProxy.ProxyAddress -ne $exchangeInformation.GetExchangeServer.InternetWebProxy) {
+    if (($osInformation.NetworkInformation.HttpProxy.ProxyAddress -ne "None") -and
+        ($exchangeInformation.BuildInformation.ServerRole -ne [HealthChecker.ExchangeServerRole]::Edge) -and
+        ($osInformation.NetworkInformation.HttpProxy.ProxyAddress -ne $exchangeInformation.GetExchangeServer.InternetWebProxy.Authority)) {
```
dpaulson45, line 281: Does this work when from a tools box that is on a different build of Exchange? 

Like when `Get-ExchangeServer` is run, the `AdminDisplayVersion` is a `string` type not an object type.

PR #1306, opened 2022-10-28
Error handling improvements for Exchange Log Collector

dpaulson45:
**Issue:**
Discovered an issue within Exchange Log Collector when trying to collect the Server Information remotely. When doing `Invoke-Command` it would return an error due to the `reg.exe` failing to export out the path `HKEY_LOCAL_MACHINE\\SOFTWARE\Microsoft\ExchangeServer`. This occurred because of the double slash. This was difficult to determine the cause from the debug logs by default. 

**Reason:**
Due to the problem debugging the issue quickly, ended up improving the error output to handle the display of information better. 

**Fix:**
Removed the slash `\` that was causing the failure. 

**Validation:**
Lab tested

### DIFF
--- a/Diagnostics/ExchangeLogCollector/ExchangeLogCollector.ps1
+++ b/Diagnostics/ExchangeLogCollector/ExchangeLogCollector.ps1
@@ -222,7 +222,7 @@ function Main {
         try {
             Invoke-Command -ComputerName $Script:ValidServers -ScriptBlock ${Function:Invoke-RemoteFunctions} -ArgumentList $argumentList -ErrorAction Stop
         } catch {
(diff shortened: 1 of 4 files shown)


bill-long:
Note the change to Write-ErrorInformation will generate a new release of the following scripts:

ExchangeExtendedProtectionManagement.ps1
HealthChecker.ps1
SetUnifiedContentPath.ps1
SetupAssist.ps1
SetupLogReviewer.ps1
Troubleshoot-ModernSearch.ps1
ExchangeLogCollector.ps1

bill-long (approved):

PR #1309, opened 2022-10-31
TLS 1.0 and 1.1 display output adjusted to recommend TLS 1.2 usage only

### LINKED ISSUE #1301: [Issue] - Health Checker doesn't warn about the risk of TLS 1.0 and 1.1
When TLS 1.0 and 1.1 are enabled and the registry is set as expected, HealthChecker shows those settings as green. Example:

image

However, these versions of TLS are no longer considered secure. So, while the settings are correct and will work, we should also indicate that this is _not_ secure and should be disabled.

lusassl-msft:
**Description:**
We (Microsoft) recommend proactively disabling TLS versions older than TLS 1.2 to prevent future protocol downgrade attacks. Note that Microsoft's TLS 1.0 implementation is free of known security vulnerabilities.
See: 
https://www.microsoft.com/en-us/security/blog/2017/06/20/tls-1-2-support-at-microsoft/
https://learn.microsoft.com/en-us/security/engineering/solving-tls1-problem

**Fix:**
Resolve #1301 

**Validation:**
Lab

image
image

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecuritySettings.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecuritySettings.ps1
@@ -51,7 +51,7 @@ function Invoke-AnalyzerSecuritySettings {
     $tlsSettings = $osInformation.TLSSettings.Registry.TLS
     $misconfiguredClientServerSettings = ($tlsSettings.Values | Where-Object { $_.TLSMisconfigured -eq $true }).Count -ne 0
     $displayLinkToDocsPage = ($tlsSettings.Values | Where-Object { $_.TLSConfiguration -ne "Enabled" -and $_.TLSConfiguration -ne "Disabled" }).Count -ne 0
-    $lowerTlsVersionDisabled = ($tlsSettings.Values | Where-Object { $_.TLSVersionDisabled -eq $true -and $_.TLSVersion -ne "1.2" }).Count -ne 0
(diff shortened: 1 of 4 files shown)


Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecuritySettings.ps1
```diff
@@ -79,11 +79,17 @@ function Invoke-AnalyzerSecuritySettings {
...
+        # TLS 1.0 or 1.1 being Enabled is Yellow as we recommend to disable this weak protocol versions
         if (($currentTlsVersion.TLSConfiguration -eq "Misconfigured" -or
                 $currentTlsVersion.TLSConfiguration -eq "Half Disabled") -or
                 ($tlsKey -eq "1.2" -and $currentTlsVersion.TLSConfiguration -eq "Disabled") -or
                 ($tlsKey -eq "1.3" -and $currentTlsVersion.TLSConfiguration -eq "Enabled")) {
             $displayWriteType = "Red"
+        } elseif ((($tlsKey -eq "1.0") -and
+            ($currentTlsVersion.TLSConfiguration -eq "Enabled")) -or
+            (($tlsKey -eq "1.1") -and
+            ($currentTlsVersion.TLSConfiguration -eq "Enabled"))) {
```
dpaulson45, line 91:
```suggestion
        } elseif ($currentTlsVersion.TLSConfiguration -eq "Enabled" -and
            ($tlsKey -eq "1.1" -or $tlsKey -eq "1.0")) {
```
This `if` statement can be simplified.

dpaulson45 (changes requested):

dpaulson45 (approved):

PR #1312, opened 2022-11-02
Exclude non-Exchange websites by default

### LINKED ISSUE #1310: [Issue] - Health Checker - Warning - Missing Configuration File
**Provide Version Number**
22.10.28.0217

**Describe the issue**
Wrong Warning - Missing Configuration File
Missing: D:\Programme\Sophos\PureMessage\EndUserWebsite\web.config

**Expected behavior**
A message is reported that a file is missing from the Sophos folder. The description reports that the file is checked in the two folders.

C:\Program Files\Microsoft\Exchange Server\V15\FrontEnd\HttpProxy
C:\Program Files\Microsoft\Exchange Server\V15\ClientAccess

**Additional context**
We use Sophos on the server.

lusassl-msft:
**Issue:**
The change introduced in PR #1286 will cause false positives (FP) if any other website than the Exchange default ones exists (e.g., added by 3rd party solutions). 

**Reason:**
We return all existing websites when `Get-WebSite` is being called and check for existing `web.config` files.

**Fix:**
We query the Exchange websites from AD configuration partition to ensure that we get all the Exchange-related websites (also additional created OWA/ECP ones). We skip any other website which is not related to Exchange.

image

Resolve #1310 

**Validation:**
Lab & affected customer

### DIFF
--- a/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
+++ b/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
@@ -2,9 +2,22 @@
 # Licensed under the MIT License.
 
 function Get-IISWebSite {
-    $webSites = Get-WebSite
-    $bindings = Get-WebBinding
+    param(
+        [array]$WebSitesToProcess
+    )
+
     $returnList = New-Object 'System.Collections.Generic.List[object]'
+    $webSites = New-Object 'System.Collections.Generic.List[object]'
+
+    if ($null -eq $WebSitesToProcess) {
+        $webSites.AddRange((Get-WebSite))
+    } else {
+        foreach ($iisWebSite in $WebSitesToProcess) {
+            $webSites.Add((Get-WebSite -Name $($iisWebSite)))
+        }
+    }
+
+    $bindings = Get-WebBinding
 
     foreach ($site in $webSites) {
         $siteBindings = $bindings |
--- a/Shared/ActiveDirectoryFunctions/Get-ExchangeProtocolContainer.ps1
+++ b/Shared/ActiveDirectoryFunctions/Get-ExchangeProtocolContainer.ps1
@@ -1,0 +1,19 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+. $PSScriptRoot\Get-OrganizationContainer.ps1
+
+function Get-ExchangeProtocolContainer {
+    [CmdletBinding()]
+    [OutputType([System.DirectoryServices.DirectoryEntry])]
+    param (
+        [string]$ComputerName = $env:COMPUTERNAME
+    )
+
+    $ComputerName = $ComputerName.Split(".")[0]
+
+    $organizationContainer = Get-OrganizationContainer
+    $protocolContainerPath = ("CN=Protocols,CN=" + $ComputerName + ",CN=Servers,CN=Exchange Administrative Group (FYDIBOHF23SPDLT),CN=Administrative Groups," + $organizationContainer.distinguishedName)
+    $protocolContainer = [ADSI]("LDAP://" + $protocolContainerPath)
+    return $protocolContainer
+}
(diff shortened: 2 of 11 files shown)


dpaulson45:
Not sure I like this approach at the moment. 

Do you know of customers that have 3rd sites for split OWA or other Exchange virtual directories? I am curious to know if there is another way to filter this out.

lusassl-msft:
Nope, I have not heard from customers with a 3rd site for OWA. We could do a `Get-WebServiceVirtualDirectory` and `Get-WebServiceVirtualDirectory -ShowMailboxVirtualDirectories` call to determine the website for front and back end. Then we can do another call for OWA and ECP (as those are the common services for which customers sometimes creating additional websites) and check if the front end website is different to what we have. If so, add the 3rd site to the list and then exclude everything else.

lusassl-msft:
@dpaulson45 made a few more adjustements to auto-detect the Exchange-related websites. Will continue to work on this tomorrow (test data and unit tests).

lusassl-msft:
Might also be possible to collect the website information from the applicationHost.config file. Have to investigate this as well.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
```diff
@@ -2,9 +2,22 @@
...
-    $bindings = Get-WebBinding
+    param(
+        [array]$WebSitesToProcess
+    )
+
     $returnList = New-Object 'System.Collections.Generic.List[object]'
+    $webSites = New-Object 'System.Collections.Generic.List[object]'
+
+    if ($null -eq $WebSitesToProcess) {
+        $webSites.Add((Get-WebSite))
```
dpaulson45, line 13: This can't be done this way. 

image

`AddRange` must be used. 

image

Shared/ActiveDirectoryFunctions/Get-ExchangeProtocolContainer.ps1
```diff
@@ -0,0 +1,22 @@
...
+    [CmdletBinding()]
+    [OutputType([System.DirectoryServices.DirectoryEntry])]
+    param (
+        [string]$ComputerName = $env:COMPUTERNAME
+    )
+
+    [array]$ComputerName = $ComputerName.Split(".")
+    if ($ComputerName.Count -gt 1) {
+        $ComputerName = $ComputerName[0]
+    }
```
dpaulson45, line 16: This isn't really needed, you can just do the following. 

```suggestion
    $ComputerName = $ComputerName.Split(".")[0]
```

As this will always work. 

image

Shared/ActiveDirectoryFunctions/Get-ExchangeWebSitesFromAd.ps1
```diff
@@ -0,0 +1,44 @@
...
+
+    begin {
+        function GetExchangeWebSiteFromCn {
+            param (
+                [string]$Site
+            )
+
+            if ($null -ne $Site) {
+                $index = $Site.IndexOf("(") + 1
+                if ($index -ne -1) {
```
dpaulson45, line 21: This will never be -1, because the line above is adding 1. Needs to change to be 0 instead.

dpaulson45 (changes requested):

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
lusassl-msft: Copy/Past error. Good catch! Fixed.

Shared/ActiveDirectoryFunctions/Get-ExchangeProtocolContainer.ps1
lusassl-msft: Fixed

Shared/ActiveDirectoryFunctions/Get-ExchangeWebSitesFromAd.ps1
lusassl-msft: Indeed 👍 . Fixed

dpaulson45 (approved):

PR #1313, opened 2022-11-02
SetupAssist address update to latest version of Exchange

dpaulson45:
**Issue:**
Customer was trying to install Exchange 2019 in an Exchange 2013 environment. SetupAssist would determine that we are on Exchange 2013 schema and determine that everything was 'good' due to the fact that Exchange 2013 had a SU that caused the AD environment to be a little off.

However, this then caused the script to not attempt to call the code for `TestPrepareAD` to determine if we can install Exchange from this server due to domain and site in comparison to the schema master. Then we also would need to know about if the user running the script is in the correct groups to be able to run `Setup.exe /PrepareAD` if required.

While addressing this issue, ended up finding a different bug within the `Test-UserGroupMemberOf` that prevented a clear output to know what group we weren't in. 

**Reason:**
Script became unhelpful to track down the issue if the user isn't part of the correct group to be able to install Exchange with `Setup.exe /PrepareAD`.

**Fix:**
Within `Test-ExchangeADSetupLevel` ended up calling `Get-SetupLogReviewer` if the Exchange Setup Log was present on the computer. If it was, then proceed to detect the `SetupBuildNumber` from the returned results to determine which Exchange version we are trying to install. If the schema version in AD doesn't match what we are trying to install, it would just run the `TestMismatchLevel` code which calls everything that is required for the end user to interrupt. 

Broke out the testing within `Test-UserGroupMemberOf` so each test result would show up correctly and be cleaner in the output and useful.

image

**Validation:**
Lab tested.

### DIFF
--- a/Setup/SetupAssist/Checks/Domain/Test-ExchangeADSetupLevel.ps1
+++ b/Setup/SetupAssist/Checks/Domain/Test-ExchangeADSetupLevel.ps1
@@ -3,6 +3,7 @@
 
 . $PSScriptRoot\..\New-TestResult.ps1
 . $PSScriptRoot\..\UserContext\Test-UserGroupMemberOf.ps1
+. $PSScriptRoot\..\..\..\Shared\SetupLogReviewerFunctions.ps1
 function Test-ExchangeADSetupLevel {
 
     # Extract for Pester Testing - Start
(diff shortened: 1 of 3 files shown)


Setup/SetupAssist/Checks/Domain/Test-ExchangeADSetupLevel.ps1
```diff
@@ -157,11 +160,61 @@ function Test-ExchangeADSetupLevel {
 
     $adLevel = GetExchangeADSetupLevel
     $testName = "Exchange AD Latest Level"
+    $setupLog = "C:\ExchangeSetupLogs\ExchangeSetup.log"
```
lusassl-msft, line 163: How to we determine the setup log path? If it's the system drive, it would be better to use `$env:SystemDrive` instead of the hard-coded `C:\` drive.

dpaulson45: Yeah, probably should use that `$env:SystemDrive`.

bill-long (approved):

PR #1328, opened 2022-11-08
Build script rework

### LINKED ISSUE #1330: [Work Item] Build Number Logic Failure
**Issue**
Currently the `build.ps1` script is generating the build version based off of the most recent commit of the files within the script that is assembled together. While in short this has worked great, however, when PRs have been waiting around for a while with an old commit and other PRs get push through head with a newer commit a conflict can occur. This conflict prevents the latest changes to be recognized as a possible increase to the build version of the script causing the auto update feature to break.

**Describe the work**
Within `build.ps1` generate an artifact that will have a calculated last commit time from all the scripts that are being published and the files that are all within them and their commit time. Have a script that only runs on a PR against main that determines if we have increased the calculated build version of the script within this PR or not. If a build number doesn't increase due to this PR, then fail the build pipeline. 

**Additional Context**
The goal of this work item is to prevent commit to come into the project, then a release occurs that doesn't increase the build number of the script.

Release [v22.11.04.1613](https://github.com/microsoft/CSS-Exchange/releases/tag/v22.11.04.1613) had this problem. Because release [v22.11.02.2001](https://github.com/microsoft/CSS-Exchange/releases/tag/v22.11.02.2001) contained a new commit than when the PR #1312 finally was merged into main and then released.

bill-long:
**Issue 1:**
It's easy to accidentally release a new script. This can happen if a script that was previously included in another script is no longer included, or if a new script is added for testing or some other reason.

**Fix 1:**
This new build logic only publishes a script if one of the following is true:

* \docs contains a *.md file matching the script name.
* \docs contains a folder with an index.md file in it, where the folder name matches the script name.

**Issue 2:**
Build.ps1 performs repetitive disk-intensive work, which takes longer than necessary.

**Fix 2:**
The new logic scans the repo, building a hashtable of dependencies and commit times. These hashtables are then used within Build.ps1 itself, so that we don't have to go back to disk for the same information repeatedly.

**Issue 3:**
Build.ps1 is very quiet and doesn't give much information about what it's up to.

**Fix 3:**
The new logic produces a ton of output, including full dependency trees for each script, and the commit time on each of those files. It also calls out unreferenced scripts in the global Shared folder as well as scripts that are both undocumented and unreferenced in other folders.

**Issue 4:**
It is possible to merge a PR to main which has older timestamps than what is already in main. This causes a new script to release without a version number change, even though the script changed.

**Fix 4:**
Switch from author time to commit time for version numbers. Then, use the new ValidateMerge.ps1 in the pipeline for PRs to main. If it discovers commit times in the PR that are older than what is already in main, it will instruct the user to rebase, which will update the commit times in the PR. Fixes #1330.


bill-long:
Still troubleshooting a build issue. Do not approve.

dpaulson45 (approved):

PR #1333, opened 2022-11-08
Added parameters for limit collection

iserrano76:
Added some parameter to limit the collection of files to a range in time.
Now you can collect files from specific days in the past and not only a number of days in the past.

**Issue:**
With the previous parameter you can collect the logs from a number of days (or hours) in the past.
With the new parameters you can collect from a date to another one.

**Reason:**
If you know that the issue happened 5 days ago you don't need to collect all the files from today to 5 days back.
You can collect only the range of time that you need and avoid collecting more information that you need.

**Fix:**
Adding parameter to set the end date and you can use a starting date and an ending date.
You can use parameter with Worth parameter set or just a starting date and an ending date.

**Validation:**
Test it in lab.


dpaulson45:
I will have to review this likely once I am back from vacation.

bill-long:
It looks like the goal of this work is to not collect the latest logs. For example, start from three days ago, but only collect two days of logs, skipping the most recent day. You can already do that with:

-LogDate (Get-Date).AddDays(-3) -EndLogDate (Get-Date).AddDays(-1)

Is the scenario where we _don't_ want the latest logs such a common scenario that we need more parameters for this?

iserrano76:
If I run the script by myself, could be  enough but I think the option with the dates is a good option to have.
For example, when I send an action plan to a customer by email and I want to collect from 11/8 to 11/10.
if I send the option with (Get-Date).AddDays(-1), but they read my email 2 or 3 day later, it will fail.
 If they collect the logs with explicit dates, it will work always and easier to understand.

bill-long:
@iserrano76 Oh I see, I actually confused your changes with the old code. I like the -LogDate (should probably be LogStartDate) and -LogEndDate parameters. Since they take a date you could do the Get-Date with an adjustment or pass an explicit date, so I think that's fine.

However, I think we should restrict the date range feature to those two parameters. -EndDaysWorth and -EndLogAge are just seem awkward. Whereas a StartDate and EndDate is a common usage pattern, for example with message tracking, audit logs, etc.

iserrano76:
I added the option on Worth and LogAge just to add the same feature in all options but I thought that is not easy to understand.
I agree with you and only allow a range when you use date is better.
I will do those changes.

iserrano76:
@bill-long, I update the files to have only data range for LogStartDate  and LogEndDate  and update it the with the last changes without conflicts.
Let me know if it is fine.

dpaulson45 (approved):

PR #1345, opened 2022-11-10
Always use computer domain to bind to RootDSE

bill-long:
**Issue:**
If a user is logged in to a machine in the Exchange forest using a user account from a trusted forest, binding to LDAP://RootDSE gives us the user forest.

**Fix:**
Get the computer's domain and explicitly bind to that RootDSE.

### DIFF
--- a/Setup/SetupAssist/Checks/Domain/Test-ExchangeADSetupLevel.ps1
+++ b/Setup/SetupAssist/Checks/Domain/Test-ExchangeADSetupLevel.ps1
@@ -19,7 +19,7 @@ function Test-ExchangeADSetupLevel {
         $localDomainPrep = $null -ne $ADSetupLevel -and $ADSetupLevel.MESO.DN -eq "Unknown"
         Test-UserGroupMemberOf -PrepareAdRequired $true -PrepareSchemaRequired ($latestExchangeVersion.$ExchangeVersion.UpperRange -ne $currentSchemaValue) # -PrepareDomainOnly $localDomainPrep
 
-        $forest = [System.DirectoryServices.ActiveDirectory.Forest]::GetCurrentForest()
+        $forest = [System.DirectoryServices.ActiveDirectory.Domain]::GetComputerDomain().Forest
         $params = @{
             TestName = "Prepare AD Requirements"
             Result   = "Failed"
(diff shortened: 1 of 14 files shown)


Setup/SetupAssist/Checks/Domain/Test-ExchangeADSetupLevel.ps1
```diff
@@ -134,7 +134,7 @@ function Test-ExchangeADSetupLevel {
     }
 
     function GetExchangeADSetupLevel {
-        $rootDSE = [ADSI]("LDAP://RootDSE")
+        $rootDSE = [ADSI]("LDAP://$([System.DirectoryServices.ActiveDirectory.Domain]::GetComputerDomain().Name)/RootDSE")
```
dpaulson45, line 137: This is the same on line 148, could just set `$rootDSE` once then in this function.

dpaulson45:
Should probably do the same thing for `$rootDSE = [ADSI]("GC://RootDSE")` as well.

Setup/SetupAssist/Checks/Domain/Test-ExchangeADSetupLevel.ps1
bill-long: Good point. Not sure why that's in there twice.

lusassl-msft:
I'm wondering if we need to adjust the `Test-CVE-2021-34470.ps1` function as well. 

```
$schemaMaster = [System.DirectoryServices.ActiveDirectory.Forest]::GetCurrentForest().SchemaRoleOwner
```

GetCurrentForest() --> Gets a Forest object for the current user context.
https://learn.microsoft.com/en-us/dotnet/api/system.directoryservices.activedirectory.forest.getcurrentforest?view=netframework-4.7.2#system-directoryservices-activedirectory-forest-getcurrentforest

bill-long:
Good point. In fact, all calls to GetCurrentForest() need to be changed to GetComputerDomain().Forest.

lusassl-msft (approved):
Found no other functions that use the problematic code. Tested HealthChecker and it seems to work fine with the updated calls (as the return object should be the same). Approved and ready to merge from my point of view.

PR #1350, opened 2022-11-14
CVE-2022-21978 check doesn't work well when running AD split permissions

lusassl-msft:
**Issue:**
Multiple customers reported that the `CVE-2022-21978` isn't working and reported that the environment is vulnerable to this CVE. 

**Reason:**
All of these customers have AD split permissions in use. The following ACE check was marked as failed for each of those customers:

```
Trying to find the entry GUID: 0296c120-40da-11d1-a9c0-0000f80367c1
ObjectDN: DC=ad,DC=domain,DC=dom
Ace Result Check Passed: False
```

I assume it's expected to not have this ACE in place in AD split permissions scenarios:
_All ACEs assigned to the Exchange Windows Permissions security group are removed from the domain object._

https://learn.microsoft.com/en-us/exchange/permissions/split-permissions/split-permissions?view=exchserver-2019#active-directory-split-permissions

**Fix:**
Added a new function (`Get-ExchangeADSplitPermissionsEnabled`) to detect if customers are running AD split permissions mode. If they do so, initiate a fallback to check for the 'objectVersion (Default)' instead of the ACE.

**Validation:**
Lab & affected customer

### DIFF
--- a/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAdPermissions.ps1
+++ b/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAdPermissions.ps1
@@ -4,6 +4,7 @@
 . $PSScriptRoot\Get-ExchangeDomainConfigVersion.ps1
 . $PSScriptRoot\..\..\..\..\Shared\ErrorMonitorFunctions.ps1
 . $PSScriptRoot\..\..\..\..\Shared\ActiveDirectoryFunctions\Get-ActiveDirectoryAcl.ps1
+. $PSScriptRoot\..\..\..\..\Shared\ActiveDirectoryFunctions\Get-ExchangeADSplitPermissionsEnabled.ps1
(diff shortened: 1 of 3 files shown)


Diagnostics/HealthChecker/DataCollection/ExchangeInformation/Get-ExchangeAdPermissions.ps1
```diff
@@ -112,10 +113,15 @@ function Get-ExchangeAdPermissions {
 
             try {
                 try {
+                    # Check if AD split permissions is enabled and if so, throw to check for objectVersion instead ACE
+                    if (Get-ExchangeADSplitPermissionsEnabled) {
+                        throw "Active Directory split permissions enabled. Fallback to 'objectVersion (Default)' validation initiated."
```
bill-long, line 118: This approach to flow control should be reconsidered in the future, but I'm approving this for now.

bill-long (approved):

PR #1376, opened 2022-12-15
Added ShowMailboxInformation parameter

iserrano76:
**Issue:**
I added the parameter ShowMailboxInformation in the LoadBalancingReport to include an additional report from the backend side.

**Reason:**
I have been working in different cases were the customers had a server with a higher load, more users active in a server, or a few users with a high activity.
The FE load looks correct but the BE activity is not correctly defined. With that option we can check quickly the activity on Backend as we had before with LoadBalancingReport on FE

**Fix:**
collect perf counters on W3SVC_2 and analyze the information in the same way that with W3SVC_1

**Validation:**
Test it in Lab


dpaulson45:
Is there a reason to not include the mailbox connections? I am wondering if we should just default it to include both all the time.

iserrano76:
I did it in that way to keep the same behavior that was before and had a similar function that we have in all Get-*VirtualDirectory cmdlets, but maybe just leave it in all cases is simpler.
As you prefer, change it is easy.

dpaulson45:
> I did it in that way to keep the same behavior that was before and had a similar function that we have in all Get-*VirtualDirectory cmdlets, but maybe just leave it in all cases is simpler. As you prefer, change is easy.

I think that might be better vs adding even more switches to that feature. Let's simply it. 

Also, let's make it so we don't lose the commit history by doing a remove and add. Let's make sure it is a rename instead of the file. 

image

This would likely require you to save out the work you did to something like a notepad, do a rebase and drop the commit and you might need to start over. In order to get a rename the easiest is to do it in cmd prompt vs GitHub Desktop as it defaults to do a delete and create file. If you have questions on this, please feel free to reach out.

Diagnostics/HealthChecker/Features/Get-LoadBalancingReport.ps1
```diff
@@ -0,0 +1,324 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+. $PSScriptRoot\..\Helpers\PerformanceCountersFunctions.ps1
+. $PSScriptRoot\..\..\..\Shared\Invoke-CatchActionErrorLoop.ps1
+function Get-LoadBalancingReport {
+    Write-Verbose "Calling: $($MyInvocation.MyCommand)"
+    $CASServers = @()
+    $MBXServers = @()
+
```
dpaulson45, line 10: Right here let's do `$getExchangeServer = Get-ExchangeServer`

This way we aren't doing multiple `Get-ExchangeServer`.

Or you do a single `Get-ExchangeServer` and do a foreach loop and add the server to a list if everything is true. 

For large environments, we don't want to run `Get-ExchangeServer` twice if we can avoid it.

iserrano76: Fixed

dpaulson45 (approved):

PR #1391, opened 2022-12-21
Cherry Pick from PR to Release instead of Main

dpaulson45:
**Issue:**
Issue was missed where the PR from external was directly into release and not main. 

**Fix:**
Cherry pick the commits from release so they can go into main. 

Not sure if we also need the merge commit as well. 

**Validation:**
NA


dpaulson45:
The other option is to do a PR from Release into Main to fix this.

bill-long (approved):

PR #1400, opened 2022-12-27
Add ReplayQueueDatabases script for replaying many transport queues

bill-long:
When Transport crashes, in some cases it will move the current queue database and logs to a Messaging.old-\<date\> nested within the queue database path. Then it creates a new, empty queue database. If there is another crash, a new Messaging.old-<date> is created with the previous Messaging.old-\<date\> nested inside it.

After several crashes, this creates several challenges.

* If shadow redundancy was failing, we need to replay these databases. This involves tedious manual work of either pointing EdgeTransport.exe.config QueueDatabasePath and QueueLoggingPath to different folders, or moving the files/folders themselves around.
* When there have been over a dozen crashes, the path becomes so deep that both Explorer and PowerShell start to show errors when moving files, depending on the state of LongPathsEnabled.
* Even with LongPathsEnabled, EdgeTransport.exe.config cannot handle these deep paths, and Transport will repeatedly crash when pointed to these paths.

This script solves this problem by automating all of this. It uses long path syntax in PowerShell to find these folders, moves them into a short path that can be used in EdgeTransport.exe.config, and replays them one at a time, taking out all the manual work.


dpaulson45:
Would recommend requiring that the script checks for running as Administrator.

Could also see some issues if someone doesn't run the script in EMS.

Shared/Confirm-ExchangeManagementShell.ps1
```diff
@@ -0,0 +1,20 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+<#
+.SYNOPSIS
+    Returns True if we are running inside Exchange Management Shell, and False otherwise.
```
dpaulson45, line 6: If we aren't going to have Error Monitoring handling logic added to this, then we should call that out in the comments as well.

Other Shared scripts that handle an error should be passing in the ability to do a custom `$CatchActionFunction`.

bill-long: I'm not sure it make sense to add a comment calling out that a parameter is not present. I mean, look at the parameters. There is no -CatchActionFunction.

dpaulson45: Fair enough

dpaulson45 (approved):

PR #1401, opened 2022-12-27
Vulnerability Report Feature

dpaulson45:
**Reason:**
It was requested to create a feature that is able to generate a report for Vulnerabilities against the entire environment. This new feature in Health Checker now generates a report and saves it out to json for the list of the servers in the environment and their vulnerabilities. 

**Validation:**
Lab tested.


Diagnostics/HealthChecker/Features/Invoke-VulnerabilityReport.ps1
```diff
@@ -0,0 +1,116 @@
...
+    $serverNames = New-Object System.Collections.Generic.List[string]
+    $exchangeServers | ForEach-Object { $serverNames.Add($_.Name.ToLower()) }
+
+    # By default, always try to import the local data and see if we can use it.
+    # This is determined by the files being the same version as the current version of the script.
+    $importData = Get-ExportedHealthCheckerFiles -Directory $XMLDirectoryPath
+
+    if ($null -ne $importData) {
+        $importData |
+            Where-Object { $_.HealthCheckerExchangeServer.HealthCheckerVersion -eq $BuildVersion } |
```
lusassl-msft, line 27: Can we extract the `DateTime` string from the xml file (`HealthChecker-e2k16-2-20221228152212.xml`) and compare it against the current DateTime and decide then whether we use the existing xml files or not (e.g., if they are > 1 week old, re-collect the data to avoid using outdated files)?

lusassl-msft: Or are we good by using the HC version as an indicator?

lusassl-msft (approved):
Comment is optional, the rest of the code looks good

Diagnostics/HealthChecker/Features/Invoke-VulnerabilityReport.ps1
dpaulson45: HC version is likely the best indicator at this time.

PR #1417, opened 2023-01-06
Only check code format on changed files

dpaulson45:
**Reason:**
Improve the performance of the CodeFormatter script in the build pipeline.

**Fix:**
Only need to check all files if something changes with the CodeFormatter files. Otherwise, only check the files that were changed within this PR.

**Validation:**
Multiple scenarios checked

### DIFF
--- a/.build/CodeFormatter.ps1
+++ b/.build/CodeFormatter.ps1
@@ -3,7 +3,10 @@
 [CmdletBinding()]
 param(
     [Switch]
-    $Save
+    $Save,
+
+    [Switch]
+    $AllFiles
 )
 
 #Requires -Version 7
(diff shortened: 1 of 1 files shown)


.build/CodeFormatter.ps1
```diff
@@ -28,9 +31,63 @@ if (-not (Load-Module -Name EncodingAnalyzer)) {
...
+    }
+
+    # Also include modified files, but not committed yet for local work.
+    $gitStatus = git status --short
+    $m = $gitStatus | Select-String "M (.*)"
+    foreach ($match in $m) {
+        $file = $match.Matches.Groups[1].Value.Trim()
+        $fullPath = Join-Path $PSScriptRoot $file
+
+        if ($filesFullPath.Contains($fullPath)) {
```
bill-long, line 67: Calling Contains on a List[string] is bugging me. Let's make this a HashSet[string]. At that point we don't even have to check it - we can blindly add it and it will still only be in there once. If Add() returns $false, you know it was already in there and can report it if you want.

bill-long (changes requested):

.build/CodeFormatter.ps1
```diff
@@ -28,9 +31,63 @@ if (-not (Load-Module -Name EncodingAnalyzer)) {
...
+    $m = $gitlog | Select-String "^(\S+) (.*)$"
+
+    foreach ($commitMatch in $m) {
+        $commitHash = $commitMatch.Matches.Groups[1].Value
+        $filesChangedInCommit = git diff-tree --no-commit-id --name-only -r $commitHash
+
+        foreach ($fileChanged in $filesChangedInCommit) {
+            $fullPath = Join-Path $repoRoot $fileChanged
+
+            if ($filesFullPath.Contains($fullPath)) {
```
bill-long, line 51: Same here. HashSet makes this check faster in the case of a large number of changed files. That's why ValidateMerge uses a HashSet here.

bill-long (changes requested):

bill-long (approved):

PR #1435, opened 2023-01-11
SetupAssist to use Get-ExchangeBuildVersionInformation

### LINKED ISSUE #1314: [Work Item] Consistent Exchange CU and AD/Schema Levels Logic
**Describe the work**
Right now, in `SetupAssist`, the `Exchange AD Latest Level` test is only looking at the latest version of Exchange for the schema version that you are on. With the work that is done in PR #1313, it is now using the current setup log to determine the version of Exchange that it is trying to install and if that doesn't match the current schema version it will follow a different code path.

For the time being this will work. However, it would be better if we can create a common code usage to do a `Get-ExchangeVersionWithAdInformation` that would provide the following:

```
[string]MajorVersion: 2019
[string]FriendlyName: "Exchange 2019"
[system.version]BuildVersion: 15.2.1118.7
[string]CU: "CU12"
[string]ReleaseDate: "04/20/2022"
[string]ExtendedSupportDate: "10/14/2025"
[bool]Supported: $true
[object]ADLevel: 
     [int]SchemaValue: 17003
     [int]MESOValue: 13243
     [int]OrgValue: 16760
```

This way the code in `HealthChecker` for `Get-ExchangeInformation` and in `SetupAssist` in `Test-ExchangeADSetupLevel` can simply and when an update does occur, we would be changing the common code and not forget to update one of the scripts.

### LINKED ISSUE #1355: [Work Item] Additional Proactive Diagnostic check for SetupAssist.
**Describe the work**
Add a diagnostic for PrepareAD and PrepareSchema to SetupAssist as a pre-req check. I've had a number of cases where SetupAssist or SetupLogReviewer doesn't catch nuanced errors/warnings so perhaps if we dump the Versions for Schema or additional checks for PrepareAD it might save us time?

**Additional Context**
Perhaps a Read-Host for Exchange Version and CU/SU or just asking for the build while linking here - https://learn.microsoft.com/en-us/exchange/new-features/build-numbers-and-release-dates?view=exchserver-2019

dpaulson45:
**Reason:**
This allows for the shared code `Get-ExchangeBuildVersionInformation` that now contains the ORG, MESO, Schema version values in it to be used so we can use the last CU install attempt build number to determine what level of AD we should be at and if we need to run `/PrepareAD`, `/PrepareDomain`, or `/PrepareSchema`

Also included test for install watermark on the server. 

**Validation:**
Lab tested.
Resolved #1314 
Resolved #1355

### DIFF
--- a/Setup/SetupAssist/Checks/UserContext/Test-UserGroupMemberOf.ps1
+++ b/Setup/SetupAssist/Checks/UserContext/Test-UserGroupMemberOf.ps1
@@ -13,12 +13,8 @@ function Test-UserGroupMemberOf {
         [bool]$PrepareDomainOnly
     )
 
-    if ($null -eq $Script:UserSid) {
-        # Call the test manually
-        Test-UserIsAdministrator
-    }
-
-    $tokenGroups = Get-TokenGroupsGlobalAndUniversal -UserSid $Script:UserSid
+    $windowsIdentity = [System.Security.Principal.WindowsIdentity]::GetCurrent()
+    $tokenGroups = Get-TokenGroupsGlobalAndUniversal -UserSid $windowsIdentity.User.Value
 
     $groupRequirements = @(
         @{
@@ -53,7 +49,7 @@ function Test-UserGroupMemberOf {
         }
     }
 
-    $principal = (New-Object System.Security.Principal.WindowsPrincipal([System.Security.Principal.WindowsIdentity]::GetCurrent()))
+    $principal = (New-Object System.Security.Principal.WindowsPrincipal($windowsIdentity))
     $params = @{
         TestName = [string]::Empty
         Details  = [string]::Empty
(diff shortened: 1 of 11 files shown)


Setup/SetupAssist/Checks/UserContext/Test-UserGroupMemberOf.ps1
```diff
@@ -13,22 +13,12 @@ function Test-UserGroupMemberOf {
...
     )
 
-    $whoami = whoami
-    $whoamiAllOutput = whoami /all
-    $userSid = ($whoamiAllOutput | Select-String $whoami.Replace("\", "\\")).Line.Replace($whoami, "").Trim()
-
-    $params = @{
-        TestName = "User Administrator"
-        Details  = "$whoami $userSid"
+    if ($null -eq $Script:UserSid) {
```
bill-long, line 16: I'd like to avoid the use of a script-scoped variable.

Also, I'm not clear why we're calling whoami /all to get the SID when `[Security.Principal.WindowsIdentity]::GetCurrent().User.Value` already has it. That's what Confirm-Administrator is using. Do we need to use whoami at all? Is it just for output?

bill-long (changes requested):

Setup/SetupAssist/Checks/UserContext/Test-UserGroupMemberOf.ps1
dpaulson45: @bill-long check out latest changes

bill-long (approved):

bill-long (approved):

PR #1444, opened 2023-01-12
Create May 22 SU doc

### LINKED ISSUE #1440: [Issue] - Health Checker should say "Install X SU or newer" when making recommendations for addressing CVEs
**Provide Version Number**
Current version of HC.

**Describe the issue**
Right now, the way that HC recommends fixes for various CVEs might imply that specific SUs need to be installed, even though a newer SU might be available. Seeing that SUs are cumulative, we should add wording that "or newer SU" can also be installed.

**Script Output**
Right now, a sample output:

Security VulnerabilityCVE-2022-21978 
Install the May 2022 SU and run /PrepareDomain or /PrepareAllDomains
See: https://aka.ms/HC-May22SU

My recommendation is to in the second line always say "Install the May 2022 SU _or newer_..."

**Additional context**
Right now HC recommendation might give people an impression that SUs need to be installed sequentially, when SUs are cumulative.

dpaulson45:
**Reason:**
Create a doc to have the aka.ms link to point to stating the action that you need to take. 

**Validation:**
Resolved #1440

### DIFF
--- a/docs/Diagnostics/HealthChecker/May22SU.md
+++ b/docs/Diagnostics/HealthChecker/May22SU.md
@@ -1,0 +1,15 @@
+# May 2022 Security Update
+
+In order to protect against CVE-2022-21978 within your environment `/PrepareDomain` must be run against each domain that contains the MESO container within it.
+
+Health Checker will query all the domains in the environment to see if it has a MESO container. If it does, it checks for a particular ACE/or version number of the MESO container to see if we are secure. If we don't pass this check, it will provide what domains you need to run `/PrepareDomain` against.
+
+In order to protect your environment from CVE-2022-21978, you must install the May 2022 SU or a newer SU/CU that contains this security fix. All SUs and CUs after May 2022 contain this fix. After you have installed this security fix, you must run `/PrepareDomain` or `/PrepareAllDomains` from the Exchange bin directory.
+
+**Included in HTML Report?**
+
+Yes
+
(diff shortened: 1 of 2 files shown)


docs/Diagnostics/HealthChecker/May22SU.md
```diff
@@ -0,0 +1,15 @@
+# May 2022 Security Update
+
+In order to protect against CVE-2022-21978 within your environment `/PrepareDomain` must be run against each domain that contains the MESO container within it.
+
+Health Checker will query all the domains in the environment to see if it has a MESO container. If it does, it checks for a particular ACE/or version number of the MESO container to see if we are secure. If we don't pass this check, it will provide what domains you need to run `/PrepareDomain` against.
```
lusassl-msft, line 5:
```suggestion
Health Checker will query all the domains in the environment to see if it has a MESO container. If it does, it checks for a particular ACE or version number of the MESO container to see if we are secure. If we don't pass this check, it will provide what domains you need to run `/PrepareDomain` against.
```

lusassl-msft (approved):

docs/Diagnostics/HealthChecker/May22SU.md
lusassl-msft: Not sure if this is a typo or expected therefore approved.

lusassl-msft (approved):

PR #1462, opened 2023-01-23
Get info from Site Id in the loadbalancingReport

iserrano76:
**Issue:**
Some customer has additional site on the Exchange servers and the script does not work correctly if the Backend for example is not in the site 2.

**Reason:**
Fix the script for system without the default websites

**Fix:**
Verify the site ID of the Default Web Site and Exchange Back End and remove dependencies from site 1 and 2

**Validation:**
Test in lab

### DIFF
--- a/Diagnostics/HealthChecker/Features/Get-LoadBalancingReport.ps1
+++ b/Diagnostics/HealthChecker/Features/Get-LoadBalancingReport.ps1
@@ -98,7 +98,7 @@ function Get-LoadBalancingReport {
 
     $currentErrors = $Error.Count
     foreach ( $CASServer in $CASServers) {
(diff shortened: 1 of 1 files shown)


dpaulson45:
@iserrano76 use correct casing for PowerShell cmdlets. 

image


It should be `Get-IISSite` Which should pass our spell checker.

iserrano76:
Hi David, 
I miss that one. it is fixed now. 
I run .\.build\CodeFormatter.ps1 script without issues before, should not script warn?

dpaulson45:
> Hi David, I miss that one. it is fixed now. I run ..build\CodeFormatter.ps1 script without issues before, should not script warn?

It is a different script called `.build\spellcheck.ps1`

dpaulson45 (approved):

PR #1476, opened 2023-02-02
Add feature to send out an email when an Auth Certificate renewal action was performed

lusassl-msft:
**Description:**
This feature was requested by many customers. The script can now send out an notification via email if an Auth Certificate renewal action was performed. The script makes use of the EWS API. It only attempts to send once per run and only if a renewal action was performed or in case that the `-TestEmailNotification` parameter was used. The text of the notification is build dynamically based on the detected environment (multiple AD sites, Hybrid...). 

Examples:
image

image

**Validation:**
Lab


Admin/MonitorExchangeAuthCertificate/MonitorExchangeAuthCertificate.ps1
```diff
@@ -434,6 +553,23 @@ function Main {
             Write-Host ""
             Write-Host ($multipleExchangeADSitesWording) -ForegroundColor Yellow
         }
+
+        if ((-not($WhatIfPreference)) -and
```
bill-long, line 557: We should use ShouldProcess, not manually check $WhatIfPreference. See the examples here: https://learn.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-shouldprocess

bill-long (changes requested):

Admin/MonitorExchangeAuthCertificate/ConfigurationAction/Register-AuthCertificateRenewalTask.ps1
```diff
@@ -58,7 +60,17 @@ function Register-AuthCertificateRenewalTask {
         $newScheduledTaskParams = @{
             Execute          = "powershell.exe"
             WorkingDirectory = "$($WorkingDirectory)"
-            Argument         = "-NonInteractive -NoLogo -NoProfile -Command `".\$($ScriptName) -ValidateAndRenewAuthCertificate `$true -IgnoreUnreachableServers `$$($IgnoreOfflineServers) -IgnoreHybridConfig `$$($IgnoreHybridConfig) -Confirm:`$false`""
+        }
+
+        $basicArgumentParameters = "-NonInteractive -NoLogo -NoProfile -Command `".\$($ScriptName) -ValidateAndRenewAuthCertificate `$true -IgnoreUnreachableServers `$$($IgnoreOfflineServers) -IgnoreHybridConfig `$$($IgnoreHybridConfig)"
+        if (-not([System.String]::IsNullOrEmpty($SendEmailNotificationTo))) {
```
bill-long, line 66: $SendEmailNotificationTo is [string[]] but this check pretends it's [string]. This will not work consistently. For example, if you provide an actual [string[]] where all the values are empty:

image

bill-long (changes requested):

Admin/MonitorExchangeAuthCertificate/ConfigurationAction/Register-AuthCertificateRenewalTask.ps1
lusassl-msft: Added `ValidatePattern` and switched the check to a `$null` check. Should be sufficient in this case.

Admin/MonitorExchangeAuthCertificate/MonitorExchangeAuthCertificate.ps1
lusassl-msft: `$WhatIfPreference` is checked at several points in the script. Will open a work item to track the work to handle this better as suggested.

lusassl-msft: https://github.com/microsoft/CSS-Exchange/issues/1487

bill-long (approved):

PR #1484, opened 2023-02-11
Add FindFrontEndActivity script

bill-long:


Diagnostics/FindFrontEndActivity.ps1
```diff
@@ -0,0 +1,198 @@
...
+            $results = New-Object System.Collections.ArrayList
+            foreach ($thisProtocol in $using:Protocol) {
+                $logPathForThisProtocol = Join-Path $httpProxyLogPath $thisProtocol
+                if (-not (Test-Path $logPathForThisProtocol)) {
+                    continue
+                }
+
+                $files = Get-ChildItem "$logPathForThisProtocol\*.log" | Where-Object {
+                    if ($null -ne $using:timeThreshold) {
+                        $_.LastWriteTime -gt $using:timeThreshold
```
dpaulson45, line 137: Not sure how common it is, but possible since this script doesn't require to be run on the server itself you can have different time zones from the execution computer and the Exchange Servers or possibly have different time zones within all the servers themselves. Best to use UTC.

Diagnostics/FindFrontEndActivity.ps1
```diff
@@ -0,0 +1,198 @@
...
+                    } else {
+                        $_.LastWriteTime -gt $using:StartTime -and $_.LastWriteTime -lt $using:EndTime
+                    }
+                } | Sort-Object LastWriteTime
+
+                $files = $files | Where-Object { Select-String $using:samAccountRegexForFileMatch $_.FullName -Quiet }
```
dpaulson45, line 143:
```suggestion
                $files = @($files | Where-Object { Select-String $using:samAccountRegexForFileMatch $_.FullName -Quiet })
```
To make sure that it is treated as an array

Diagnostics/FindFrontEndActivity.ps1
```diff
@@ -0,0 +1,198 @@
...
+.PARAMETER StartTime
+    Logs older than this time are not searched. This is a DateTime value, such as (Get-Date).AddDays(-1)
+    or "2023-02-11 08:00". Use this parameter to search old logs. Use -TimeSpan to search the
+    most recent logs.
+.PARAMETER EndTime
+    Logs newer than this time are not searched. This is a DateTime value, such as (Get-Date).AddDays(-1)
+    or "2023-02-11 09:00". Use this parameter to search old logs. Use -TimeSpan to search the
+    most recent logs.
+.LINK
+    https://microsoft.github.io/CSS-Exchange/Diagnostics/FindFrontEndActivity
```
dpaulson45, line 54: Should probably change to an aka.ms link

Diagnostics/FindFrontEndActivity.ps1
```diff
@@ -0,0 +1,198 @@
...
+    $StartTime,
+
+    [Parameter(ParameterSetName = 'Range', Mandatory = $true)]
+    [DateTime]
+    $EndTime
+)
+
+begin {
+    $serverNames = New-Object System.Collections.ArrayList
+    $samAccountRegexForFileMatch = [string]::Join('|', ($SamAccountName | ForEach-Object { "\\$_," }))
```
dpaulson45, line 108: Might want to verify that the passed value isn't something like this `contoso\user`

Diagnostics/FindFrontEndActivity.ps1
```diff
@@ -0,0 +1,198 @@
...
+}
+
+end {
+    $timeThreshold = $null
+    if ($PSCmdlet.ParameterSetName -eq 'Recent') {
+        $timeThreshold = (Get-Date).Add(-$TimeSpan)
+    }
+
+    $jobs = $serverNames | ForEach-Object {
+        Invoke-Command -AsJob -ComputerName $_ -ScriptBlock {
```
dpaulson45, line 125: This does require the script to be run as Administrator otherwise, you will get an error back.

image

dpaulson45 (changes requested):

Diagnostics/FindFrontEndActivity.ps1
bill-long: That shouldn't be necessary. DateTime contains an offset. I just tested this in my lab, and it pulls the correct file:

image

dpaulson45: Perfect, didn't know that.

Diagnostics/FindFrontEndActivity.ps1
```diff
@@ -0,0 +1,206 @@
...
+    [Parameter(ParameterSetName = 'Range', Mandatory = $true)]
+    [DateTime]
+    $EndTime
+)
+
+begin {
+    . $PSScriptRoot\..\Shared\Confirm-Administrator.ps1
+
+    if (-not (Confirm-Administrator)) {
+        Write-Host "This script must be run as an Administrator."
```
dpaulson45, line 111: should exit if not Admin

PR #1508, opened 2023-02-17
BinSearchFolder false-positive caused by PR 1495

### LINKED ISSUE #1506: [Issue] - Health Checker 23.02.16.1810 throwing new error about Bin Search folder
**Provide Version Number**
23.02.16.1810

**Describe the issue**
in the 23.02.16.1810 version i'm seeing a new error for "Bin Search Folder Not Found"

havent seen this issue before in previous runs and i'm now seeing it on all instances 

also didnt see it on feb 14 in wake of the installs of the feb updates

**Expected behavior**
unclear what the issue is as looking at the linked article the suggested values for the bin search path appear to match exactly

**Script Output**
"Bin Search Folder Not Found"

**Additional context**

lusassl-msft:
**Issue:**
`BinSearchFolder` is reported as missing. However, it's there and not missing so it's a false-positive.

**Reason:**
Due to the work to reduce the size of the exported file, we added the `Raw` parameter at some places. By using the `Raw` parameter, the contents of a file (web.config in this case) is returned as one string, instead of an array of strings. As a result, it broke our logic in `Invoke-AnalyzerFrequentConfigurationIssues` and caused the false-positive
PR #1495

**Fix:**
Changed the logic in `Invoke-AnalyzerFrequentConfigurationIssues` to handle this properly by converting the content to `xml` and searching for the `BinSearchFolders` key.

Resolved #1506

**Validation:**
Lab

### DIFF
--- a/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
+++ b/Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
@@ -25,9 +25,18 @@ function Get-IISWebSite {
         $configurationFilePath = (Get-WebConfigFile "IIS:\Sites\$($site.Name)").FullName
         $webConfigExists = Test-Path $configurationFilePath
         $webConfigContent = $null
+        $validWebConfig = $false
 
         if ($webConfigExists) {
             $webConfigContent = Get-Content $configurationFilePath -Raw
+
+            try {
+                [xml]$webConfigContent | Out-Null
+                $validWebConfig = $true
+            } catch {
+                # Inside of Invoke-Command, can't use Invoke-CatchActions
+                Write-Verbose "Failed to convert IIS web config '$configurationFilePath' to xml. Exception: $($_.Exception)"
+            }
         }
 
         $returnList.Add([PSCustomObject]@{
@@ -49,6 +58,7 @@ function Get-IISWebSite {
                     Location = $configurationFilePath
                     Content  = $webConfigContent
                     Exist    = $webConfigExists
+                    Valid    = $validWebConfig
                 }
             }
         )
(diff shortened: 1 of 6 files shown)


Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
```diff
@@ -25,9 +25,18 @@ function Get-IISWebSite {
...
 
         if ($webConfigExists) {
             $webConfigContent = Get-Content $configurationFilePath -Raw
+
+            try {
+                [xml]$webConfigContent | Out-Null
+                $validWebConfig = $true
+            } catch {
+                # Inside of Invoke-Command, can't use Invoke-CatchActions
+                Write-Verbose "Failed to convert IIS web config '$configurationFilePath' to xml. Exception: $($_.Exception)"
```
lusassl-msft, line 38:
```suggestion
                Write-Verbose "Failed to convert IIS web config '$($configurationFilePath)' to xml. Exception: $($_.Exception)"
```

Think we need to change this as suggested.

dpaulson45 (approved):

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
dpaulson45: `$configurationFilePath` is a string so you shouldn't need to do that. Only if it is an object where you are doing something like `$configurationFilePath.Attribute` then you would need to do that.

lusassl-msft: You're right:
image

PR #1525, opened 2023-02-28
False Warning with ServerList parameter

iserrano76:
**Issue:**
If you use the ServerList parameter the script shows that is not in the list in any case because it is not used the Name property.

**Reason:**
Remove the false warning

**Fix:**
Compare with the correct propert.

**Validation:**
Test in lab


dpaulson45:
The list variable appears to only be used for a list of server names. You should store it this way instead.

iserrano76:
> 

Just added changes to limit the information on the CASServers and  MBXServer variables.

dpaulson45 (approved):

PR #1539, opened 2023-03-02
Allow piping servers to ExchangeLogCollector

bill-long:
Allow easily running ExchangeLogCollector against many or all servers:

image


dpaulson45:
For some reason in my lab, this isn't working when doing the pipeline to determine the available free space. But if I pass my server manually, it works. 

image

bill-long:
Looks like that was because $Servers could be an array of anything, not just string. Piping it was producing an array of server objects, not names.

dpaulson45 (approved):

PR #1549, opened 2023-03-08
Remove Organization dependency when collecting server data

dpaulson45:
**Issue:**
Really shouldn't need to pass ORG information to server data collection to have it create a data object.

**Reason:**
This breaks down the ability to speed up the script and pointless to pass ORG information to the data collection of the server to 'analyze' the data to produce a different type of object. This should be handled in the analyzer section of the script. 

**Fix:**
Remove dependencies that required information from Organization and handle the data analysis in the analyzer code. 

**Validation:**
Lab tested/pester tested.


bill-long (dismissed):

Diagnostics/HealthChecker/Analyzer/Get-FilteredSettingOverrideInformation.ps1
```diff
@@ -0,0 +1,118 @@
...
+                                    ComponentName  = $entry.ComponentName
+                                    SectionName    = $entry.SectionName
+                                    ParameterName  = $FilterParameterName
+                                    ParameterValue = $value
+                                    Status         = $entry.Status
+                                    TrueStatus     = $status
+                                    FromAdSettings = $usedAdSettings
+                                })
+                        } elseif ($status -eq "Accepted") {
+                            Write-Verbose "Already have 1 Accepted value added to list no need to add another one. Skip adding $($entry.Name)"
```
lusassl-msft, line 102: This would mean that we filter out all duplicate overrides which enable a feature (e.g., SerializedDataSining) with the status `Accepted` and so, we don't show them in the final output. If that's what we wanted to do, it works as expected. Otherwise, we might need to adjust the logic to not exclude duplicate items.

image

image

lusassl-msft (approved):

PR #1596, opened 2023-03-17
Missing credentials for connect-AzureAD

Petri-X:
There was a parameter for credentials on the main .PS1, but that was not used with Connect-AzureAD

**Issue:**
Even when you are using -credential, that still prompting the credential box for user. And because of that this cannot be used as a job.

**Reason:**
Error correction

**Fix:**
Even when you are using -credential, that still prompting the credential box for user. And because of that this cannot be used as a job.

**Validation:**
n/a

### DIFF
--- a/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
+++ b/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
@@ -416,7 +416,7 @@ begin {
             if ($null -ne (Get-InstalledModule -Name "AzureAD" -ErrorAction SilentlyContinue)) {
                 Import-Module "AzureAD" -ErrorAction Stop
                 Write-Host "`nPrompting user for authentication, please minimize this window if you do not see an authorization prompt as it may be in the background"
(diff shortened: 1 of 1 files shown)


lusassl-msft:
@Petri-X the `credential` parameter is used to run the script against Exchange Server on-premises. For runs against Exchange Online, we create an app in Azure and use an app secret with a short lifetime of 7 days. 

It's required to log on with a user who has the permissions to create the app and generate the app secret. It's not expected that the user/password will be passed via the `-Credential` parameter and therefore we show the Azure logon dialogue.

Credential parameter description: `This optional parameter lets you pass administrator credential object while running on Exchange Server on-premises.`

Adding the `-credential` parameter to the `Connect-AzureAD` call would change the logic of the script.

Petri-X:
@lusassl-msft 
Just one question, the reason I want to use the -Credential on "connect-AzureAD" is the possibility to start running this scanning on the jobs. Yes, I know that running patches is possible, but that always requires extra PS session and signing. But when scanning hundreds thousands mailboxes, individual PS sessions are not good enough.

Are you aware if anybody has considered this from that point of view?

The credentials are of course coming from "get-credentials", but that might as bad?

PR #1598, opened 2023-03-17
Use deep traversal search folder to improve performance

bill-long:
The CVE-2023-23397.ps1 script can take a long time to run.

In the current version, we bind to each individual folder in the mailbox and perform a search in that folder. This means repeated calls against the same mailbox. The more folders, the more calls.

Also, this search is synchronous. The FindItems call will not return until the server has the results. So we spend time waiting for these calls, and potentially more time on folders with lots of items.

This PR uses a different approach, essentially making the search asynchronous.

* In the first pass, we create a single search folder in each mailbox. This search folder searches the whole mailbox for relevant items.
* In the second pass, we read the results from that folder.

This allows the server to do the work in the background. The script can just come by later and read the results.

Usage:

To take advantage of the new feature, run with the same syntax as usual, but add -UseSearchFolders. Example:

```powershell
Get-Mailbox -ResultSize Unlimited | .\CVE-2023-23397.ps1 -Environment Online -UseSearchFolders
```

This switch only applies to Audit mode. Cleanup mode has no syntax changes. To take maximum advantage of the search folders, it's best to leave them in place until cleanup is done, so you can repeatedly and quickly search for any new items. After cleanup is completed, the search folders can be removed with:

```powershell
Get-Mailbox -ResultSize Unlimited | .\CVE-2023-23997.ps1 -Environment Online -UseSearchFolders -SearchFolderCleanup
```

### DIFF
--- a/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
+++ b/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
@@ -250,8 +250,6 @@ begin {
             $Service = New-Object Microsoft.Exchange.WebServices.Data.ExchangeService([Microsoft.Exchange.WebServices.Data.ExchangeVersion]::Exchange2016)
         }
 
-        $Service.Timeout = 900000 # 15 minutes
-
         if ($Environment -eq "Onprem") {
             $Service.Credentials = New-Object Microsoft.Exchange.WebServices.Data.WebCredentials($credential.UserName, [Runtime.InteropServices.Marshal]::PtrToStringAuto([Runtime.InteropServices.Marshal]::SecureStringToBSTR($credential.Password)))
         } else {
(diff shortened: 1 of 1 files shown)


lusassl-msft:
Promote new search feature via PR #1611

Security/src/CVE-2023-23397/CVE-2023-23397.ps1
```diff
@@ -759,59 +830,124 @@ begin {
...
+
+                    if ($itemResults.Count -lt 1) {
+                        $mailboxProcessed += 1
+                        continue
+                    }
+
+                    $items = $ewsService.LoadPropertiesForItems($itemResults, $PropertySet)
+
                     foreach ($item in $items) {
                         if (-not $IdsProcessed.Contains($item.Item.Id)) {
```
varungupta-msft, line 952:
```suggestion
                        if (-not $IdsProcessed.Contains($item.Item.Id) -or $[System.String]::IsNullOrEmpty($item.Item.ExtendedProperties[0].Value)) {
```

This would fix the false positive (entries with no value in PidLidParameter column) that customers are getting in audit mode.

SharmaAkash1 (approved):

lusassl-msft (approved):

PR #1606, opened 2023-03-20
Fix ordered list

bill-long:
Indenting the content _should_ keep the content under step 2 without causing the list numbering to start over.


Nasir1786 (commented):
look Good

Nasir1786 (approved):

PR #1619, opened 2023-03-22
Remove false positive reported during Audit mode

varungupta-msft:
**Issue:**
When the script is run in audit mode a lot of false positives are reported in the csv file.

**Fix:**
Added an additional check to make sure that the vulnerable property is not null or empty before adding it to the csv


dpaulson45:
This makes the logic different between the old search method and the new search method, we shouldn't do this.

dpaulson45:
@ninobb please view as well. Do we want to all of a sudden change how we are auditing items now?

bill-long (changes requested):
If line 1040 changes, that _only_ changes the behavior when not using the -SearchFolders flag. So we would also need to change line 970, so you get the same results. Otherwise, using -SearchFolders you would still get blank values, but without -SearchFolders you would not.

However, we should also consider whether it makes sense to change the output. This might surprise some users who run the script and get very different results than with the previous version.

ninobb:
I feel this is a good change (if we have changed both types of searches). It lowers the "noise", if we are certain we are ignoring "empty" entries only. We should also do a documentation / FAQ change to the effect "The script reported messages with blank entries but it does not do that anymore?" or something similar, to address the fact that we have changed the behavior of this.

bill-long (approved):

bill-long:
Approved. We need to push a docs change prior to releasing this.

dpaulson45:
We should just do that in this PR right?

bill-long:
> We should just do that in this PR right?

Either way is fine with me.

bill-long (approved):

PR #1624, opened 2023-03-23
Add Certificate Based Authentication (CBA) support to CVE-2023-23397

lusassl-msft:
**Description:**
This PR will bring CBA support to the `CVE-2023-23397` script. It's possible to use a certificate to authenticate after the `CVE-2023-23397Application` was created. 

An administrator who has access to the private key of the certificate can acquire the required token to perform the `Audit` or `Cleanup` operations without the requirement of assigning any further permissions like `Global Administrator` or `Application Administrator`.

### DIFF
--- a/.build/cspell-words.txt
+++ b/.build/cspell-words.txt
@@ -58,6 +58,7 @@ hostnames
 HRESULT
 hsts
 httperr
+ietf
 imap
 Infoworker
 INSUFF
@@ -105,11 +106,12 @@ ntlm
 NTFS
 NUMA
 nupkg
+onmicrosoft
 onprem
 OutlookiOS
 Perflib
 perfmon
-PKCS
+Pkcs
 Prelicensing
 PROCMON
 QWORD
--- a/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
+++ b/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
@@ -43,6 +43,12 @@
     This optional parameter allows you to provide Azure environment
 .PARAMETER AzureApplicationName
     This optional parameter allows you to provide Azure application name
+.PARAMETER CertificateThumbprint
+    This optional parameter allows you to provide a certificate thumbprint to use the Azure application created by the script
+.PARAMETER AppId
+    This optional parameter allows you to provide the AppId of the Azure application created by the script
+.PARAMETER Organization
+    This optional parameter allows you to provide the organization name e.g., contoso.onmicrosoft.com
 .PARAMETER EWSOnlineURL
     This optional parameter allows you to provide EWS online url
 .PARAMETER EWSOnlineScope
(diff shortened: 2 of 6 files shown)


Security/src/CVE-2023-23397/CVE-2023-23397.ps1
```diff
@@ -628,7 +659,15 @@ begin {
 
         # if token is going to expire in next 5 min then refresh it
         if ($null -eq $script:tokenLastRefreshTime -or $script:tokenLastRefreshTime.AddMinutes(55) -lt (Get-Date)) {
```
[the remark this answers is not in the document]
lusassl-msft: Resolved via upcoming commit.

.build/cspell-words.txt
```diff
@@ -58,6 +58,7 @@ hostnames
 HRESULT
 hsts
 httperr
+ietf
```
dpaulson45, line 61: Can we remove this from the dictionary and do something like "ieTf" in the string and do a `.ToLower()` to avoid adding these odd words into the dictionary list.

Shared/AzureFunctions/New-JsonWebToken.ps1
```diff
@@ -0,0 +1,116 @@
...
+        [string]$SigningAlgorithm = "RS256"
+    )
+
+    <#
+        Shared function to create a signed Json Web Token (JWT) by using a certificate.
+        It is also possible to use a secret key to sign the token, but that is not supported in this function.
+        The function returns the token as a string if successful, otherwise it returns $null.
+        https://www.rfc-editor.org/rfc/rfc7519
+        https://learn.microsoft.com/azure/active-directory/develop/active-directory-certificate-credentials
+        https://learn.microsoft.com/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow
```
dpaulson45, line 38: This currently does NOT return a `$null` value the way you have it set up.

A return in `process` will execute the `end` block still. Then returning ".." base off this line:

```
    end {
        Write-Verbose "Returning Json Web Token"
        return ($headerBase64 + "." + $payloadBase64 + "." + $signature)
    }
```

dpaulson45 (changes requested):

Shared/AzureFunctions/New-JsonWebToken.ps1
```diff
@@ -0,0 +1,116 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+function New-JsonWebToken {
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseShouldProcessForStateChangingFunctions', '', Justification = 'Only non-destructive operations are performed in this function.')]
+    [CmdletBinding()]
```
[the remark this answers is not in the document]
lusassl-msft: Sure. We can rename it and `Get` is better. Will rename with the next commit.

.build/cspell-words.txt
lusassl-msft: Isn't that what the words list exists for? `ietf` (in lower case) is commonly used. Doing something like ieTf and then doing a `.ToLower()` to make spellcheck happy is not necessarily the most beautiful solution IMO.

dpaulson45: Okay, let's keep it and do another commit for Test-HMAEAS for line 259 and make that change there as well. Could either be this PR or a different PR.

lusassl-msft: Done

docs/Security/CVE-2023-23397/index.md
```diff
@@ -55,7 +55,11 @@ Set-Mailbox -Identity "<UserWhoRunsTheScript>" -ThrottlingPolicy "CVE-2023-23397
 
 ### Prerequisites to run the script for Exchange Online
 
-To run this script in an Exchange Online environment, you need to be a Global Administrator or an Application Administrator. The script will create an application with full access permission on all the mailboxes. To scan cloud mailboxes, you also need to add the Azure-AD module to PowerShell. See [Install AzureAD PowerShell for Graph](https://learn.microsoft.com/powershell/azure/active-directory/install-adv2?view=azureadps-2.0) for more information.
+To run this script in an Exchange Online environment, you need to be a `Global Administrator` or an `Application Administrator`. The script will create an application with full access permission on all the mailboxes.
```
SharmaAkash1, line 58: Should we add details on being able to run script as owner without needing Global or App admin permissions?

SharmaAkash1 (dismissed):

docs/Security/CVE-2023-23397/index.md
SharmaAkash1: Intention is for customers to be aware of that approach - in case they want to avoid the cert management caveats

PR #1635, opened 2023-03-28
Update wording of missing property

### LINKED ISSUE #1634: [Issue] CVE-2023-23397.ps1 reports error when property already cleaned
Script output shows:

```powershell
Failed to clear property for entry number: 1, Line number: 2
```

This just means the property isn't present. It was already cleaned up. We should fix this output to be clearer and to not indicate an error occurred.

dpaulson45:
**Issue:**
The current output is confusing. Customers appear to run the script twice trying to remove the item, but that isn't a valid way of verifying that the item has been removed. 

**Fix:**
Cleaned up the wording of the error. 

Resolved #1634 

**Validation:**
N/A


dpaulson45:
Closing because PR #1636 is going to address this.

PR #1640, opened 2023-03-29
Resolving Autodiscover redirect block issue

varungupta-msft:
**Issue:**
While fetching the EWS URL from Autodiscover redirect is blocked.

**Fix:**
Created a callback to allow redirect

### DIFF
--- a/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
+++ b/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
@@ -291,7 +291,12 @@ begin {
                 CheckOnpremCredentials -ewsService $Service
             } else {
                 try {
-                    $Service.AutodiscoverUrl($Credential.UserName)
+                    $redirectionCallback = {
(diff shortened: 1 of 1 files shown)


Security/src/CVE-2023-23397/CVE-2023-23397.ps1
```diff
@@ -291,7 +291,12 @@ begin {
                 CheckOnpremCredentials -ewsService $Service
             } else {
                 try {
-                    $Service.AutodiscoverUrl($Credential.UserName)
+                    $redirectionCallback = {
+                        param([string]$url)
+                        return $url.ToLower().StartsWith("https://autodiscover-s.outlook.com")
```
lusassl-msft, line 296: This would only work with WW cloud but not with GCC, GCC-High or China. PR #1638 comes with a change that makes the script work with these clouds. We should probably first validate and merge #1638 and then re-use the logic to determine the correct cloud endpoint for the redirectionCallback as well.

PR #1130, opened 2022-07-11
Matt avtestprocessupdate

Canthv0:
**Issue:**
Expanding script to also scan exchange processes for "unknown" modules being loaded.

**Reason:**
AV will load into processes and not just scan files.

**Fix:**
Added code to pull all exchange processes that are listed on the AV exclusions page.
Scans loaded modules looking for one that are not in a white list.

**Validation:**
Provide if applicable

PR #1641, opened 2023-03-30
Adding Process analysis

iserrano76:
**Issue:**
Included a Test for Exchange processes looking for 3rd party DLLs

**Reason:**
Include processes test in the script.
It is based in the #1130 PR but Including signature validation of the files.

**Fix:**
Check the files running are signed by a know issuer (Microsoft or VeriSign in case of FIPS files)
If the file is not signed, it will verify the company or the module name is well-known.

**Validation:**
Test it in lab.


Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -104,205 +200,434 @@ if (-not($exchangeShell.ShellLoaded)) {
...
+                    Write-SimpleLogFile -String ("############################################################") -name $LogFile
+                    Write-SimpleLogFile -String ("Working on Process $($Instance.Path)") -name $LogFile
+                    Write-SimpleLogFile -String ("Working on PID $($Instance.Id)") -name $LogFile
+                    #$CommandLine = (Get-WmiObject Win32_Process -Filter "ProcessId = '$($Instance.Id)'" | Select-Object CommandLine).CommandLine
+                    $CommandLine = (Get-CimInstance  -Query "Select * from Win32_Process where ProcessId = '$($Instance.Id)'" | Select-Object CommandLine).CommandLine
+
+                    Write-SimpleLogFile -String ("CommandLine: $CommandLine") -name $LogFile
+
+                    $k = 0
+                    foreach ($module in $Instance.Modules ) {
```
bill-long, line 422: Please give $k a meaningful name, like $modulesProcessed or something.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -104,205 +200,434 @@ if (-not($exchangeShell.ShellLoaded)) {
...
+        [array]$RunningProcess = Get-Process $ProcessName -ErrorAction SilentlyContinue
+
+        # Look at if we have found it
+        if ($null -eq $RunningProcess) {
+            Write-SimpleLogFile -string "Process $Process not found" -name $LogFile
         } else {
+            Write-SimpleLogFile -string "Found $Process" -name $LogFile
+
+            $j = 0
+            # Pull each instance of the process
```
bill-long, line 408: $j also needs a meaningful name. The name should describe what it's counting.

bill-long (changes requested):

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -62,6 +78,86 @@ param (
...
+
+if ( $IncludeW3wpProcesses -and ( -not $ProcessesAnalysis ) ) {
+    Write-Error "IncludeW3wpProcesses parameter works with Processes Analysis"
+    exit
+}
+
+if ( $Recurse -and ( -not $DirectoriesAnalysis ) ) {
+    Write-Error "Recurse parameter works with Directories Analysis"
+    exit
+}
```
dpaulson45, line 159: This should be handled with `ParameterSetNames`

iserrano76: Hi David,

I was thinking about that when I wrote the script, but I thought this solution is simpler.

If I choose the parameterSetNames I need to add many options and add complexity. 
That was what I tried:

    [Parameter(Mandatory = $true, ParameterSetName = 'DirectoryAnalysis')]
    [Parameter(Mandatory = $true, ParameterSetName = 'DirectoryProcessesAnalysis')]
    [Parameter(Mandatory = $true, ParameterSetName = 'DirectoryExtensionsAnalysis')]
    [Parameter(Mandatory = $true, ParameterSetName = 'DirectoryProcessesExtensionsAnalysis')]
    [switch]
    $DirectoriesAnalysis,

    [Parameter(Mandatory = $false, ParameterSetName = 'NoSelectedAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'DirectoryAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'DirectoryProcessesAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'DirectoryProcessesExtensionsAnalysis')]
    [switch]
    $Recurse,

    [Parameter(Mandatory = $true, ParameterSetName = 'DirectoryProcessesAnalysis')]
    [Parameter(Mandatory = $true, ParameterSetName = 'DirectoryProcessesExtensionsAnalysis')]
    [Parameter(Mandatory = $true, ParameterSetName = 'ProcessesAnalysis')]
    [Parameter(Mandatory = $true, ParameterSetName = 'ProcessesExtensionsAnalysis')]
    [switch]
    $ProcessesAnalysis,

    [Parameter(Mandatory = $false, ParameterSetName = 'NoSelectedAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'DirectoryProcessesExtensionsAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'ProcessesAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'ProcessesExtensionsAnalysis')]
    [switch]
    $IncludeW3wpProcesses,

    [Parameter(Mandatory = $true, ParameterSetName = 'DirectoryExtensionsAnalysis')]
    [Parameter(Mandatory = $true, ParameterSetName = 'DirectoryProcessesExtensionsAnalysis')]
    [Parameter(Mandatory = $true, ParameterSetName = 'ProcessesExtensionsAnalysis')]
    [Parameter(Mandatory = $true, ParameterSetName = 'ExtensionsAnalysis')]
    [switch]
    $ExtensionsAnalysis,

    [Parameter(Mandatory = $false, ParameterSetName = 'NoSelectedAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'DirectoryAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'DirectoryProcessesAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'DirectoryExtensionsAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'DirectoryProcessesExtensionsAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'ProcessesAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'ProcessesExtensionsAnalysis')]
    [Parameter(Mandatory = $false, ParameterSetName = 'ExtensionsAnalysis')]
    [switch]
    $OpenLog


And the message in some case of a wrong parameter is less clear, for example:
image

But with the current script:
image

That was the reason why I decided not to use ParameterSetName. I do not know if it is possible with ParameterSetName in a different way.

What do you think?

dpaulson45: After looking at this a little more, I would almost go the reverse. Enable switches to disable things to check to make it more like `.\Test-ExchAVExclusions.ps1` without anything just does everything by default behavior. Which is what you are going for by doing this section of code:

image

That would make it a little less complicated for end users to run. Then it wouldn't matter if `-Recurse` or `-IncludeW3wProcess` is enabled with the wrong set of parameters, because they also included a switch that deliberately disabled checks related to it.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -104,205 +206,432 @@ if (-not($exchangeShell.ShellLoaded)) {
...
+            $Offline = $true
         }
-        # If -recurse then we need to find all SubFolders and Add them to the list to be tested
-        if ($Recurse) {
-
-            # Add the root folder
-            $FolderList.Add($path.ToLower())
+    } catch {
+        $Offline = $true
+    }
```
dpaulson45, line 228: We can make this simpler:
try {
        $response = $null
        $Offline = $true 
        $response = Invoke-WebRequest http://crl.microsoft.com/pki/crl/products/CodeSigPCA.crl
      $Offline = $null -eq $response -or $response.StatusCode -ne 200
      } catch {
      Write-verbose "Failed web request with inner error: $_"
       }

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -104,205 +206,432 @@ if (-not($exchangeShell.ShellLoaded)) {
...
+                            if ( ($module.FileName.ToLower().StartsWith(('c:\windows\assembly\NativeImages_').ToLower()) -and (
+                                        $module.FileName.ToLower().EndsWith('.ni.dll') -or $module.FileName.ToLower().EndsWith('.ni.exe') -or
+                                        $module.FileName.ToLower().EndsWith('.wrapper.dll') ) ) -or
+                            ($module.FileName.ToLower().StartsWith('c:\windows\system32\') -and [environment]::OSVersion.Version.Major -le 6 ) -or
+                            ($module.FileName.ToLower().StartsWith(('c:\windows\WinSxS\').ToLower()) -and [environment]::OSVersion.Version.Major -le 6 ) -or
+                            ($module.FileName.ToLower().StartsWith('c:\windows\assembly\') -and [environment]::OSVersion.Version.Major -le 6 ) -or
+                            ($module.FileName.ToLower().StartsWith('c:\windows\microsoft.net\assembly\gac_64\') -and [environment]::OSVersion.Version.Major -le 6 ) -or
+                            ($module.FileName.ToLower() -eq (Join-Path ($env:ExchangeInstallPath).ToLower() 'bin\osafehtm.dll') -and ([byte]$serverExchangeInstallDirectory.MsiProductMinor) -gt 0 ) ) {
```
dpaulson45, line 439: Add comment for what we are trying to do here. This is extremely complicated and need more context here. But I believe this is testing for things that aren't signed for known modules that are on Windows 2012 R2 or below. 

We should just do a list of known unsigned files if it isn't long. As this would be better for security reasons.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -104,205 +206,432 @@ if (-not($exchangeShell.ShellLoaded)) {
...
+                                    $chain = New-Object -TypeName System.Security.Cryptography.X509Certificates.X509Chain
+                                    $chain.Build($cert) | Out-Null
+                                    $rootCertificate = $null
+                                    $rootCertificate = $chain.ChainElements[$chain.ChainElements.Count - 1].Certificate.Subject
+                                    if ( $null -eq $rootCertificate) {
+                                        $BadProcessList["$($Instance.Id) - $($module.FileName)"] = "$($Instance.Id), $($module.FileName), $($module.Company), $($module.Description), $($module.Product),  $($Instance.Path), $CommandLine"
+                                        Write-SimpleLogFile -String ("[FAIL] - We could not get root certificate on Module: $($module.FileName) `n`tCompany: $($module.Company) `n`tDescription: $($module.Description) `n`tProduct: $($module.Product) `n`tProcess ID: $($Instance.Id) `n`tCommand Line: $CommandLine") -name $LogFile -OutHost
+                                    } else {
+                                        $FIPSPath = Join-Path $env:ExchangeInstallPath 'FIP-FS\Bin\TE'
```
dpaulson45, line 466: Variable set and never changed inside of a loop. Best practice is to not do this.

dpaulson45, line 466: Using `$env:ExchangeInstallPath` but not a guarantee that it is there. We already define the value in `$ExchangePath` we should be using that.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -104,205 +206,432 @@ if (-not($exchangeShell.ShellLoaded)) {
...
+                                        if ($module.FileName.ToLower().StartsWith($FIPSPath.ToLower()) ) {
+                                            if ($Offline) {
+                                                if ( -not (CheckIfISAcceptedRootCA -CAString $rootCertificate -isFIPFS -Offline) ) {
+                                                    $BadProcessList["$($Instance.Id) - $($module.FileName)"] = "$($Instance.Id), $($module.FileName), $($module.Company), $($module.Description), $($module.Product),  $($Instance.Path), $CommandLine"
+                                                    Write-SimpleLogFile -String ("[FAIL] - root do not expected (FIP-FS) on Module: $($module.FileName) `n`tCompany: $($module.Company) `n`tDescription: $($module.Description) `n`tProduct: $($module.Product) `n`tRoot CA $rootCertificate `n`tProcess ID: $($Instance.Id) `n`tCommand Line: $CommandLine") -name $LogFile -OutHost
+                                                }
```
dpaulson45, line 472: Why are we checking to see if the module is signed correctly in this script? 

This script is to check for AV Exclusions the easiest way possible. If there is a threat concern that we are trying to validate here. That would need to be done in a different script, not in an AV Exclusions script.

dpaulson45 (changes requested):

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
iserrano76: We are checking that the modules are signed correctly by well-know CAs.
```
CN=Microsoft Corporate Root CA, O=Microsoft Corporation
CN=Microsoft Root Authority, OU=Microsoft Corporation, OU=Copyright (c) 1997 Microsoft Corp.
CN=Microsoft Root Certificate Authority, DC=microsoft, DC=com
CN=Microsoft Root Certificate Authority 2010, O=Microsoft Corporation, L=Redmond, S=Washington, C=US
CN=Microsoft Root Certificate Authority 2011, O=Microsoft Corporation, L=Redmond, S=Washington, C=US
```
In case or FIPS files Verisign is valid as well.
```
CN=VeriSign Class 3 Public Primary Certification Authority - G5, OU="(c) 2006 VeriSign, Inc. - For authorized use only", OU=VeriSign Trust Network, O="VeriSign, Inc.", C=US
```
I think it is better to use the signature than the company name in the file.
In this way, we can verify that the module is not from a different company and that our modules were not modified.

dpaulson45:
Closing PR, as this has already been updated.

PR #1644, opened 2023-03-31
CVE-2023-23397.ps1 failed to add API permission in Gallatin environment

DennyMei:
Describe the issue
Run .\https://github.com/advisories/GHSA-4778-q233-jmh5.ps1 -CreateAzureApplication -AzureEnvironmentName AzureChinaCloud
Application was created but the full_access_as_app permission was failed to add.

Expected behavior
full_access_as_app permission should be added to the application.

Script Output
No warning message for this.

Additional context
We found the name of Exchange APP is "Microsoft.Exchange" in Gallatin instead of "Office 365 Exchange Online" in WW.
So we should consider use Application ID instead of DisplayName in the following codes:
#Get Service Principal of MS Graph Resource API
$ews_SP = Get-AzureADServicePrincipal -All $true | Where-Object { $_.DisplayName -eq "Office 365 Exchange Online" }

Snapshot:
image

### DIFF
--- a/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
+++ b/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
@@ -587,7 +587,7 @@ begin {
         Add-AzureADApplicationOwner -ObjectId $aadApplication.ObjectId -RefObjectId $currentUser.ObjectId | Out-Null
 
         #Get Service Principal of MS Graph Resource API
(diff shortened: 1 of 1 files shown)


lusassl-msft:
This is what we plan to do with PR #1638 as well.

lusassl-msft:
Thanks for your contribution @DennyMei. We’ve made a similar change with the latest release of the script. Please give it try and let us know if it works as expected if run against Gallatin.

PR #1645, opened 2023-04-02
Don't include date filter in search folder criteria

bill-long:
Currently, if -StartTimeFilter or -EndTimeFilter is provided in combination with -UseSearchFolders, the results can be unexpected., because the search criteria is set at the time of search folder creation, and it never changes. Thus, the script always returns results based on the time filters provided during the first run against a given mailbox. Even if subsequent runs use different filters, the results never change, unless the search folders are cleaned up.

In this PR, we no longer include the time filters in the search folder criteria. The search folders will always contain all items that have the PidLidReminderFileParameter property. The script reads those results and then filters them if the time filter parameters are provided. This way, the script can be repeatedly run with different filters and return the correct results without having to recreate the search folders.


lusassl-msft (approved):
Change looks good to me

PR #1647, opened 2023-04-03
Update Exchange-Protection.md

sujithnair79:
**Issue:**
From line number 285 we should consider removing this as this step is not needed to be performed. 
.\ExchangeExtendedProtectionManagement.ps1 –RollbackType "RestoreIISAppConfig" 

**Reason:**
If accidently performed by end customer, they may be ending up with disabling the Extended Protection completely and they will have to rerun the script to make modification which can be time consuming.

**Fix:**
Edit the doc with updated step.

**Validation:**
Provide if applicable


bill-long:
This change looks good to me. We have the same command in two places, and the second one does seem out of place:

image

bill-long (approved):

lusassl-msft (approved):

bill-long:
@sujithnair79 Thank you!

Issue #1671, opened 2023-04-19
[Issue] Invoke-WebRequestWithProxyDetection needs parameter rework


bill-long:
Invoke-WebRequestWithProxyDetection can be called in a simple way by just passing -Uri, or it can be called by passing -ParametersObject with a hashtable. This is weird, because _any_ function can be called by passing a hashtable of parameters already, using [splatting](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting). With that approach, we still get the normal parameter handling and validation, instead of bypassing it with -ParametersObject.

Invoke-WebRequestWithProxyDetection should be changed so that all possible parameters are exposed as actual parameters to the function. Then, if you want to pass a hashtable, the user can use splatting to do so.

PR #1673, opened 2023-04-21
Log output improved to provide more context when troubleshooting web request scenarios

lusassl-msft:
**Description:**
We need to log out more information to make it easier to troubleshoot scenarios where `Invoke-WebRequest` is being used.

### DIFF
--- a/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
+++ b/Security/src/CVE-2023-23397/CVE-2023-23397.ps1
@@ -724,6 +724,8 @@ begin {
             }
         }
 
+        # Specify secret expiration time which must be in ISO 8601 format and is always in UTC time
+        $pwdEndDateTime = ([DateTime]::UtcNow).AddDays(7).ToString("o")
         # Graph API call to create a new application password
         $newAadApplicationPasswordParams = @{
             Query       = "applications/$($getAadApplicationResponse.value.id)/addPassword"
@@ -731,7 +733,7 @@ begin {
             Body        = @{
                 "passwordCredential" = @{
                     "displayName" = "AppAccessKey"
-                    "endDateTime" = (Get-Date).AddDays(7).ToString("yyyy-MM-ddTHH:mm:ssZ")
+                    "endDateTime" = $pwdEndDateTime
                 }
             } | ConvertTo-Json
             Method      = "POST"
@@ -914,6 +916,7 @@ begin {
     }
 } end {
     Write-Host ("CVE-2023-23397 script version $($BuildVersion)") -ForegroundColor Green
+    Write-Verbose "PowerShell version: $($PSVersionTable.PSVersion)"
 
     # Using either of these switches implies -UseSearchFolders
     if ($SearchFolderCleanup -or $SkipSearchFolderCreation) {
--- a/Shared/ScriptUpdateFunctions/Invoke-WebRequestWithProxyDetection.ps1
+++ b/Shared/ScriptUpdateFunctions/Invoke-WebRequestWithProxyDetection.ps1
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 
 . $PSScriptRoot\Confirm-ProxyServer.ps1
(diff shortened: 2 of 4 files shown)


Shared/ScriptUpdateFunctions/Invoke-WebRequestWithProxyDetection.ps1
```diff
@@ -47,5 +48,14 @@ function Invoke-WebRequestWithProxyDetection {
...
-    Invoke-WebRequest @params
+    try {
+        Invoke-WebRequest @params
+    } catch {
+        Write-Verbose $_.Exception.Message
+        Write-Verbose "Response: $($_.Exception.Response)"
+        Write-Verbose "Status: $($_.Exception.Status)"
+        Write-Verbose "Stack: $($_.Exception.StackTrace)"
+        Write-Verbose "CategoryInfo: $($_.CategoryInfo)"
+        Write-Verbose "FullyQualifiedErrorId: $($_.FullyQualifiedErrorId)"
```
dpaulson45, line 59: Is there a reason to not use `Shared\Write-ErrorInformation.ps1` here?

Security/src/CVE-2023-23397/CVE-2023-23397.ps1
```diff
@@ -724,14 +724,16 @@ begin {
             }
         }
 
+        # Specify secret expiration time which must be in ISO 8601 format and is always in UTC time
+        $pwdEndDateTime = ((Get-Date).AddDays(7)).ToString("yyyy-MM-ddTHH:mm:ssZ")
```
dpaulson45, line 728: This isn't UTC time.

image

dpaulson45: Also, why did we change it to be a variable if we only use it in the 1 location?

Shared/ScriptUpdateFunctions/Invoke-WebRequestWithProxyDetection.ps1
lusassl-msft: Will use the shared function in future as I plan to make some general changes to the function to make it shared (#1671).

Security/src/CVE-2023-23397/CVE-2023-23397.ps1
lusassl-msft: Hm wondering if this could be the reason for the HTTP 400 errors that we see. Good catch indeed!

bill-long: The correct way to do this is:

```powershell
[DateTime]::UtcNow.ToString("o")
```

lusassl-msft: ([DateTime]::UtcNow).AddDays(7).ToString("o")

lusassl-msft: Works as expected. Will make the change as suggested.

bill-long: Or if you prefer (Get-Date).ToUniversalTime().ToString("o"). Get-Date also returns a DateTime struct.

bill-long: Just saying we shouldn't be trying to manually produce ISO8601 format when DateTime already does that.

lusassl-msft: Yes, I wasn't aware that we had something ready. `([DateTime]::UtcNow).AddDays(7).ToString("o")` is fine.

Shared/ScriptUpdateFunctions/Invoke-WebRequestWithProxyDetection.ps1
lusassl-msft: Done.

dpaulson45 (approved):

PR #1696, opened 2023-05-08
Include Protocols name for Cipher Suites

### LINKED ISSUE #976: Health Checker - TLS Cipher include protocol
**Describe The Request**
Include the protocol that the cipher supports.

image

Get the protocol version values from here: 

https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptderivekey#bcrypt_kdf_tls_prf-ltls_prf

dpaulson45:
**Issue:**
Requested to include the supported protocols for the cipher suite.

**Reason:**
Make it easier to know what cipher is for what protocol. 

**Fix:**
Resolved #976 

**Validation:**
Lab tested

image

### DIFF
--- a/Shared/TLS/Get-TlsCipherSuiteInformation.ps1
+++ b/Shared/TLS/Get-TlsCipherSuiteInformation.ps1
@@ -28,7 +28,6 @@ function Get-TlsCipherSuiteInformation {
                 elseif ($p -eq 770) { $name = "TLS_1_1" }
                 elseif ($p -eq 771) { $name = "TLS_1_2" }
                 elseif ($p -eq 772) { $name = "TLS_1_3" }
-                elseif ($p -eq 256) { $name = "PROTOCOL_DTLS_1_0_OPENSSL_PRE_0_9_8f" }
                 elseif ($p -eq 32528) { $name = "TLS_1_3_DRAFT_16" }
                 elseif ($p -eq 32530) { $name = "TLS_1_3_DRAFT_18" }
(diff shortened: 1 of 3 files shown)


Shared/TLS/Get-TlsCipherSuiteInformation.ps1
```diff
@@ -12,6 +12,37 @@ function Get-TlsCipherSuiteInformation {
...
+        function GetProtocolNames {
+            param(
+                [int[]]$Protocol
+            )
+            $protocolNames = New-Object System.Collections.Generic.List[string]
+
+            foreach ($p in $Protocol) {
+                $name = [string]::Empty
+
+                if ($p -eq 2) { $name = "SSL_2_0" }
```
lusassl-msft, line 25: Do we have a reference for this mapping that we can add as a comment?

dpaulson45: Not a Microsoft source. 

https://github.com/tintinweb/scapy-ssl_tls/blob/8fde2f50d0849f97ae7916d9e985a6595fe0999f/scapy_ssl_tls/ssl_tls.py#L249

lusassl-msft: Hm yeah, looks like this is specified within the RFC per version. A good overview can also be found here: https://tlsfingerprint.io/top/versions 
I think we should be good then.

lusassl-msft (approved):

PR #1703, opened 2023-05-09
Added control in case 3rd party AV

iserrano76:
**Issue:**
If 3rd party AV disabled Microsoft Defender you got an error trying to use Get-MpComputerStatus

**Reason:**
To avoid errors in the execution and add explanation on output.

**Fix:**
 I added a control to avoid errors and explain that it is not possible.

**Validation:**
Test in lab

### DIFF
--- a/Setup/SetExchAVExclusions/Set-ExchAVExclusions.ps1
+++ b/Setup/SetExchAVExclusions/Set-ExchAVExclusions.ps1
@@ -108,16 +108,16 @@ if (-not $ListRecommendedExclusions) {
     $checkCmdLet = $null
     $checkCmdLet = Get-Command Get-MpComputerStatus -ErrorAction SilentlyContinue
     if ($null -eq $checkCmdLet) {
-        Write-Error "Get-MpComputerStatus cmdLet is not available"
-        Write-Error "This script only sets Exclusions on Microsoft Defender"
+        Write-Host "Get-MpComputerStatus cmdLet is not available" -ForegroundColor Red
+        Write-Host "This script only sets Exclusions on Microsoft Defender" -ForegroundColor Red
         Write-Host "If you have any other Antivirus you can use -ListRecommendedExclusions parameter to get the Recommended Exclusion List"
         exit
     } else {
         $mpStatus = $null
         $mpStatus = Get-MpComputerStatus -ErrorAction SilentlyContinue
         if ($null -eq $mpStatus) {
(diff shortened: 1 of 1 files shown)


Setup/SetExchAVExclusions/Set-ExchAVExclusions.ps1
```diff
@@ -105,15 +105,27 @@ if (-not $ListRecommendedExclusions) {
...
 
-    $mpStatus = $null
-    $mpStatus = Get-MpComputerStatus -ErrorAction SilentlyContinue
-    if ($null -eq $mpStatus) {
-        Write-Error "We cannot get Microsoft Defender information"
+    $checkCmdLet = $null
+    $checkCmdLet = Get-Command Get-MpComputerStatus -ErrorAction SilentlyContinue
+    if ($null -eq $checkCmdLet) {
+        Write-Error "Get-MpComputerStatus cmdLet is not available"
+        Write-Error "This script only sets Exclusions on Microsoft Defender"
```
bill-long, line 112: We should change all the Write-Error calls to either Write-Warning or Write-Host -ForegroundColor Red. Write-Error is awkward because it produces multiple lines of output. This means writing one error is noisy and writing two is even worse:

image

We should only Write-Error if this is really a fatal error where we're going to need a call stack. line number, etc. For this output, we don't need that.

Setup/SetExchAVExclusions/Set-ExchAVExclusions.ps1
```diff
@@ -105,15 +105,27 @@ if (-not $ListRecommendedExclusions) {
...
         exit
     } else {
-        if (-not $mpStatus.AntivirusEnabled ) {
-            Write-Warning "Microsoft Defender is not enabled."
-            Write-Warning "We will apply the exclusions but they do not take effect until you Enabled Microsoft Defender."
+        $mpStatus = $null
+        $mpStatus = Get-MpComputerStatus -ErrorAction SilentlyContinue
+        if ($null -eq $mpStatus) {
+            Write-Error "We cannot get Microsoft Defender information"
+            Write-Error "This script only sets Exclusions on Microsoft Defender"
```
bill-long, line 120: Same as the other. Let's avoid Write-Error, especially multiple of them.

bill-long (changes requested):

bill-long (approved):

PR #1718, opened 2023-05-15
Have HTML Report use OutputFilePath directory

### LINKED ISSUE #1717: [Issue] - Health Checker - -OutputFilePath not honored for -BuildHtmlServersReport
**Provide Version Number**
Version 23.05.08.1516

**[DO-OVER] Describe the issue**
When supplying -BuildHtmlServersReport and -OutputFilePath, -OutputFilePath is not honored, instead current directory is where it lands.  This seems to be due to `Out-File` on line 7835 (as of this version).

**Describe the issue**
When I supply -OutputFilePath, the destination that I supply is only honored for certain outputs.  For example, the server TXT and XML files land where I want, but the Debug file, as well as the script updater log, both land in the script calling user's $PWD.  When I try to figure out what's going on in the code, it looks like this should be handled by the function Invoke-SetOutputInstanceLocation, which seems to be called in all the right places.  But it's clearly not working.  The thing that that is most pressing is that the HTML report is landing somewhere other than what is specified in -OutputFilePath (lands in System32 when I use Scheduled Tasks).

**Expected behavior**
I expect that all of the script's output files would land in the specified OutputFilePath.

**Script Output**
N/A, I'm not getting errors and the script is up to date.

**Additional context**
N/A

dpaulson45:
**Issue:**
#1717 

**Reason:**
The HTML Report should be used as a file name and if `OutputFilePath` is used, combined the two in order to get the desired results. 

**Fix:**
Detect if `$HtmlReportFile` is already a file path, if it is use it. Otherwise, combined `$OutputFilePath` and `$HtmlReportFile` to get the desired result. 

Resolved #1717 

**Validation:**
Lab tested

### DIFF
--- a/Diagnostics/HealthChecker/Features/Get-HtmlServerReport.ps1
+++ b/Diagnostics/HealthChecker/Features/Get-HtmlServerReport.ps1
@@ -3,7 +3,10 @@
 
 function Get-HtmlServerReport {
     param(
-        [Parameter(Mandatory = $true)][array]$AnalyzedHtmlServerValues
+        [Parameter(Mandatory = $true)]
+        [array]$AnalyzedHtmlServerValues,
+
+        [string]$HtmlOutFilePath
     )
     Write-Verbose "Calling: $($MyInvocation.MyCommand)"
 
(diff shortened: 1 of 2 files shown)


lusassl-msft (approved):

JeremyTBradshaw:
@dpaulson45, @lusassl-msft thanks for the amazing and quick support on this.  One question - I see there's a new Release today, but the HealthChecker.ps1 version is still the same (23.05.08.1516).  Will I just need to wait and watch for the next release to come, or do you think something didn't complete with the new release that came out today?

PR #1546, opened 2023-03-06
Adding some controls to avoid errors in the execution

iserrano76:
**Issue:**
You can received errors in the execution if you run it in a Environment without 2016/2019 or in old Windows Versions

**Reason:**
You can run the script in a Exchange 2013 or Windows 2012 R2 or older and you received errors.
Additionally, you can run disabled or enabled twice and receive errors

**Fix:**
Added Windows and Exchange version checks and controls on which versions apply.
Included verification if the disabling is already included so as not to re-launch it and give an error.
Included a parameter to see the current status.

**Validation:**
Test in lab

PR #1722, opened 2023-05-17
Replacement of #1546

### LINKED ISSUE #651: [Work Item] add web.config check
Need to add a check to validate all the web.config files have the HttpRequestFilteringModule in them

### LINKED ISSUE #1724: New Features on Test-AMSI script.
The current Test-AMSI.ps1 script performs the following actions:

- Test: Do a http request to ecp/x.js on an end Point and depending on the response, considers that AMSI is working or not.
- CheckAMSIProviders: List the AMSI provider information on local server.
- DisableAMSI: Create a SettingOverwrite to disable AMSI at organization level.
- EnableAMSI: Delete the previous SettingOverwrite to disable AMSI at organization level.
- RestartIIS: Restart IIS on local server.

In the changes I want to introduce: 
- Exchange versions and windows versions verification. Windows 2016 and Exchange 2016 CU21+/ 2019 CU11+ are required.
- You could run all option on local server, on remote server, on a list of servers, on all servers in the organization, on all servers in a site. With those options you can split the application of changes and tests by machines, sites or for all servers.
- Check AMSIProviders:
	- The original script shows only one AMSI provider, in case of multiple providers it does not show information. This has been corrected in the new one and show all providers.
	- Check the web.config files (FrontEnd and ClientAccess) to confirm that it includes the HttpRequestFilteringModule.
	- If an AMSI provider is Windows Defender, we check that it is an AMSI compliant version (1.1.18300.4+). 
	- Verify the configuration override at organization level and at server level.
- TestAMSI:
	- The script detects whether the endpoint is an Exchange server or a name-sharing endpoint.
	- If it is an Exchange server, we confirm that the detection works by checking the LoggingHttpRequestFiltering files. 
- DisableAMSI: 
	- We check if you already have any overridden settings that disable AMSI.
	- You can disable AMSI at  organization level and at server level. 
	- Server level is useful in case of testing to avoid impact the whole organization.
- Enable AMSI:
	- Check if you already have any configuration that enables AMSI.
	- You can disable AMSI at organization level and at  server level. 
	- Server level is useful in case of testing to avoid impact the whole organization.
- Include warnings in case of multisite environment to replicate and link to Exchange blog about AMSI.

More about AMSI integration with Exchange Server
https://techcommunity.microsoft.com/t5/exchange-team-blog/more-about-amsi-integration-with-exchange-server/ba-p/2572371

iserrano76:
Replacement of #1546

Issue:
You can received errors in the execution if you run it in a Environment without 2016/2019 or in old Windows Versions
and adding options

Reason:
You can run the script in a Exchange 2013 or Windows 2012 R2 or older and you received errors.
Additionally, you can run disabled or enabled twice and receive errors
Resolve #1724
Resolve #651 

Fix:
Added Windows and Exchange version checks and controls on which versions apply.
Included verification if the disabling is already included so as not to re-launch it and give an error.
Included a parameter to see the current status.

Validation:
Test in lab


Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                $msgDetectedTimeStamp = "You should find result around $((Get-Date).ToUniversalTime().ToString("M/d/yyy h:mm:ss tt")) UTC"
                 Write-Output $msgDetectedTimeStamp
+                if ( $isServer ) {
+                    Write-Output ""
+                    Write-Host "Checking logs on $server at $($testTime.ToString("M/d/yyy h:mm:ss tt"))"
+                    $HttpRequestFilteringLogFolder = $null
+                    if ( $server.ToLower() -eq $env:COMPUTERNAME.ToLower() ) {
+                        $HttpRequestFilteringLogFolder = Join-Path $env:ExchangeInstallPath "Logging\HttpRequestFiltering\"
+                    } else {
+                        $remoteExchangePath = (Invoke-Command -ComputerName $server -ScriptBlock { (Get-ChildItem Env:ExchangeInstallPath).Value } -ErrorAction SilentlyContinue -ErrorVariable InvokeError)
```
dpaulson45, line 216: This line is duplicated. Avoid duplication at all costs to avoid bugs.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                                            $folder = Get-ChildItem  $WindowsDefenderPath | Sort-Object LastWriteTime -Descending | Select-Object -First 1
+                                            $process = Join-Path $folder.FullName "MpCmdRun.exe"
+                                            if ( Test-Path $process -PathType Leaf ) {
+                                                $DefenderVersion = (& $process -SignatureUpdate | Where-Object { $_.StartsWith('Engine Version:') }).Split(' ')[2]
+                                                if ( [int]$DefenderVersion.Split('.')[0] -gt 1 -or
+                                            ( [int]$DefenderVersion.Split('.')[0] -eq 1 -and [int]$DefenderVersion.Split('.')[1] -gt 1 ) -or
+                                            ( [int]$DefenderVersion.Split('.')[0] -eq 1 -and [int]$DefenderVersion.Split('.')[1] -eq 1 -and [int]$DefenderVersion.Split('.')[2] -gt 18300 ) -or
+                                            ( [int]$DefenderVersion.Split('.')[0] -eq 1 -and [int]$DefenderVersion.Split('.')[1] -eq 1 -and [int]$DefenderVersion.Split('.')[2] -eq 18300 -and [int]$DefenderVersion.Split('.')[3] -ge 4) ) {
```
dpaulson45, line 440: We should be using `System.Version` for comparison here.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+            }
+
+            if ( $FEEcpWebConfig -and $CAEEcpWebConfig) {
+                if ( Test-Path $FEEcpWebConfig -PathType Leaf) {
+                    $FEFilterModule = $null
+                    $FEFilterModule = Get-Content $FEEcpWebConfig | Where-Object { $_ -match '<add name="HttpRequestFilteringModule" type="Microsoft.Exchange.HttpRequestFiltering.HttpRequestFilteringModule, Microsoft.Exchange.HttpRequestFiltering, Version=15.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35"' }
+                    Write-Output ""
+                    if ( $FEFilterModule ) {
+                        Write-Host "We found HttpRequestFilteringModule on FrontEnd ECP web.config" -ForegroundColor Green
+                        Write-Host "on: %ExchangeInstallPath%FrontEnd\HttpProxy\ecp\web.config"
```
dpaulson45, line 508: Use the actual location stored in variable.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                    } else {
+                        Write-Warning "We did not find HttpRequestFilteringModule on FrontEnd ECP web.config"
+                        Write-Warning "on: %ExchangeInstallPath%FrontEnd\HttpProxy\ecp\web.config"
```
dpaulson45, line 511: Use the actual location stored in variable.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                    }
+                } else {
+                    Write-Warning "We did not find web.config for FrontEnd ECP"
+                    Write-Warning "on: %ExchangeInstallPath%FrontEnd\HttpProxy\ecp\web.config"
```
dpaulson45, line 515: Use the actual location stored in variable.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                if ( Test-Path $FEEcpWebConfig -PathType Leaf) {
+                    $CEFilterModule = $null
+                    $CEFilterModule = Get-Content $CAEEcpWebConfig | Where-Object { $_ -match '<add name="HttpRequestFilteringModule" type="Microsoft.Exchange.HttpRequestFiltering.HttpRequestFilteringModule, Microsoft.Exchange.HttpRequestFiltering, Version=15.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35"' }
+                    Write-Output ""
+                    if ( $CEFilterModule ) {
+                        Write-Host "We found HttpRequestFilteringModule on ClientAccess ECP web.config" -ForegroundColor Green
+                        Write-Host "on %ExchangeInstallPath%ClientAccess\ecp\web.config"
+                    } else {
+                        Write-Warning "We did not find HttpRequestFilteringModule on ClientAccess ECP web.config"
+                        Write-Warning "on %ExchangeInstallPath%ClientAccess\ecp\web.config"
```
dpaulson45, line 527: Use the actual location stored in variable.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                    }
+                } else {
+                    Write-Warning "We did not find web.config for ClientAccess ECP"
+                    Write-Warning "on %ExchangeInstallPath%ClientAccess\ecp\web.config"
```
dpaulson45, line 531: Use the actual location stored in variable.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                        Write-Warning "on: %ExchangeInstallPath%FrontEnd\HttpProxy\ecp\web.config"
+                    }
+                } else {
+                    Write-Warning "We did not find web.config for FrontEnd ECP"
+                    Write-Warning "on: %ExchangeInstallPath%FrontEnd\HttpProxy\ecp\web.config"
+                }
+
+                if ( Test-Path $FEEcpWebConfig -PathType Leaf) {
+                    $CEFilterModule = $null
+                    $CEFilterModule = Get-Content $CAEEcpWebConfig | Where-Object { $_ -match '<add name="HttpRequestFilteringModule" type="Microsoft.Exchange.HttpRequestFiltering.HttpRequestFilteringModule, Microsoft.Exchange.HttpRequestFiltering, Version=15.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35"' }
```
dpaulson45, line 520: If all we are doing is a string match, use `Select-String` instead.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                if ( $remoteExchangePath ) {
+                    $FEEcpWebConfig = Join-Path "\\$ExchangeServer\$($remoteExchangePath.Replace(':','$'))" "FrontEnd\HttpProxy\ecp\web.config"
+                    $CAEEcpWebConfig = Join-Path "\\$ExchangeServer\$($remoteExchangePath.Replace(':','$'))" "ClientAccess\ecp\web.config"
+                }
+            }
+
+            if ( $FEEcpWebConfig -and $CAEEcpWebConfig) {
+                if ( Test-Path $FEEcpWebConfig -PathType Leaf) {
+                    $FEFilterModule = $null
+                    $FEFilterModule = Get-Content $FEEcpWebConfig | Where-Object { $_ -match '<add name="HttpRequestFilteringModule" type="Microsoft.Exchange.HttpRequestFiltering.HttpRequestFilteringModule, Microsoft.Exchange.HttpRequestFiltering, Version=15.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35"' }
```
dpaulson45, line 504: If all we are doing is a string match, use `Select-String` instead.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+        }
+        if ( $null -eq $AMSIProviders ) {
+            Write-Warning "We did not find any AMSI Provider"
+        } else {
+            foreach ( $provider in $AMSIProviders) {
+                $provider -match '[0-9A-Fa-f\-]{36}' | Out-Null
+                foreach ($m in $Matches.Values ) {
+                    $key = "HKLM:\SOFTWARE\Classes\ClSid\{$m}"
+                    Write-Output $msgNewLine
+                    if ( $ExchangeServer.ToLower() -eq $env:COMPUTERNAME.ToLower() ) {
```
dpaulson45, line 354: Utilize the shared function of `Invoke-ScriptBlockHandler` in `Shared` directory.

dpaulson45 (changes requested):

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                            Write-Warning "Cannot get Remote Exchange installation path on $server"
+                        }
+                    }
+                    if ( Test-Path $HttpRequestFilteringLogFolder -PathType Container ) {
+                        $file = $null
+                        $timeout = (Get-Date).AddMinutes(1)
+                        $detected = $false
+                        do {
+                            $file = $null
+                            Get-ChildItem $HttpRequestFilteringLogFolder | Sort-Object LastWriteTime -Descending | Select-Object -First 1 -Property * | Out-Null
```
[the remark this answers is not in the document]
iserrano76: It is an old one no need it anymore, I will remove it

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                                    } catch {
+                                        Write-Verbose ("We could not parse the date time on: {0}" -f $line)
+                                    }
+                                    if ( $DateTime ) {
+                                        $marginTime = New-TimeSpan -Seconds 5
+                                        if ($testTime.ToUniversalTime().Subtract($DateTime) -lt $marginTime -and $testTime.ToUniversalTime().Subtract($DateTime) -gt - $marginTime ) {
+                                            if ( $line.UrlHost.ToLower() -eq $server.ToLower() -and $line.UrlStem.ToLower() -eq '/ecp/x.js'.ToLower() -and $line.ScanResult.ToLower() -eq 'Detected'.ToLower() ) {
+                                                Write-Output ""
+                                                Write-Host "We found a detection in HttpRequestFiltering logs: " -ForegroundColor Green
+                                                Write-Host "$line"
```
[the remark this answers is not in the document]
iserrano76: It comes from the original script and It is fine display, but it will be better to use allways the same. I will change it.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+                        $remoteExchangePath = (Invoke-Command -ComputerName $server -ScriptBlock { (Get-ChildItem Env:ExchangeInstallPath).Value } -ErrorAction SilentlyContinue -ErrorVariable InvokeError)
+                        if ( $remoteExchangePath ) {
+                            $HttpRequestFilteringLogFolder = Join-Path "\\$server\$($remoteExchangePath.Replace(':','$'))" "Logging\HttpRequestFiltering\"
+                        } else {
+                            Write-Warning "Cannot get Remote Exchange installation path on $server"
+                        }
+                    }
+                    if ( Test-Path $HttpRequestFilteringLogFolder -PathType Container ) {
+                        $file = $null
+                        $timeout = (Get-Date).AddMinutes(1)
```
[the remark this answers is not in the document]
iserrano76: I think it is really weird that reach the timeout, it is just wait for write the HttpRequestFiltering and takes just a few seconds. I think one minute is more than enough and maybe a parameter for only that part could be confusing. I am not sure if it is necessary but I can add it.

Admin/Test-AMSI.ps1
iserrano76: In previous PR I used the variable, but I was not sure it is better and finally I change it. 
I will change back.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
-                $msgDetectedTimeStamp = "for a Detected result around " + $dateTime.ToUniversalTime()
+                $host.ui.RawUI.ForegroundColor = $currentForegroundColor
+                $msgDetectedTimeStamp = "You should find result around $((Get-Date).ToUniversalTime().ToString("M/d/yyy h:mm:ss tt")) UTC"
                 Write-Output $msgDetectedTimeStamp
+                if ( $isServer ) {
+                    Write-Output ""
+                    Write-Host "Checking logs on $server at $($testTime.ToString("M/d/yyy h:mm:ss tt"))"
+                    $HttpRequestFilteringLogFolder = $null
+                    if ( $server.ToLower() -eq $env:COMPUTERNAME.ToLower() ) {
+                        $HttpRequestFilteringLogFolder = Join-Path $env:ExchangeInstallPath "Logging\HttpRequestFiltering\"
```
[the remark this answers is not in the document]
iserrano76: It will execute that line only if you are running the script in the same server, so it should local.
is that what you mean?
It is working in my tests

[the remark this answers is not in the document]
dpaulson45: It won't work if you launch PowerShell.exe and run the script. `Confirm-ExchangeShell` will load EMS, and it won't set that variable as far as I know.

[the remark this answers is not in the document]
iserrano76: In all test that I did it works, including in a powershell before and after load `Confirm-ExchangeShell`:
image
As it is a Environment variable and that line is only executed in case that we are in the expected Exchange server (locally) I think it should be available in any case. I am assuming that all Exchange server has the ExchangeInstallPath variable. am I wrong?
In any case, if you think it is better, I can use `HKLM:\SOFTWARE\Microsoft\ExchangeServer\v15\Setup\MsiInstallPath` instead.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
+    if ( $PSCmdlet.ParameterSetName -eq "TestAMSI" -or $PSCmdlet.ParameterSetName -eq "TestAMSIAll" ) {
+        $TestAMSI = $true
+    }
+
+    if ( $null -eq $ServerList -and ( $RestartIIS -or $CheckAMSIConfig -or $TestAMSI ) -and ( -not $AllServers ) ) {
+        $ServerList = $env:COMPUTERNAME
+    }
+
+    Write-Host "AMSI only support on Exchange 2016 CU21 or newer, Exchange 2019 CU10 or newer and running on Windows 2016 or newer" -ForegroundColor Green
+    $SupportedExchangeServers = Get-ExchangeServer | Where-Object { $_.IsClientAccessServer -and ( ($_.AdminDisplayVersion.Minor -eq 1 -and $_.AdminDisplayVersion.Build -ge 2308) -or ($_.AdminDisplayVersion.Minor -eq 2 -and $_.AdminDisplayVersion.Build -ge 922) ) } | Select-Object Name, Site
```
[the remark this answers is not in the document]
iserrano76: Ok,
If I run the script in an Exchange server, I do not have any issue, it returns ServerVersion Data in any case, including with simple PowerShell with `Confirm-ExchangeShell`:
image
and to be honest I always run it on an Exchange server :)
but if I did it in a computer with management tools I have that error.
I will change it to handle it if it is a string.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,782 @@
...
-    )
-
-    if ($Ignore) {
-        Add-Type @"
-        using System;
-        using System.Net;
-        using System.Net.Security;
-        using System.Security.Cryptography.X509Certificates;
-        public class ServerCertificateValidationBehavior
+    Add-Type @"
```
lusassl-msft, line 145: We have the `Enable-TrustAnyCertificateCallback.ps1` shared function (can be found here: `Shared/CertificateFunctions/Enable-TrustAnyCertificateCallback.ps1`) which could be used here.

Admin/Test-AMSI.ps1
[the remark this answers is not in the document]
dpaulson45: Wasn't aware of this, not sure if this didn't work in order OS/PowerShell. But it definitely works now so we can keep it as long as we are on the Exchange Server.

iserrano76:
Resolve #1724

dpaulson45:
> Resolve #1724

I think this needs to be in the main comment description to have it work.

lusassl-msft:
> > Resolve #1724
> 
> I think this needs to be in the main comment description to have it work.

Yup, looks good now: 
image

iserrano76:
and the issue appear as linked as well:
image
good to know :)

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+    . $PSScriptRoot\..\Shared\Invoke-ScriptBlockHandler.ps1
+    . $PSScriptRoot\..\Shared\CertificateFunctions\Enable-TrustAnyCertificateCallback.ps1
+
+    function Get-ExchangeInstallPath {
+        param(
+            [Parameter(Mandatory = $false)]
+            [string]$ExchangeServer
+        )
+        $getMSIInstallPathSB = { (Get-ItemProperty -Path Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\ExchangeServer\v15\Setup -ErrorAction SilentlyContinue).MsiInstallPath }
+        if ($ExchangeServer) {
```
lusassl-msft, line 156: This is not needed as `Invoke-ScriptblockHandler` contains the logic to determine if the call is executed against the local or remote machine. 

`Invoke-ScriptBlockHandler -ComputerName $ExchangeServer -ScriptBlock { (Get-ItemProperty -Path Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\ExchangeServer\v15\Setup -ErrorAction SilentlyContinue).MsiInstallPath }`

should do the job

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+                Write-Host "If you are using a 3rd Party AntiVirus Product that may not be AMSI capable (Please Check with your AntiVirus Provider for Exchange AMSI Support)"
+                $host.ui.RawUI.ForegroundColor = $currentForegroundColor
+            } elseif ( $_.Exception.Message.StartsWith("The remote name could not be resolved:") ) {
+                $host.ui.RawUI.ForegroundColor = "Red"
+                Write-Host $msgNewLine
+                Write-Host $Message
+                Write-Host $msgNewLine
+                $host.ui.RawUI.ForegroundColor = $currentForegroundColor
+            } else {
+                $host.ui.RawUI.ForegroundColor = "Red"
```
lusassl-msft, line 287: We should get rid of the `$host.ui.RawUI.ForegroundColor` stuff in the script and instead, simply do a `Write-Host -ForegroundColor Red`

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+        Write-Host "Ended test at $(Get-Date -Format "yyyy-MM-dd HH:mm:ss")"
+    }
+
+    function GetWindowsMayorVersion {
+        param(
+            [Parameter(Mandatory = $true)]
+            [string]$ExchangeServer
+        )
+
+        if ( $ExchangeServer.ToLower() -eq $env:COMPUTERNAME.ToLower() ) {
```
lusassl-msft, line 310: The function can be simplified by just using `Invoke-ScriptblockHandler`. No need to use `Get-CimInstance` or `Get-WmiObject`

`Invoke-ScriptBlockHandler -ComputerName $ExchangeServer -ScriptBlock { [System.Environment]::OSVersion.Version.Major }`

image

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
             }
         }
-        return
     }
-    if ($CheckAMSIProviders) {
-        $AMSI = Get-ChildItem 'HKLM:\SOFTWARE\Microsoft\AMSI\Providers' -Recurse
-        $AMSI -match '[0-9A-Fa-f\-]{36}' | Out-Null
-        $Matches.Values | ForEach-Object { Get-ChildItem "HKLM:\SOFTWARE\Classes\ClSid\{$_}" | Format-Table -AutoSize }
+
+    function GetExchangeVersion {
```
lusassl-msft, line 326: We have a shared function for this: `Get-ExchangeBuildVersionInformation.ps1`

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+
+    if ( $null -eq $ServerList -and ( $RestartIIS -or $CheckAMSIConfig -or $TestAMSI ) -and ( -not $AllServers ) ) {
+        $ServerList = $env:COMPUTERNAME
+    }
+
+    Write-Host ""
+    Write-Host "AMSI only support on Exchange 2016 CU21 or newer, Exchange 2019 CU10 or newer and running on Windows 2016 or newer" -ForegroundColor Green
+    Write-Host ""
+
+    $SupportedExchangeServers = Get-ExchangeServer | Where-Object {
```
lusassl-msft, line 538: Just use the shared function `Get-ExchangeBuildVersionInformation.ps1` to query the build information and use them to find out if we are running the correct version:

image
image
image

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+            } else {
+                Write-Warning "We did not find any Exchange server with name: $server"
+                if ( $TestAMSI ) {
+                    $filterList += $server
+                }
+            }
+        }
+    }
+
+    if ( ( ( $filterList.count -gt 0 ) -or
```
lusassl-msft, line 606: Nit: sometimes you use space between the brackets and sometimes you don't. I'd recommend using the same syntax everywhere in the script (I personally prefer not to use space, and this is what we do in the other scripts on CSS-Exchange too).

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+            $TestAMSI -or
+           ( ( $EnableAMSI -or $DisableAMSI) -and
+            -not $ServerList ) ) -and
+        $SupportedExchangeServers.count -gt 0 ) {
+
+        if ($TestAMSI) {
+            foreach ($server in $filterList) {
```
lusassl-msft, line 613: Nit: example where no space is used

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+                    Write-Host "Checking $($server):" -ForegroundColor Magenta
+                    Write-Host ""
+                    CheckAMSIConfig -ExchangeServer $server
+                    Write-Host ""
+                }
+            }
+            Write-Host $bar
+            Write-Host ""
+            $getSO = $null
+            $getSO = Get-SettingOverride -ErrorAction SilentlyContinue | Where-Object {
```
lusassl-msft, line 641: We have the `Get-FilteredSettingOverrideInformation.ps1` function that we use in `HealthChecker` to filter out `Setting Overrides`. However, the function is not yet a shared function but could be simply changed to become one.
Might be useful for this script and other future scripts as well. What do you think @dpaulson45

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+        }
+
+        $needsRefresh = 0
+        if ($EnableAMSI) {
+            $getSO = $null
+            if ($filterList) {
+                foreach ( $server in $filterList ) {
+                    Write-Host $bar
+                    Write-Host ""
+                    $getSO = Get-SettingOverride -ErrorAction SilentlyContinue | Where-Object {
```
lusassl-msft, line 667: `Get-FilteredSettingOverrideInformation.ps1` could be useful here as well

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+                        Write-Host "Enabled on $server" -ForegroundColor Green
+                        Get-ExchangeDiagnosticInfo -Process Microsoft.Exchange.Directory.TopologyService -Component VariantConfiguration -Argument Refresh -Server $server | Out-Null
+                        Write-Host "Refreshed Get-ExchangeDiagnosticInfo on $server"
+                        $needsRefresh++
+                    }
+                }
+            } else {
+                Write-Host $bar
+                Write-Host ""
+                $getSO = Get-SettingOverride -ErrorAction SilentlyContinue | Where-Object {
```
lusassl-msft, line 687: `Get-FilteredSettingOverrideInformation.ps1` could be useful here as well

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+            }
+        }
+
+        if ($DisableAMSI) {
+            $getSO = $null
+            if ($filterList) {
+                foreach ($server in $filterList) {
+                    Write-Host $bar
+                    Write-Host ""
+                    $getSO = Get-SettingOverride -ErrorAction SilentlyContinue | Where-Object {
```
lusassl-msft, line 714: `Get-FilteredSettingOverrideInformation.ps1` could be useful here as well

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+                        Write-Host "Refreshed Get-ExchangeDiagnosticInfo on $server"
+                        $needsRefresh++
+                    } else {
+                        Write-Warning "AMSI is already disabled on Exchange configuration for $server by SettingOverride $($getSO.Identity)"
+                    }
+                }
+            } else {
+                Write-Host $bar
+                Write-Host ""
+                $getSO = Get-SettingOverride -ErrorAction SilentlyContinue | Where-Object {
```
lusassl-msft, line 731: `Get-FilteredSettingOverrideInformation.ps1` could be useful here as well

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+            if ($filterList) {
+                foreach ($server in $filterList) {
+                    Write-Host $bar
+                    Write-Host ""
+                    $getSO = Get-SettingOverride -ErrorAction SilentlyContinue | Where-Object {
+                        ( $_.SectionName -eq 'HttpRequestFiltering' ) -and
+                        ( $_.Parameters -eq 'Enabled=False' ) -and
+                        ( $_.Server -contains $server ) }
+                    if ( $null -eq $getSO ) {
+                        New-SettingOverride -Name "DisablingAMSIScan-$server" -Component Cafe -Section HttpRequestFiltering -Parameters ("Enabled=False") -Reason "Disabled via CSS-Exchange Script" -Server $server
```
lusassl-msft, line 719: Maybe we should use `$PSCmdlet.ShouldProcess` here? Not sure how `New-SettingOverride` behaves if someone uses `WhatIf`.
https://learn.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-shouldprocess?view=powershell-7.3

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,793 @@
...
+                }
+            }
+        }
+    } else {
+        Write-Warning "We did not find Exchange servers with AMSI support"
+    }
+    Write-Host $bar
+    Write-Host ""
+    Write-Host 'You can find additional information at:'
+    Write-Host 'https://techcommunity.microsoft.com/t5/exchange-team-blog/more-about-amsi-integration-with-exchange-server/ba-p/2572371' -ForegroundColor Cyan
```
lusassl-msft, line 791: It might be better to create an aka.ms short link for this (not sure if we already have one for the AMSI blog post)?

lusassl-msft (changes requested):

Admin/Test-AMSI.ps1
iserrano76: It is something that comes from the original version but I think that David suggested to do it in another PR.

Admin/Test-AMSI.ps1
iserrano76: I changed it to show:
image

Admin/Test-AMSI.ps1
iserrano76: That's fine with me.
Do you want me to do it or do you do it since you have ownership of the previous ones?

lusassl-msft: @tweekerz @ninobb do you know if we already have a short-link for `https://techcommunity.microsoft.com/t5/exchange-team-blog/more-about-amsi-integration-with-exchange-server/ba-p/2572371` ?

lusassl-msft: @iserrano76 it doesn't look like we have one and so, I've created one: https://aka.ms/ExchangeAMSI

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,813 @@
...
+        $ServerList = $env:COMPUTERNAME
+    }
+
+    Write-Host ""
+    Write-Host "AMSI only support on Exchange 2016 CU21 or newer, Exchange 2019 CU10 or newer and running on Windows 2016 or newer" -ForegroundColor Green
+    Write-Host ""
+
+    $SupportedExchangeServers = Get-ExchangeServer | Where-Object {
+        ($_.IsClientAccessServer) -and
+        ((((Get-ExchangeBuildVersionInformation -AdminDisplayVersion ($_.adminDisplayVersion.ToString())).BuildVersion.Minor -eq 1) -and
```
lusassl-msft, line 539: I think this could be simplified (maybe like this - have not tested)

```
$exchangeServer = New-Object 'System.Collections.Generic.List[object]'
Get-ExchangeServer | Where-Object { $_.IsClientAccessServer } | ForEach-Object {
    $server = $_
    $versionInformation = Get-ExchangeBuildVersionInformation -AdminDisplayVersion $_.AdminDisplayVersion
    switch ($versionInformation.MajorVersion) {
        "Exchange2016" { if ($versionInformation.BuildVersion -ge "15.1.2308.8") { $exchangeServer.Add($server); break } }
        "Exchange2019" { if ($versionInformation.BuildVersion -ge "15.2.922.7") { $exchangeServer.Add($server); break } }
    }
}
```

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,849 @@
...
+            }
+            Invoke-WebRequest https://$ExchangeServer/ecp/x.js -Method POST -Headers @{ "Host" = "$ExchangeServer" } -WebSession $CookieContainer
         } catch [System.Net.WebException] {
-            if ($_.Exception.Message -notlike "*: (400)*") {
-                $Message = ($_.Exception.Message).ToString().Trim()
-                Write-Output $msgNewLine
-                Write-Error $Message
+            $Message = ($_.Exception.Message).ToString().Trim()
+            $currentForegroundColor = $host.ui.RawUI.ForegroundColor
+            if ($_.Exception.Message -eq "The underlying connection was closed: Could not establish trust relationship for the SSL/TLS secure channel.") {
```
lusassl-msft, line 206: I'm not sure if this works properly for example, on systems running on a different language then en-US. Would be better to check for the status code 9 (see: https://learn.microsoft.com/en-us/dotnet/api/system.net.webexceptionstatus?view=net-7.0). 

image

lusassl-msft (changes requested):

Admin/Test-AMSI.ps1
iserrano76: This is something that comes from the original script but it looks a better solution. I will check it.

lusassl-msft: It's the same as checking for status `9` which is `TrustFailure` (see: https://learn.microsoft.com/en-us/dotnet/api/system.net.webexceptionstatus?view=net-7.0) but it's up to you which syntax you preferer.

iserrano76: yep, it is the same verification, as it is a enumeration but I think it is more clear to read but agree with you, thanks for the advise.

Admin/Test-AMSI.ps1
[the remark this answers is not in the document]
dpaulson45: Resolving conversation.

Admin/Test-AMSI.ps1
dpaulson45: Yep, we can do this in a different PR before we make another change to the script that isn't a major issue.

Admin/Test-AMSI.ps1
lusassl-msft: @iserrano76 do you want to check if `Get-FilteredSettingOverrideInformation` could be used here to simplify the code or is this something you plan to do in a later PR (or something that you not consider to use)?

iserrano76: I did some test and I think Get-FilteredSettingOverrideInformation is a better solution but I was waiting to confirm if it is going to be included in shared and at the same time I need more time to fully understand and check how to include it.
Maybe another PR later is better.

lusassl-msft: I'm fine with adding it via a new PR. Resolving this conversation for now.

Admin/Test-AMSI.ps1
lusassl-msft: Resolving. Might be included with another PR.

Admin/Test-AMSI.ps1
lusassl-msft: Resolving. Might be included with another PR.

Admin/Test-AMSI.ps1
lusassl-msft: Resolving. Might be included with another PR.

Admin/Test-AMSI.ps1
lusassl-msft: Resolving. Might be included with another PR.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,893 @@
...
+                        }
+                    }
+                }
+            } else {
+                Write-Host $bar
+                Write-Host ""
+                $getSOs = $null
+                $getSOs = Get-SettingOverride -ErrorAction SilentlyContinue | Where-Object {
+                    ($_.ComponentName.ToLower() -eq 'Cafe'.ToLower()) -and
+                    ($_.SectionName.ToLower() -eq 'HttpRequestFiltering'.ToLower()) -and
```
lusassl-msft, line 732: It would be better to not use `.ToLower()` for string comparison (see: https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#recommendations-for-string-usage). I've also learned this recently as it could lead to F/P when using specific cultures (e.g.tr-TR). Better would be using `$string.Equals("StringToCompare", [System.StringComparison]::OrdinalIgnoreCase)`

lusassl-msft (changes requested):
Added another recommendation to improve code quality.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,893 @@
...
+                                    }
+                                    Write-Host "It should be version 1.1.18300.4 or newest."
+                                    if (Test-Path $WindowsDefenderPath -PathType Container) {
+                                        $folder = Get-ChildItem  $WindowsDefenderPath | Sort-Object LastWriteTime -Descending | Select-Object -First 1
+                                        $process = Join-Path $folder.FullName "MpCmdRun.exe"
+                                        if (Test-Path $process -PathType Leaf) {
+                                            $DefenderVersion = $null
+                                            $DefenderVersion = [System.Version]::new((& $process -SignatureUpdate | Where-Object { $_.StartsWith('Engine Version:') }).Split(' ')[2])
+                                            if ($DefenderVersion) {
+                                                if (($DefenderVersion.Major -gt 1) -or
```
lusassl-msft, line 383: As `$DefenderVersion` is a `[System.Version]` type object, we could simplify the check by just doing an:

`$DefenderVersion -ge "1.1.18300.4"`

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,893 @@
...
+    if (-not (Confirm-Administrator)) {
+        Write-Host $msgNewLine
+        Write-Warning "This script needs to be executed in elevated mode. Start the Exchange Management Shell as an Administrator and try again."
+        exit
+    }
+
+    $exchangeShell = Confirm-ExchangeShell
+    if (-not($exchangeShell.ShellLoaded)) {
+        Write-Host $msgNewLine
+        Write-Warning "Failed to load Exchange Shell Module..."
```
lusassl-msft, line 530: Nit: Rename to `Exchange Management Shell` instead `Exchange Shell Module`

lusassl-msft (changes requested):
Some more findings that could be improved.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,891 @@
...
+                                        if ((Get-MpComputerStatus).RealTimeProtectionEnabled) {
+                                            Write-Host "Windows Defender has Real Time Protection Enabled" -ForegroundColor Green
+                                        } else {
+                                            Write-Warning "Windows Defender has Real Time Protection Disabled"
+                                        }
+                                    }
+                                    Write-Host "It should be version 1.1.18300.4 or newest."
+                                    if (Test-Path $WindowsDefenderPath -PathType Container) {
+                                        $folder = Get-ChildItem  $WindowsDefenderPath | Sort-Object LastWriteTime -Descending | Select-Object -First 1
+                                        $process = Join-Path $folder.FullName "MpCmdRun.exe"
```
dpaulson45, line 379: This doesn't make sense here. Why are we listing out all the child items of `$WindowsDefenderPath` which is a directory, then sorting the objects by last write time and just assuming that is a folder? This doesn't look right.

dpaulson45 (changes requested):

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,891 @@
...
+            return $false
+        }
+    }
+
+    function CheckServerAMSI {
+        param(
+            [Parameter(Mandatory = $true)]
+            [string]$ExchangeServer,
+            [Parameter(Mandatory = $false)]
+            [switch]$IsServer
```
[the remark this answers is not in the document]
iserrano76: You can run the AMSI test against a server or against a non-server name, for example an FQDN in a load balancer. If it is not a server we only check the response, if it is a server we will do additional test.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,891 @@
...
                 $host.ui.RawUI.ForegroundColor = "Green"
-                Write-Output "We sent an test request to the ECP Virtual Directory of the server requested"
-                $host.ui.RawUI.ForegroundColor = "Red"
-                Write-Output "The remote server returned an error: (400) Bad Request"
+                Write-Host "We sent an test request to the ECP Virtual Directory of the server requested"
+                $host.ui.RawUI.ForegroundColor = "Yellow"
+                Write-Host "The remote server returned an error: (400) Bad Request"
                 $host.ui.RawUI.ForegroundColor = "Green"
-                Write-Output "---------------------------------------------------------------------------------------------------------------"
                 $host.ui.RawUI.ForegroundColor = "Yellow"
```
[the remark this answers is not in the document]
iserrano76: I think that is something that comes from the original script and I remove an old line but looks I do not remove color change I will remove it.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,891 @@
...
+                        }
+                    } else {
+                        Write-Host "Cannot get Exchange installation path on $server" -ForegroundColor Red
+                    }
+
+                    if (Test-Path $HttpRequestFilteringLogFolder -PathType Container) {
+                        $file = $null
+                        $timeout = (Get-Date).AddMinutes(1)
+                        $detected = $false
+                        $marginTime = New-TimeSpan -Seconds 5
```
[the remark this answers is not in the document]
lusassl-msft: It might be easier to use a timestamp/ correlation id to locate the entry in the log. Something like this: 
image

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,891 @@
...
+                                        if (($testTime.ToUniversalTime().Subtract($DateTime) -lt $marginTime) -and
+                                            ($testTime.ToUniversalTime().Subtract($DateTime) -gt - $marginTime)) {
+                                            if (($line.UrlHost.Equals($server, [System.StringComparison]::OrdinalIgnoreCase)) -and
+                                                ($line.UrlStem.Equals('/ecp/x.js', [System.StringComparison]::OrdinalIgnoreCase)) -and
+                                                ($line.ScanResult.Equals('Detected', [System.StringComparison]::OrdinalIgnoreCase))) {
+                                                Write-Host ""
+                                                Write-Host "We found a detection in HttpRequestFiltering logs: " -ForegroundColor Green
+                                                Write-Host "$line"
+                                                $detected = $true
+                                            }
```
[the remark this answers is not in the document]
iserrano76: In while condition, we stop is it is detected or if we reach the timeout:
while ((-not $detected) -and ((Get-Date) -lt $timeout))
is it not rigth?

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,891 @@
...
+        if ($getSOs) {
+            $getSOs | Out-Host
+            foreach ($so in $getSOs) {
+                if ($so.Status -eq 'Accepted') {
+                    Write-Warning "AMSI is Disabled by $($so.Name) SettingOverride for $ExchangeServer"
+                } else {
+                    Write-Host "AMSI is Disabled by $($so.Name) SettingOverride for $ExchangeServer but it is not Accepted." -ForegroundColor Red
+                }
+            }
+        }
```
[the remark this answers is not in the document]
iserrano76: I am not sure if you are speaking about Get-FilteredSettingOverrideInformation.ps1
If it is the one, I was speaking about it with @lusassl-msft, and we agree to include it in a new PR.
Rigth now it is not in shared it is in:
/Diagnostics/HealthChecker/Analyzer
I think this is a better solution, but I need more time to understand it correctly and I guess could be better if we move it that funcion to shared.

Admin/Test-AMSI.ps1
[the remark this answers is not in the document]
iserrano76: I will change the parameter name to clarify the behavior.
first parameter Server 
second parameter IsExchangeServer when we do additional test

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,891 @@
...
+                        if ($server.Equals($env:COMPUTERNAME, [System.StringComparison]::OrdinalIgnoreCase)) {
+                            $HttpRequestFilteringLogFolder = Join-Path $ExchangePath "Logging\HttpRequestFiltering\"
+                        } else {
+                            $HttpRequestFilteringLogFolder = Join-Path "\\$server\$($ExchangePath.Replace(':','$'))" "Logging\HttpRequestFiltering\"
+                        }
+                    } else {
+                        Write-Host "Cannot get Exchange installation path on $server" -ForegroundColor Red
+                    }
+
+                    if (Test-Path $HttpRequestFilteringLogFolder -PathType Container) {
```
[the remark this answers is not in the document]
iserrano76: I will move inside ` if ($ExchangePath) {`

Admin/Test-AMSI.ps1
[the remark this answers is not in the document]
iserrano76: If i use an id in the request it is easier to identify and do not need to include margin time.
That simplify the code a lot.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,891 @@
...
+                        $marginTime = New-TimeSpan -Seconds 5
+                        do {
+                            Start-Sleep -Seconds 2
+                            $file = $null
+                            $file = Get-ChildItem $HttpRequestFilteringLogFolder -Filter "HttpRequestFiltering_*.log" | Sort-Object LastWriteTime -Descending | Select-Object -First 1 -Property *
+                            if ($file) {
+                                $csv = Import-Csv $file.FullName
+                                foreach ($line in $csv) {
+                                    $DateTime = $null
+                                    try {
```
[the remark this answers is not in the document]
iserrano76: with the previous logic we do not need to check the time as it is in the request but I can simplify the code as well with select-String
`$found = $file | Get-Content | Select-String "/ecp/CSS-Test-$StringDate.js"`
Thanks for the advise.

Admin/Test-AMSI.ps1
```diff
@@ -1,144 +1,891 @@
...
+        )
+
+        Write-Host ""
+        Write-Host "AMSI Providers detection:" -ForegroundColor Green
+
+        $AMSIProvidersSB = {
+            $AMSIProviders = Get-ChildItem 'HKLM:\SOFTWARE\Microsoft\AMSI\Providers' -Recurse -ErrorAction SilentlyContinue
+            if ($AMSIProviders) {
+                Write-Host "Providers:"
+                $AMSIProviders.Name | Out-Host
```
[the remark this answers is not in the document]
iserrano76: I think that was something that started in the previous script but I can simplified.

Admin/Test-AMSI.ps1
iserrano76: uhm, I think you are right.
I found a better way to do this with the information in the registry key

lusassl-msft (approved):

dpaulson45 (approved):

PR #1779, opened 2023-07-27
RBA Script - Add Delegate Rule Check

Shanefe:
**Issue:**
Admin sometime configure User Delegation instead of Resource Delegates and this will break RBA functionality.
Check of User Delegate Rules and error out. 

**Reason:**
Recent RBA escalations has this a Root Cause

**Fix:**
add check for "Delegate Rule" and Error


dpaulson45:
@Shanefe when making changes, try to use Visual Studio code for the entire repository to avoid issues likes saving file as a non-BOM file.

Calendar/Get-RBASummary.ps1
```diff
@@ -55,6 +55,22 @@ function ValidateMailbox {
     Write-Host ;
 }
 
+# Validate that there are not delegate rules that will block RBA functionality
+function ValidateInboxRules {
+    Write-Host "Checking for Delegate Rules that will block RBA functionality..."
+    Write-Host -NoNewline "Running : "; Write-Host -ForegroundColor Cyan "Get-InboxRule -mailbox $Identity -IncludeHidden"
+    $rules = Get-InboxRule -mailbox $Identity -IncludeHidden
+    # Note as far as I can tell "Delegate Rule <GUID>" is not localized.
+    if ($rules.Name -like "Delegate Rule*") {
```
lusassl-msft, line 64: If it possible that there is more than one rule returned? If yes, I'd recommend doing a `foreach` here.

Shanefe: There will only be one Delegate Rule, and the -Like will find it in the collection of rules returned.

Shanefe: @lusassl-msft let me know if you require the ForEach for readability, etc., otherwise I would say that the code is good to go. 

I have tried it both ways and it will work either way, this way seems cleaner.

lusassl-msft: If there could only be one delegate rule, the current approach is fine for me.

lusassl-msft (approved):

bill-long (approved):

PR #1783, opened 2023-08-01
Add OICE ExchAVExclusion

### LINKED ISSUE #1714: [Issue] - Set-ExchAVExclusions  - OICE exclusion
About the script Set-ExchAVExclusions I noticed a strange behaviour about OICE exclusions.
It is managed this way:
 
#'$env:SystemRoot\Temp\OICE_<GUID>'
        $possibleOICEFolders = Get-ChildItem $env:SystemRoot\temp -Directory -Filter OICE_*.0
        $possibleOICEFolders | ForEach-Object {
            if ( $_.Name.Length -gt 41) {
                $possibleGUID = $_.Name.Substring(5, 36)
                $result = [System.Guid]::Empty
                if ( [System.Guid]::TryParse($possibleGUID, [System.Management.Automation.PSReference]$result) ) {
                    $BaseFolders.Add($_.FullName.ToLower())
                }
            }
 
 
I understand that the OICE* folder must already exists to be added in the exclusion list. But, AFAIK, that kind of folder could be created after, during Exchange “life”.
So I think a generic exclusion should be used instead, such as:
%SystemRoot%\Temp\OICE_????????-????-????-????-????????????.0

As stated here :
https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/configure-extension-file-exclusions-microsoft-defender-antivirus?view=o365-worldwide#use-wildcards-in-the-file-name-and-folder-path-or-extension-exclusion-lists

An asterisk “*” in a folder exclusion will stand in place for a single folder

So, cannot be used like OICE_*. 

But:
The “?” replaces a single character in a folder name.

iserrano76:
**Issue:**
Added OICE folder to ExchAVExclusion

**Reason:**
Include OICE exclusion with wildcards
Resolve #1714 

**Fix:**
Short description of the fix

**Validation:**
Test in lab

### DIFF
--- a/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
+++ b/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
@@ -70,8 +70,28 @@ param (
 . $PSScriptRoot\..\..\Shared\Confirm-Administrator.ps1
 . $PSScriptRoot\..\..\Shared\Confirm-ExchangeShell.ps1
 . $PSScriptRoot\..\..\Shared\Get-ExchAVExclusions.ps1
+. $PSScriptRoot\..\..\Shared\ScriptUpdateFunctions\Test-ScriptVersion.ps1
 . $PSScriptRoot\Write-SimpleLogFile.ps1
 . $PSScriptRoot\Start-SleepWithProgress.ps1
+
+$BuildVersion = ""
+
+Write-Host ("Test-ExchAVExclusions.ps1 script version $($BuildVersion)") -ForegroundColor Green
+
+if ($ScriptUpdateOnly) {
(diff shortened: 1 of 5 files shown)


dpaulson45:
Moving forward, @iserrano76 please do small changes within a single PR with a single commit.

This PR calls out that it is fixing an issue, but you did much more than this. Please do not do this again moving forward. When fixing an issue, only fix that issue in the PR. 

If you want to make other changes, please do that in a different PR.

dpaulson45 (approved):

dpaulson45 (changes requested):
Please make parameter changes. Also update the docs as well for the new parameters.

dpaulson45 (approved):

bill-long (approved):

dpaulson45:
Also resolves #1785

PR #1816, opened 2023-08-21
Update Invoke-AnalyzerNicSettings.ps1

technion:
Change format of IP4 address.

**Issue:**
Output currently created by the tool looks like this:
`Address: 192.168.1.10\24 Gateway: 192.168.1.1`

**Reason:**
This subnet prefix formatting is backwards to any existing standard. Microsoft's own documentation here for example uses forward slashes:

https://learn.microsoft.com/en-us/microsoft-365/enterprise/urls-and-ip-address-ranges?view=o365-worldwide#exchange-online

The Wikipedia page describes the format as containing a slash, and not a backslash:

https://en.wikipedia.org/wiki/Subnet

**Fix:**
Change the slash direction

**Validation:**
Not much to validate on a single character change

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerNicSettings.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerNicSettings.ps1
@@ -159,7 +159,7 @@ function Invoke-AnalyzerNicSettings {
         Add-AnalyzedResultInformation -Name "IPv4 Address" @baseParams
 
         foreach ($address in $adapter.IPv4Addresses) {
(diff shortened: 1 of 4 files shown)


lusassl-msft:
Closing/Re-opening as pipeline does not run

dpaulson45:
@technion will approve once you can also change the pester testing files to make them pass as well. We can't make changes to your fork to update this.

technion:
@lusassl-msft @dpaulson45 Thanks for looking at this. I've updated the Pester tests and ensured they run cleanly. I've just sent a new push.

technion:
@microsoft-github-policy-service agree

lusassl-msft (approved):
Thanks for your contribution. Approved.

PR #1818, opened 2023-08-25
HealthChecker HSTS check

### LINKED ISSUE #1748: [Work Item] - HSTS check for HealthChecker
**Describe the work**
We're seeing more customers enable HSTS when this isn't officially supported by Exchange - https://learn.microsoft.com/en-us/iis/get-started/whats-new-in-iis-10-version-1709/iis-10-version-1709-hsts

**Additional Context**
We can check this and advise it be disabled for each Site like so:
**(Get-IISSite -Name "Default Web Site").HSTS
(Get-IISSite -Name "Exchange Back End").HSTS**

lusassl-msft:
**Description:**
Exchange Server 2016 and 2019 do support HSTS now. This PR will add the first version of a HC HSTS check to detect misconfigurations that can lead to issues on Exchange Server.

Update Pester testing for the following:
- [x] Basic Hsts settings both being used (custom and native) 
- [x] Hsts Setting and Custom Header value
- [x] When Hsts is enabled on the `Exchange Back End `
- [x] When the value for `max-age` is at a lower value than expected and configured correctly
- [x] When `redirectHttpToHttps` is set to `$true`

**Validated:**
E19 and E16 lab

Resolve #1748

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
@@ -392,8 +392,11 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
                 if (($webSite.Hsts.NativeHstsSettings.enabled) -or
                     ($webSite.Hsts.HstsViaCustomHeader.enabled)) {
                     $params = $baseParams + @{
-                        Details          = "HSTS Enabled: $($webSite.Name)"
-                        DisplayWriteType = if ($isExchangeBackEnd) { "Red" } else { "Green" }
+                        Name                = "HSTS Enabled"
+                        Details             = "$($webSite.Name)"
+                        TestingName         = "hsts-Enabled-$($webSite.Name)"
+                        DisplayTestingValue = $true
+                        DisplayWriteType    = if ($isExchangeBackEnd) { "Red" } else { "Green" }
                     }
                     Add-AnalyzedResultInformation @params
 
@@ -402,6 +405,8 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
                         $params = $baseParams + @{
                             Details                = "HSTS on 'Exchange Back End' is not supported and can cause issues"
                             DisplayWriteType       = "Red"
+                            TestingName            = "hsts-BackendNotSupported"
+                            DisplayTestingValue    = $true
                             DisplayCustomTabNumber = 2
                         }
                         Add-AnalyzedResultInformation @params
@@ -415,6 +420,8 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
                             Details                = ("HSTS configured via customHeader and native IIS control - please remove one configuration" +
                                 "`r`n`t`tHSTS native IIS control has a higher weight than the customHeader and will be used")
                             DisplayWriteType       = "Yellow"
+                            TestingName            = "hsts-conflict"
+                            DisplayTestingValue    = $true
                             DisplayCustomTabNumber = 2
                         }
                         Add-AnalyzedResultInformation @params
@@ -436,18 +443,24 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
                     $params = $baseParams + @{
                         Details                = "max-age: $maxAgeValue"
                         DisplayWriteType       = $hstsMaxAgeWriteType
+                        TestingName            = "hsts-max-age-$($webSite.Name)"
+                        DisplayTestingValue    = $maxAgeValue
                         DisplayCustomTabNumber = 2
                     }
                     Add-AnalyzedResultInformation @params
 
                     $params = $baseParams + @{
                         Details                = "includeSubDomains: $($hstsConfiguration.includeSubDomains)"
+                        TestingName            = "hsts-includeSubDomains-$($webSite.Name)"
+                        DisplayTestingValue    = $hstsConfiguration.includeSubDomains
                         DisplayCustomTabNumber = 2
                     }
                     Add-AnalyzedResultInformation @params
 
                     $params = $baseParams + @{
                         Details                = "preload: $($hstsConfiguration.preload)"
+                        TestingName            = "hsts-preload-$($webSite.Name)"
+                        DisplayTestingValue    = $hstsConfiguration.preload
                         DisplayCustomTabNumber = 2
                     }
                     Add-AnalyzedResultInformation @params
@@ -455,6 +468,8 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
                     $redirectHttpToHttpsConfigured = $hstsConfiguration.redirectHttpToHttps
                     $params = $baseParams + @{
                         Details                = "redirectHttpToHttps: $redirectHttpToHttpsConfigured"
+                        TestingName            = "hsts-redirectHttpToHttps-$($webSite.Name)"
+                        DisplayTestingValue    = $redirectHttpToHttpsConfigured
                         DisplayCustomTabNumber = 2
                     }
                     if ($redirectHttpToHttpsConfigured) {
@@ -469,6 +484,8 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
                 $params = $baseParams + @{
                     Details                = "`r`n`t`tMore Information about HSTS: https://aka.ms/HC-HSTS"
                     DisplayWriteType       = "Yellow"
+                    TestingName            = 'hsts-MoreInfo'
+                    DisplayTestingValue    = $true
                     DisplayCustomTabNumber = 2
                 }
                 Add-AnalyzedResultInformation @params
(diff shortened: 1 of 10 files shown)


Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
```diff
@@ -42,20 +42,73 @@ function Get-IISWebSite {
...
+            redirectHttpToHttps = $false
+        }
+        $customHeaderHsts = $null
 
         if ($webConfigExists) {
             $webConfigContent = (Get-Content $configurationFilePath -Raw).Trim()
 
             try {
-                [xml]$webConfigContent | Out-Null
+                $webConfigContentXml = [xml]$webConfigContent #| Out-Null
```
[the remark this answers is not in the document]
lusassl-msft: Yes, it shouldn't be there. Forgot to remove it before pushing the commit. Will be removed with the next commit.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
```diff
@@ -42,20 +42,73 @@ function Get-IISWebSite {
...
                 $validWebConfig = $true
             } catch {
                 # Inside of Invoke-Command, can't use Invoke-CatchActions
                 Write-Verbose "Failed to convert IIS web config '$configurationFilePath' to xml. Exception: $($_.Exception)"
             }
         }
 
+        if ($validWebConfig) {
+            # Extract HSTS settings from web.config
```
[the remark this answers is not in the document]
lusassl-msft: Done. Added some more context to this. Let me know if this is sufficent.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
```diff
@@ -372,6 +373,106 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
...
+                    if (($webSite.Hsts.NativeHstsSettings.enabled) -and
+                    ($webSite.Hsts.HstsViaCustomHeader.enabled)) {
+                        $showAdditionalHstsInformation = $true
+                        Write-Verbose "HSTS conflict detected"
+                        $params = $baseParams + @{
+                            Details                = "HSTS configured via customHeader and native IIS control - please remove one configuration"
+                            DisplayWriteType       = "Red"
+                            DisplayCustomTabNumber = 2
+                        }
+                        Add-AnalyzedResultInformation @params
```
dpaulson45, line 419: When we are providing that this is bad. We just point to the doc which doesn't call out that you can't do both, or at least as far as I can tell. Should we update the doc, explain why this is bad in the script or in a CSS-Exchange Doc?

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
```diff
@@ -42,20 +42,87 @@ function Get-IISWebSite {
...
+                The following code searches for a customHeader with name Strict-Transport-Security. If we find the header, we then extract the directive
+                and return them as PSCustomObject. We're looking for the following directive: max-age, includeSubDomains, preload, redirectHttpToHttps
+            #>
+            $customHeaderHsts = ($webConfigContentXml.configuration.'system.webServer'.httpProtocol.customHeaders.add | Where-Object {
+                ($_.name -eq "Strict-Transport-Security")
+                }).value
+            if ($null -ne $customHeaderHsts) {
+                Write-Verbose "Hsts via custom header configuration detected"
+                $customHeaderHstsObj.enabled = $true
+                $maxAgeIndex = $customHeaderHsts.IndexOf("max-age=")
```
dpaulson45, line 90: Need to determine how to handle if value that was entered was not in the correct case sensitive based off the doc. 

1. Does the setting still work and how do we verify? 
2. Depending on the answer to # 1, then we need to determine how to handle this for `IndexOf` as that is case sensitive. If the IIS setting doesn't care, then we can do a simple `ToLower()` on the `string` and be okay. But if it is case sensitive, then we need to have an additional parameter stating that it is misconfigured due to case sensitive.

dpaulson45: image

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
```diff
@@ -42,20 +42,87 @@ function Get-IISWebSite {
...
+                These settings are returned when running the Get-WebSite cmdlet and there is no need to prepare the data as they are ready for use.
+
+                Via customHeader configuration (when running IIS older than the version mentioned before where the native HSTS config is not available
+                or when admins prefer to do it via customHeader as there is no requirement to do it via native HSTS control instead of using customHeader).
+                HSTS via customHeader configuration are stored under httpProtocol.customHeaders element. As we get the content in the previous
+                call, we can simply use these data to extract the customHeader with name Strict-Transport-Security (if exists) and can then prepare
+                the data for further processing.
+                The following code searches for a customHeader with name Strict-Transport-Security. If we find the header, we then extract the directive
+                and return them as PSCustomObject. We're looking for the following directive: max-age, includeSubDomains, preload, redirectHttpToHttps
```
dpaulson45, line 82: Is it possible to configure this and have it work on the `web.config` file for the virtual directory? We can have `customHeaders` section in that file as well. 

Is it possible to have `customHeaders` on the `applicationHost.config` ?

dpaulson45: We might need to update or docs on this, as i am guessing this could be very bad: 
image

image

Trying to make sure that `applicationHost.config` got updated. Found out we created a new configuration file. 

image

image

dpaulson45: Actually, this should be that bad. We should still inherit the settings from the original `applicationHost.config` file as well.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
lusassl-msft: Need to understand how this behaves if both ways are configured (will test that later today). We can then decide if we want to have this covered in the HSTS documentation or if it makes sense to maintain a separate check documentation.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
lusassl-msft: Will be fixed with next commit.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
lusassl-msft: Do you think we should add more context on this to the HSTS documentation on learn.microsoft.com?
This is what we have right now:
image

dpaulson45: Not sure, we do state **not** to do it, but we know how good some people are at reading.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
lusassl-msft: Okay, I've tested this one. If you configure HSTS via native HSTS control + customHeader, the setting configured via native will be used. If you remove the native HSTS configuration, IIS will send the customHeader configuration. 
I'd suggest changing this from DisplayWriteType Red to Yellow as it should not break anything. Do you agree?

lusassl-msft: Addressed with the latest commit.

Diagnostics/HealthChecker/DataCollection/ExchangeInformation/IISInformation/Get-IISWebSite.ps1
lusassl-msft: Yeah, let's see how this goes and if we get escalations regarding this. We could then change it to a red warning box or improve the wording.

dpaulson45:
Didn't go into depth with the customHeaders testing. If we end up having issues with that, we will need to create some more. Since that is for a legacy type of config, didn't want to spend more time than needed on it.

dpaulson45 (approved):

PR #1835, opened 2023-09-29
Include .NET 4.8.1 detection

dpaulson45:
**Reason:**
Noticed that we released [.NET Framework 4.8.1](https://devblogs.microsoft.com/dotnet/announcing-dotnet-framework-481/) last year and wasn't made aware. Adding the code to detect the latest version of .NET properly. 

**Fix:**
Add detection for the latest version of .NET and pester testing around it. 

**Validation:**
Lab tested and Pester tested. 

image

Because it has been out for a year. Want to verify that PG is aware and to make sure we don't support this, even though the documentation does state that we don't support this at this time.


dpaulson45:
Updated Health Checker to show this additional information. 

image

lusassl-msft (approved):

PR #1840, opened 2023-10-03
Collect the virtual directory information in HC

dpaulson45:
**Reason:**
There are some additional checks that we can add to verify that we are configured correctly for the virtual directories. For now, let's just collect them all and store them in a property. 

This will add some time to the data collection process because we aren't using the `-ADPropertiesOnly` parameter. We could add this switch to the collection if we need to. However, we will lose the ability to rely on some properties of the output. 

**Validation:**
Lab tested and Pester tested


lusassl-msft (approved):

dpaulson45:
Need to add in Pester testing for this as well.

PR #1847, opened 2023-10-06
Improve OS detection when running on Server Core

lusassl-msft:
**Description:**
OS detection via `Get-ServerOperatingSystemVersion` function wasn't working well in case that the script was executed on or against an Exchange Server that is running on Windows Server Core.

With this PR we change the way of how `Get-ServerOperatingSystemVersion` queries the OS information for proper OS and installation type detection from `wmi` call to a registry call.

**Validation:**
Lab / Pester

### DIFF
--- a/Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-ServerOperatingSystemVersion.ps1
+++ b/Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-ServerOperatingSystemVersion.ps1
@@ -1,25 +1,43 @@
 ﻿# Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-. $PSScriptRoot\..\..\..\..\Shared\Get-WmiObjectHandler.ps1
+. $PSScriptRoot\..\..\..\..\Shared\Get-RemoteRegistryValue.ps1
 function Get-ServerOperatingSystemVersion {
     [CmdletBinding()]
-    [OutputType("System.String")]
+    [OutputType("System.Object")]
     param(
-        [string]$OsCaption
+        [string]$ComputerName = $env:COMPUTERNAME,
+
+        [ScriptBlock]$CatchActionFunction
     )
     begin {
         Write-Verbose "Calling: $($MyInvocation.MyCommand)"
         $osReturnValue = [string]::Empty
     }
     process {
-        if ([string]::IsNullOrEmpty($OsCaption)) {
-            Write-Verbose "Getting the local machine version build number"
-            $OsCaption = (Get-WmiObjectHandler -Class "Win32_OperatingSystem").Caption
+        Write-Verbose "Getting the version build information for computer: $ComputerName"
+        $baseParams = @{
+            MachineName         = $ComputerName
+            CatchActionFunction = $CatchActionFunction
         }
-        Write-Verbose "OsCaption: '$OsCaption'"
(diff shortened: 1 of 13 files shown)


Diagnostics/HealthChecker/Tests/HealthCheckerTest.CommonMocks.NotPublished.ps1
```diff
@@ -96,6 +96,8 @@ Mock Get-RemoteRegistryValue {
         "MinimumConnectionTimeout" { return 0 }
         "LmCompatibilityLevel" { return $null }
         "UBR" { return 720 }
+        "ProductName" { return Import-Clixml "$Script:MockDataCollectionRoot\OS\RemoteRegistryValueProductName.xml" }
+        "InstallationType" { return Import-Clixml "$Script:MockDataCollectionRoot\OS\RemoteRegistryValueInstallationType.xml" }
```
dpaulson45, line 100: Do we need the files? Or can we just provide a return value here instead?

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-ServerOperatingSystemVersion.ps1
```diff
@@ -1,39 +1,59 @@
...
+            GetValue = "ProductName"
         }
-        Write-Verbose "OsCaption: '$OsCaption'"
 
-        switch -Wildcard ($OsCaption) {
+        # Find out if we're running on Server Core to output on the 'Operating System Information' page
+        $installationTypeParams = $baseParams + @{
+            SubKey   = "SOFTWARE\Microsoft\Windows NT\CurrentVersion"
+            GetValue = "InstallationType"
+        }
```
dpaulson45, line 34: Doesn't matter to much, but for variables like this that aren't going to change, look better in `begin` block so that doesn't look so empty. Not a required change, just for future reference.

Diagnostics/HealthChecker/Tests/HealthCheckerTest.CommonMocks.NotPublished.ps1
lusassl-msft: I think a return value should work. Will change it.

Diagnostics/HealthChecker/DataCollection/ServerInformation/Get-ServerOperatingSystemVersion.ps1
lusassl-msft: Indeed. Will move it to begin

Diagnostics/HealthChecker/Tests/HealthCheckerTest.CommonMocks.NotPublished.ps1
lusassl-msft: Let's keep the files for now as we have different return values based on the Exchange version / OS (for example, Windows Server 2019 Datacenter and Server Core on Windows Server 2019)

dpaulson45: Yeah, that is true. I was thinking about this yesterday. Going to need to create a better pester testing setup for this so we don't need to update and add files into Exchange 2013 testing sections any longer. This would be a good example of this.

PR #1848, opened 2023-10-09
Include W3wp process in the analysis.

iserrano76:
**Issue:**
w3wp process is key in the Exchange system and it is interesting to know if a 3rd party module is included in the w3wp process including if it is not in the recommended exclusion list.

**Reason:**
Get report of the w3wp process modules

**Fix:**
Added w3wp process for analysis

**Validation:**

Test in lab:
image

### DIFF
--- a/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
+++ b/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
@@ -416,9 +416,8 @@ $ModuleAllowList.add("SCCXT.dll")
 Write-SimpleLogFile -string ("Allow List Module Count: " + $ModuleAllowList.count) -Name $LogFile
 
 $UnexpectedModuleFound = 0
-
-
 $showWarning = $false
+
 # Gather each process and work thru their module list to remove any known modules.
 foreach ($process in $ServerProcess) {
 
@@ -440,13 +439,11 @@ foreach ($process in $ServerProcess) {
         }
 
         if ($ProcessModules.count -gt 0) {
-
             if ($UnexpectedModuleFound -eq 0) {
                 "`n####################################################################################################" | Out-File $OutputProcessPath  -Append
                 "$((Get-Date).ToString())" | Out-File $OutputProcessPath -Append
                 "####################################################################################################" | Out-File $OutputProcessPath -Append
             }
-
             Write-Warning ("Possible AV Modules found in process $($process.ProcessName)")
             $UnexpectedModuleFound++
             foreach ($module in $ProcessModules) {
--- a/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
+++ b/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
@@ -383,20 +383,19 @@ $ServerProcess = Get-Process | Sort-Object -Property ProcessName
 $ModuleAllowList = New-Object Collections.Generic.List[string]
 
 # cSpell:disable
-$ModuleAllowList.add("Google.Protobuf.ni.dll")
-$ModuleAllowList.add("Microsoft.RightsManagementServices.Core.ni.dll")
-$ModuleAllowList.add("Newtonsoft.Json.ni.dll")
-$ModuleAllowList.add("Microsoft.Cloud.InstrumentationFramework.Events.ni.dll")
+$ModuleAllowList.add("Google.Protobuf.dll")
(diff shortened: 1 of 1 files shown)


Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -391,6 +394,7 @@ $ModuleAllowList.add("ExDbFailureItemApi.dll")
 $ModuleAllowList.add("Microsoft.Cloud.InstrumentationFramework.Metrics.ni.dll")
 $ModuleAllowList.add("IfxMetrics.dll")
 $ModuleAllowList.add("ManagedBlingSigned.dll")
+$ModuleAllowList.add("ManagedBlingSigned.ni.dll")
```
dpaulson45, line 397: This list should be adjusted to have the default `name.dll` files, then at the end, we also need to add in `name.ni.dll` to avoid this type of adding occurring right now we have a mixed results which we don't want.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -438,8 +443,13 @@ foreach ($process in $ServerProcess) {
             Write-Warning ("Possible AV Modules found in process $($process.ProcessName)")
             $UnexpectedModuleFound++
             foreach ($module in $ProcessModules) {
-                $OutString = ("[FAIL] - PROCESS: $($process.ProcessName) MODULE: $($module.ModuleName) COMPANY: $($module.Company)")
-                Write-SimpleLogFile -string $OutString -Name $LogFile
+                if ( $process.MainModule.ModuleName -eq "W3wp.exe" -and $showWarning -eq $false) {
+                    Write-Warning "W3wp.exe is not present in the recommended Exclusion list but we found 3rd Party modules on it and could affect Exchange performance."
```
dpaulson45, line 447: Might need to change the wording here as it could be Exchange Performance and functionality.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
dpaulson45: Or, we could have this added in as the test.

```
        # Clear out modules from the allow list
        foreach ($module in $ModuleAllowList) {
            $ProcessModules = $ProcessModules | Where-Object { $_.ModuleName -ne $module -and $_.ModuleName -ne $($module.Replace(".dll", ".ni.dll")) }
        }
```

But, in order to do that we need to have all the `.dll`s to be lower case.

dpaulson45 (changes requested):

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
iserrano76: something like this:
    W3wp.exe is not present in the recommended Exclusion list but we found 3rd Party modules on it and could affect Exchange performance or functionality.
?

dpaulson45: Yeah, I think that is better.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -436,7 +439,7 @@ foreach ($process in $ServerProcess) {
 
         # Clear out modules from the allow list
         foreach ($module in $ModuleAllowList) {
-            $ProcessModules = $ProcessModules | Where-Object { $_.ModuleName -ne $module }
+            $ProcessModules = $ProcessModules | Where-Object { $_.ModuleName -ne $module -and $_.ModuleName -ne $($module.Replace(".dll", ".ni.dll")) }
```
dpaulson45, line 442: you currently have these that will not convert.

```
$ModuleAllowList.add("SCCOLE2.DLL")
$ModuleAllowList.add("SCCSD.DLL")
$ModuleAllowList.add("SCCXT.DLL")
```

dpaulson45 (changes requested):

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
iserrano76: I think it is not a big issue, because those DLLs are not .net libraries so it is not need it to check ni files, but I will change all DLL to dll and we will have all files in the same format.

dpaulson45:
@iserrano76 the build pipeline is now failing caused by too many empty lines.

dpaulson45 (approved):

PR #1855, opened 2023-10-16
Improve IIS module handling on legacy OS

### LINKED ISSUE #1854: [Issue] - Health Checker TokenCacheModule check doesn't work as expected on WS < 2016
**Describe the issue**
The `TokenCacheModule` check doesn't work properly when executed against OS that are older than Windows Server 2016 (for example, Windows Server 2012 R2). This causes false-positives and leads to questions from customers who are using the script to validate that the TokenCacheModule was successfully added back post October 2023 SU installation.

**Expected behavior**
The check should return a valid result regardless if it was executed against a legacy OS or not.

**Additional context**
The root cause is that the TokenCacheModule IIS state will never returned on legacy OS. This is caused by the module skip list that we use within the `Get-IISModules` function to skip system modules that were not signed by Microsoft on OS older Windows Server 2016. We need to adjust the `Get-IISModules` function to return the IIS load state for each module and on each OS regardless if signed or not or on the skip list. 

```[tasklist]
### Tasks
- [x] Improve the Get-IISModules function to return all modules that are loaded
- [x] Adjust the "IIS module anomalies detected" output for any module that was skipped due to SkipLegacyOSModulesCheck $true
- [x] Update and add a unit test to make sure that the function returns all modules as we expect to be
```

lusassl-msft:
**Issue:**
On Windows Server versions < 2016, the default modules that come with IIS are not signed by Microsoft. We therefore skip the signature status check for those modules and don't return them when doing a `Get-IISModules` call. However, this leads to false-positives when performing additional checks (for example, the TokenCacheModule vulnerability check) as we never report that these modules are loaded by IIS.

**Reason:**
Resolve #1854 

**Fix:**
We now return all modules that are loaded by IIS via `Get-IISModules` call. For modules that are on the skip list, we return the default `ModuleSigned` status which is `$null` (otherwise `$true` or `$false`). We also return a `SignatureStatus` of `-1`.

**Validation:**
Lab

### DIFF
--- a/Shared/IISFunctions/Get-IISModules.ps1
+++ b/Shared/IISFunctions/Get-IISModules.ps1
@@ -24,43 +24,6 @@ function Get-IISModules {
         Write-Verbose "Calling: $($MyInvocation.MyCommand)"
         $modulesToCheckList = New-Object 'System.Collections.Generic.List[object]'
 
-        # Add all modules here which should be skipped on legacy OS (pre-Windows Server 2016)
-        $modulesToSkip = @(
-            "$env:windir\system32\inetSrv\cachUri.dll",
-            "$env:windir\system32\inetSrv\cachFile.dll",
-            "$env:windir\system32\inetSrv\cachtokn.dll",
-            "$env:windir\system32\inetSrv\cachHttp.dll",
-            "$env:windir\system32\inetSrv\compStat.dll",
-            "$env:windir\system32\inetSrv\defDoc.dll",
-            "$env:windir\system32\inetSrv\dirList.dll",
-            "$env:windir\system32\inetSrv\protsUp.dll",
-            "$env:windir\system32\inetSrv\redirect.dll",
-            "$env:windir\system32\inetSrv\static.dll",
-            "$env:windir\system32\inetSrv\authAnon.dll",
-            "$env:windir\system32\inetSrv\cusTerr.dll",
-            "$env:windir\system32\inetSrv\logHttp.dll",
-            "$env:windir\system32\inetSrv\iisEtw.dll",
-            "$env:windir\system32\inetSrv\iisFreb.dll",
-            "$env:windir\system32\inetSrv\iisReQs.dll",
-            "$env:windir\system32\inetSrv\isApi.dll",
-            "$env:windir\system32\inetSrv\compDyn.dll",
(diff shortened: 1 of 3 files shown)


Shared/IISFunctions/Get-IISModules.ps1
```diff
@@ -147,11 +159,12 @@ function Get-IISModules {
                                         if ($null -ne $signature.SignerCertificate.Subject) {
                                             Write-Verbose "Signer information found. Subject: $($signature.SignerCertificate.Subject)"
                                             $signatureDetails.Signer = $signature.SignerCertificate.Subject.ToString()
-                                            $signatureDetails.IsMicrosoftSigned = $signature.SignerCertificate.Subject -cmatch "O=Microsoft Corporation, L=Redmond, S=Washington"
+                                            $signatureDetails.IsMicrosoftSigned = $signature.SignerCertificate.Subject -cmatch $signerSubject
                                         }
                                     }
                                 } else {
                                     Write-Verbose "No signature information found for module $($m.name)"
+                                    $isModuleSigned = $false
```
dpaulson45, line 167: Not required to be here, but fine to have.

dpaulson45 (approved):

Shared/IISFunctions/Get-IISModules.ps1
```diff
@@ -98,15 +61,59 @@ function Get-IISModules {
...
+                [Parameter(Mandatory = $false)]
+                [bool]$SkipLegacyOSModules = $false,
+
                 [Parameter(Mandatory = $false)]
                 [ScriptBlock]$CatchActionFunction
             )
+            begin {
+                # Add all modules here which should be skipped on legacy OS (pre-Windows Server 2016)
+                $modulesToSkip = @(
+                    "$env:windir\system32\inetSrv\cachUri.dll",
```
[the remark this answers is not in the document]
lusassl-msft: Assuming that all servers are configured the same way, this should work. However, we should address this to make sure it works even if that wouldn’t be the case. I will file an issue to track this.

Shared/IISFunctions/Get-IISModules.ps1
lusassl-msft: Indeed. This is a leftover which I forgot to remove. Will remove it together with the environment variable adjustment

PR #1857, opened 2023-10-18
Address not thrown Get-DynamicDistributionGroup error

dpaulson45:
**Issue:**
Customer reported an error in the script. It was an unhandled error caused by `Get-DynamicDistributionGroup`. For some reason the cmdlet wasn't honoring the `ErrorAction` of `stop` to get thrown into the `catch`. 

**Reason:**
To avoid false positive issues. 

**Fix:**
Did a loop error catch around the cmdlet. 

**Validation:**
Lab tested, couldn't repro the issue. Pending customer confirmation.


lusassl-msft (approved):

dpaulson45:
Feedback was provided that the information about HC running into a problem was addressed, but they still see the error show up to the screen at the time of execute. Not a way around this if the catch doesn't work or we use the filter parameter option. 

For now, just going to merge, if more issues pop up will address at that time.

PR #1871, opened 2023-11-01
Improve Credential Guard Running Check

### LINKED ISSUE #1867: [Issue] - HealthChecker Credential Guard Protection doesn’t work as expected
HealthChecker reported Credential Guard is not active, even though it definitely was. Can we have the detection mechanism changed?
 
As far as I see, the script checks a registry value to detect if Credential Guard is active:
 
```
    $credGuardParams = $baseParams + @{
        SubKey   = "SYSTEM\CurrentControlSet\Control\LSA"
        GetValue = "LsaCfgFlags"
    }
```
 
However, per the official doc, other methods are recommended for detection:

https://learn.microsoft.com/en-us/windows/security/identity-protection/credential-guard/configure

In our case, all three methods listed by the article (PowerShell, msinfo32, event logs) confirmed Credential Guard is active, only HealthChecker did not.
I think running the PS cmdlet from the article should be straightforward enough for HealthChecker as well:
 
```
(Get-CimInstance -ClassName Win32_DeviceGuard -Namespace root\Microsoft\Windows\DeviceGuard).SecurityServicesRunning
The command generates the following output:
·       0: Credential Guard is disabled (not running)
·       1: Credential Guard is enabled (running)
```

dpaulson45:
**Issue:**
Issue described in #1867 

**Reason:**
Health Checker didn't properly determine if Credential Guard was enabled on a system. Adding another way to check to see if it is enabled.

**Fix:**
Add in option to check for the `Get-CimInstance` of the credential guard check described in article: 
https://learn.microsoft.com/en-us/windows/security/identity-protection/credential-guard/configure

Resolved #1867 

**Validation:**
Lab tested

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
@@ -107,12 +107,25 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
     }
     Add-AnalyzedResultInformation @params
 
-    $displayValue = $credentialGuardValue = $osInformation.RegistryValues.CredentialGuard -ne 0
+    $credGuardRunning = $false
+    $credGuardUnknown = $osInformation.CredentialGuardCimInstance -eq "Unknown"
+
+    if (-not ($credGuardUnknown)) {
+        # CredentialGuardCimInstance is an array type and not sure if we can have multiple here, so just going to loop thru and handle it this way.
+        $credGuardRunning = $null -ne ($osInformation.CredentialGuardCimInstance | Where-Object { $_ -ne 0 })
+    }
+
+    $displayValue = $credentialGuardValue = $osInformation.RegistryValues.CredentialGuard -ne 0 -or $credGuardRunning
     $displayWriteType = "Grey"
 
     if ($credentialGuardValue) {
         $displayValue = "{0} `r`n`t`tError: Credential Guard is not supported on an Exchange Server. This can cause a performance hit on the server." -f $credentialGuardValue
         $displayWriteType = "Red"
+    }
+
+    if ($credGuardUnknown -and (-not ($credentialGuardValue))) {
+        $displayValue = "Unknown `r`n`t`tWarning: Unable to determine Credential Guard status. If enabled, this can cause a performance hit on the server."
+        $displayWriteType = "Yellow"
     }
 
     $params = $baseParams + @{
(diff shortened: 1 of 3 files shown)


bill-long (approved):

lusassl-msft (approved):

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
```diff
@@ -107,14 +107,27 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
...
-    $displayValue = $credentialGuardValue = $osInformation.RegistryValues.CredentialGuard -ne 0
+    $credGuardRunning = $false
+    $credGuardUnknown = $osInformation.CredentialGuardCimInstance -eq "Unknown"
+
+    if (-not ($credGuardUnknown)) {
+        # CredentialGuardCimInstance is an array type and not sure if we can have multiple here, so just going to loop thru and handle it this way.
+        $credGuardRunning = $null -ne ($osInformation.CredentialGuardCimInstance | Where-Object { $_ -ne 0 })
+    }
+
+    $displayValue = $credentialGuardValue = $osInformation.RegistryValues.CredentialGuard -ne 0 -or $credGuardRunning
```
lusassl-msft, line 118: We keep the `$osInformation.RegistryValues.CredentialGuard -ne 0` for backward compatibility, right?

lusassl-msft: Hm, looks like it can also be configured via GPO. We might need to validate this registry path as well: `SOFTWARE\Policies\Microsoft\Windows\DeviceGuard`

```
ValueName: lsacfgflags
item: decimal: 0 => Disabled
item: decimal: 1 => Enabled with UEFI lock
item: decimal: 2 => Enabled without lock
item: decimal: 3 => Not Configured
```

https://gpsearch.azurewebsites.net/#10939

dpaulson45: > We keep the `$osInformation.RegistryValues.CredentialGuard -ne 0` for backward compatibility, right?
> 
> **_Edit:_** Looks like this is still a supported way to configure Credential Guard (as per: https://learn.microsoft.com/en-us/windows/security/identity-protection/credential-guard/configure?tabs=intune#configure-credential-guard-with-intune)

yes, I wanted to keep it there just in case on older systems doesn't have the option of `Get-CimInstance` for credential guard.

PR #1876, opened 2023-11-10
Check sharing status

Shanefe:
**Issue:**
Script to validate the key features of the sharing relationship between Calendar Owners and Receiver. 

**Reason:**
Help Troubleshoot issues with Modern Calendar Sharing

**Fix:**
Add initial PS file.

**Validation:**
Tested with and without PII access
Torus and Tenant Admin


Shanefe (commented):
I need to remove the change in CalLogSummary...

bill-long (changes requested):
The build is passing, but you'll notice this at the bottom:

```powershell
WARNING: The following scripts are undocumented, unreferenced, and do not declare NotPublished in the filename:
WARNING: D:\a\1\s\Calendar\Check-SharingStatus.ps1
```

The script must have a corresponding file in docs to be considered documented, otherwise it will not be included in the release. After writing the doc, you will also need to add it to mkdocs.yml at the root.

Shanefe:
Thanks, Fixed.
I will have to look at how I missed it.
Shane

Shanefe (commented):
Done

Calendar/Check-SharingStatus.ps1
```diff
@@ -0,0 +1,326 @@
...
+    $csvString = $csvString.Replace("Mailbox: ", "")
+    $csvString = $csvString.Replace("Entry MailboxOwner:", "")
+    $csvString = $csvString.Replace("Recipient:", "")
+    $csvString = $csvString.Replace("RecipientType:", "")
+    $csvString = $csvString.Replace("Handler=", "")
+    $csvString = $csvString.Replace("ms-exchange-", "")
+    $csvString = $csvString.Replace("DetailLevel=", "")
+
+    # Convert the CSV string to an object
+    $csvObject = $csvString | ConvertFrom-Csv
```
dpaulson45, line 98: Instead of doing a csv string array, can't we just create a new PSCustomObject and add that to a list?

Calendar/Check-SharingStatus.ps1
```diff
@@ -0,0 +1,326 @@
...
+    Write-Output "------------------------"
+    Write-Output "Looking for Modern Calendar Sharing Accept data for [$Identity]."
+
+    # Call the Export-MailboxDiagnosticLogs cmdlet and store the output in a variable
+    try {
+        # Call the Export-MailboxDiagnosticLogs cmdlet and store the output in a variable
+        # -ErrorAction is not supported on Export-MailboxDiagnosticLogs
+        # $logOutput = Export-MailboxDiagnosticLogs $Identity -ComponentName AcceptCalendarSharingInvite -ErrorAction SilentlyContinue
+
+        $logOutput = Export-MailboxDiagnosticLogs $Identity -ComponentName AcceptCalendarSharingInvite
```
dpaulson45, line 141: Can you email a xml output from this cmdlet and from line 65? We might be able to create something that can be used over and over again a little easier if i am able to see the data.

dpaulson45 (changes requested):

Shanefe:
Here are the XML's.
At this point, the Exported logs are fairly flat files.

dpaulson45:
@Shanefe you are going to need to email me internally, looks like it was stripped.

Calendar/Check-SharingStatus.ps1
dpaulson45: After looking at the logs that was provided, I still think the object type would be better way to go about reading the data. However, I will keep this up to you. If you want to keep it as is, that is fine.

dpaulson45 (approved):

bill-long (approved):

PR #1880, opened 2023-11-15
Refactor VSSTester

bill-long:
This is a significant refactoring of VSSTester.

* All Read-Host calls are removed and replaced with parameters.
* All script-scoped variables are removed.
* Output is more terse.
* Event logs are saved in evtx format rather than Export-CliXML.
* Output folder is created if it does not exist instead of erroring out.
* Each run of the script creates a subfolder with a timestamp name and puts all the data for that run in that folder, instead of appending and overwriting files.
* Script now has the option to start circular tracing, watch for the Microsoft Exchange Writer to disappear, and then stop the traces automatically.
* Must now be run with parameters. Cannot be run with no parameters as before.


Databases/VSSTester/DiskShadow/Invoke-CreateDiskShadowFile.ps1
```diff
@@ -244,45 +189,19 @@ function Invoke-CreateDiskShadowFile {
...
+        Write-Host "  Since different volumes are used for this database's EDB and logs, we need two drive"
+        Write-Host "  letters to expose the backup snapshot."
+    }
 
-        $nl
+    $dbSnapVol = $DatabaseDriveLetter
+    if ($comp) {
+        $logSnapVol = $LogDriveLetter
+    } else {
+        $logSnapVol = $dbSnapVol
```
dpaulson45, line 204: Makes more sense to add this to the previous `if` statement.

Databases/VSSTester/Logging/Get-VSSWriter.ps1
```diff
@@ -0,0 +1,24 @@
...
+    [CmdletBinding()]
+    param ()
+
+    $writersText = vssadmin list writers
+    if ($LASTEXITCODE) {
+        Write-Warning $writersText
+        throw "Unable to list vss writers"
+    }
+
+    $lineNumber = 3
```
dpaulson45, line 14: This shouldn't need to be here if we defined it in the `for` loop.

Databases/VSSTester/VSSTester.ps1
```diff
@@ -1,33 +1,81 @@
...
+    [Parameter(Mandatory = $true, ParameterSetName = "TraceOnly")]
+    [Parameter(Mandatory = $true, ParameterSetName = "DiskShadow")]
+    [Parameter(Mandatory = $true, ParameterSetName = "WaitForWriterFailure")]
+    [string]
+    $DatabaseName,
+
+    # Drive letter on which to expose the database snapshot.
+    [Parameter(Mandatory = $true, ParameterSetName = "DiskShadow")]
+    [string]
+    $DatabaseDriveLetter,
```
dpaulson45, line 54: Haven't done any testing, but is it a concern if `M:` or `M:\` is provided?

Databases/VSSTester/VSSTester.ps1
```diff
@@ -40,167 +88,125 @@ param(
...
-            while ($null -notmatch $continue)
-            exit
-        } catch { }
+        Write-Host "$(Get-Date) Stopping traces..."
+        Invoke-DisableExTRATracing -ServerName $serverName -Database $dbForBackup -OutputPath $LoggingPath
+        Invoke-DisableDiagnosticsLogging
+        Invoke-DisableVSSTracing
+        Write-Host "$(Get-Date) Tracing stopped."
+
+        if ($collectEventLogs) {
```
dpaulson45, line 202: Maybe we should add in an `else` statement with a `Write-Host` stating that we aren't collecting data because `$collectEventLogs` isn't set.

dpaulson45 (changes requested):

Databases/VSSTester/DiskShadow/Invoke-CreateDiskShadowFile.ps1
```diff
@@ -3,22 +3,42 @@
...
     #	creates the DiskShadow.dsh file that will be written to below
     #	-------------------------------------------------------------
-    $nl
-    Get-Date
-    Write-Host "Creating DiskShadow config file..." -ForegroundColor Green $nl
-    Write-Host "--------------------------------------------------------------------------------------------------------------"
-    $nl
-    New-Item -Path $path\DiskShadow.dsh -type file -Force | Out-Null
+    Write-Host "$(Get-Date) Creating DiskShadow config file..."
+    New-Item -Path $OutputPath\DiskShadow.dsh -type file -Force | Out-Null
```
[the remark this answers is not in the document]
bill-long: Not sure. It seems there are multiple scenarios where we don't enforce parameter casing for some reason.

Databases/VSSTester/DiskShadow/Invoke-CreateDiskShadowFile.ps1
```diff
@@ -3,22 +3,42 @@
 
 function Invoke-CreateDiskShadowFile {
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSAvoidUsingWMICmdlet', '', Justification = 'Required to get drives on old systems')]
```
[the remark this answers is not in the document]
bill-long: No idea. I haven't tested other approaches. I suggest a separate PR for that if that is a goal.

Databases/VSSTester/DiskShadow/Invoke-CreateDiskShadowFile.ps1
bill-long: Done.

A huge portion of this file could use more improvement, but this is essentially the same logic from before this PR. We check `[string]::Compare()` for some reason instead of just doing -eq, and then we check whether they match once before writing the `add volume` lines, then we check again before writing the text about the drive letters, and then we check a third time before writing the `expose` lines.

Moving this into the previous if gets us down to 2 comparisons. Further improvement can be done in a future PR.

Databases/VSSTester/Logging/Get-VSSWriter.ps1
bill-long: Good catch. That was left over from before I turned it into a `for`.

Databases/VSSTester/VSSTester.ps1
bill-long: Not anymore. Added [ValidateLength] attribute.

image

Databases/VSSTester/VSSTester.ps1
bill-long: Done.

dpaulson45 (approved):

PR #1891, opened 2023-11-22
Save RBA Log, RBA Logs summary, Better coverage of known issues

Shanefe:
**Issue:**
Known issue was not well flagged.
Save Logs as we collected it.
Summarize RBA Log with more details.

### DIFF
--- a/Calendar/Get-RBASummary.ps1
+++ b/Calendar/Get-RBASummary.ps1
@@ -58,6 +58,7 @@ function ValidateMailbox {
     if ($null -eq $script:Place) {
         Write-Error "Error: Get-Place returned Null for $Identity."
         Write-Host -ForegroundColor Red "Make sure you are running from the correct forest.  Get-Place does not cross forest boundaries."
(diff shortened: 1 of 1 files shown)


Calendar/Get-RBASummary.ps1
```diff
@@ -522,6 +540,52 @@ function RBALogSummary {
         if ($AcceptLogs.count -eq 0 -and $TentativeLogs.count -eq 0 -and $DeclineLogs.count -eq 0) {
             Write-Host -ForegroundColor Red "`t No meetings were processed in the RBA Log."
         }
+
+        if ($UpdatedLogs.count -ne 0) {
+            $LastUpdated = ($UpdatedLogs[0] -Split ",")[0].Trim()
+            Write-Host "`t $($UpdatedLogs.count) Updates to meetings between $FirstDate and $LastDate"
```
dpaulson45, line 546: It is possible that `$FirstDate` is empty, because it is only set inside of a `if` statement.

dpaulson45 (approved):

PR #1894, opened 2023-11-27
Output changes and parameter for waiting time in Test-ExchAVExclusions

### LINKED ISSUE #1764: [Work Item] Test-ExchAVExclusions improvements
```[tasklist]
### Tasks
- [ ] Five minute hard-coded timeout is too long. Consider reducing the timeout and/or allow changing it via parameter and doing other work while waiting.
- [ ] Output should go to the same location as the script. Use $PSScriptRoot.
```

iserrano76:
**Issue:**
Unify the output file with wrong exclusions in just one and use a parameter for waiting time for AV actions.
Resolved #1764

**Reason:**
Simplify the output and allow users to have a smaller time waiting for the AV actions.

**Validation:**
Test in lab


Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -350,126 +444,44 @@ foreach ($extension in $extensionsList) {
 #Delete Random Folder
 Remove-Item $randomFolder
 
+$OutputPath = Join-Path $env:LOCALAPPDATA BadExclusions.txt
```
dpaulson45, line 447: Based off the issue, this should be using `$PSScriptRoot` instead of `$env:LOCALAPPDATA`

dpaulson45: make sure you also change the documentation as well.

dpaulson45 (changes requested):

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
iserrano76: I found something weird when I was modifying this code. 
The code that write in LOCALAPPDATA is the function Write-SimpleLogFile but We have 2 functions with the same name and the same code one is under Diagnostics\AVtester and another one is on Performance\ExPerfWiz.
I will modify the one on EVTester to change write on LOCALAPPDATA and move to the path received.

Diagnostics/AVTester/Write-SimpleLogFile.ps1
```diff
@@ -51,8 +54,15 @@ function Write-SimpleLogFile {
         [Parameter(Mandatory = $true)]
         [string]$String,
 
+        [parameter(Mandatory)]
+        [ValidateScript({
+                if (Test-Path -Path $_ -PathType Container) { $true }
+                else { throw "Path $_ is not valid" }
+            })]
+        [String]$Path,
+
```
dpaulson45, line 63: This is going to break `Set-ExchAVExclusions.ps1` the way you have this setup. By creating a new mandatory parameter to `Write-SimpleLogFile` that is used in this other script, we need to also modify that script or update it so it accepts a default value.

bill-long: The syntax of `Some-Command -Path C:\must\be\a\folder -FileName MustBeAFile.log` is very unusual. Unless we have a justification for this, I suggest changing to the same syntax as every other command that takes a file, which is to use a single parameter:

```
PS C:\> Start-Transcript -Path C:\Users\bilong\Documents\file.log
Transcript started, output file is C:\Users\bilong\Documents\file.log
PS C:\> Get-Process | Export-Csv -Path C:\Users\bilong\Documents\export.csv
PS C:\> Add-Content -Path C:\Users\bilong\Documents\export.csv -Value "Foo"
```

iserrano76: > This is going to break `Set-ExchAVExclusions.ps1` the way you have this setup. By creating a new mandatory parameter to `Write-SimpleLogFile` that is used in this other script, we need to also modify that script or update it so it accepts a default value.

I miss that part

PR #1911, opened 2023-12-18
Make the basic build process work on MacOS

lusassl-msft:
**Description:**
The `build.ps1` script wasn't working properly on MacOs. I've made minor changes to it to make it work. However, Pester unit tests are still not working, need to make some more adjustments to them to make them work as well. However, since this isn't a high priority item, I will look into this in the near future.

I’ve tested this on my own MacOs device and build went fine.


.build/Build.ps1
```diff
@@ -139,7 +139,14 @@ if ($nonUnique.Count -gt 0) {
...
 
-$disclaimer = [IO.File]::ReadAllLines("$PSScriptRoot\..\LICENSE")
+if ($IsMacOS -or
+    $IsLinux) {
+    Write-Host "We're running the build process on MacOS / Linux"
+    $disclaimer = [IO.File]::ReadAllLines("$((Get-Location).Path)/../LICENSE")
+} else {
+    Write-Host "We're running the build process on Windows"
+    $disclaimer = [IO.File]::ReadAllLines("$PSScriptRoot\..\LICENSE")
+}
```
bill-long, line 149:
```suggestion
$disclaimer = [IO.File]::ReadAllLines([IO.Path]::Combine($PSScriptRoot, "..", "LICENSE"))
```

Suggest replacing all of this with an environment-agnostic path combine. If we really want to call out what OS we're running on, I would like to separately call out all three, not combine MacOS and Linux together, since they are not the same.

bill-long (changes requested):

lusassl-msft:
Will make a few more adjustments here. So converting it to draft for now.

bill-long (approved):

PR #1913, opened 2023-12-20
Move Write-SimpleLogFile to share and remove dup.

iserrano76:
**Issue:**
We have the same function, Write-SimpleLogFile, duplicate in Diagnostics/AVTester and in Performance/ExPerfWiz

**Reason:**
To simplify the code 

**Fix:**
Move the function to Shared and review the code to avoid issues.

**Validation:**
Test in Lab.


Shared/Write-SimpleLogFile.ps1
```diff
@@ -47,14 +44,14 @@ Write-SimpleLogFile -String "Start ProcessB" -FileName MyLogFile.log -OutHost -P
 Writes "[Date] - Start ProcessB" to C:\temp\MyLogFile and to the Host
 
 #>
-function Write-SimpleLogFile {
+function global:Write-SimpleLogFile {
```
dpaulson45, line 47: @bill-long please provide your take on making a shared function a `global` function in order to make `ExPerfwiz` work. I personally don't think this should be the correct course of action.

bill-long: I agree. It looks like the only thing that uses global functions is ExPerfWiz. I'd rather not pollute the rest of the code base with unnecessary global-scoped functions just to make ExPerfWiz work.

iserrano76: I was trying to put the same code in just one function and avoid duplication that can cause errors and simplify the code.
Right now we have 2 functions with almost the same content in Diagnostics/AVTester and in Performance/ExPerfWiz.
In fact, the current Write-SimpleLogFile in Performance/ExPerfWiz is global, so it is something that it is happening at this point.
Anyway, if you prefer, I can leave it as it is with the duplication and just cancel the PR, or let me know if you have any other suggestion :)

bill-long: My only other suggestion would be to figure out how to make it work in ExPerfWiz without being global. I haven't looked at the ExPerfWiz code, but since literally no other script in the entire repository uses global functions, there is probably a way to do it.

dpaulson45: I don't think it would worth the investment of time to do this. I don't see the point of `ExPerfWiz` any longer. Since it was re-written into a module, it went downhill and isn't worth our time in my opinion. We have `SimplePerf` which is 100% reliable and isn't complex like `ExPerfWiz` now is. 

I would recommend to not make this change and just close this PR. 

If you are really wanting to remove duplicate code, it would look into how every other script is using `Shared\LoggerFunctions` to handle writing out logging.

iserrano76: I used Write-SimpleLogFile in order to keep using the same functions, but I will take a look on that and try to use it in the script that I was working :)
Thanks for the advice.
I will close this one.

Issue #1922, opened 2023-12-28
[Issue] SourceSideValidations - Get-MailboxDatabase Identity not found


DrSleez:
**Describe the issue**
When running the script two issues appear when trying to execute Get-MailboxDatabase with the Get-Mailbox -PublicFolder function.

**Expected behavior**
The cmdlet Get-MailboxDatabase should return the appropiate Database and in this case the ServerName

**Script Output**
```powershell
Die Argumenttransformation für den Parameter "Identity" kann nicht verarbeitet werden. Der Wert "PFDatabase" vom Typ
"Deserialized.Microsoft.Exchange.Data.Directory.ADObjectId" kann nicht in den Typ
"Microsoft.Exchange.Configuration.Tasks.DatabaseIdParameter" konvertiert werden.
    + CategoryInfo          : InvalidData: (:) [Get-MailboxDatabase], ParameterBindin...mationException
    + FullyQualifiedErrorId : ParameterArgumentTransformationError,Get-MailboxDatabase 
```

**Solution idea**
In our case it was necessary to pass the name parameter of the mailbox to the cmdlet. E.g. 
(Get-MailboxDatabase (Get-Mailbox -PublicFolder (Get-OrganizationConfig).RootPublicFolderMailbox.HierarchyMailboxGuid.ToString()).Database.Name).Server.Name in the function Get-FolderData 
And  $db = Get-MailboxDatabase $_.Database.Name -Status in line 2616 in the check if anyDatabase is down.
Script Version: 23.12.08.2252
Version: Exchange 2019 CU13 Nov23SU
Build Number: 15.02.1258.028

lusassl-msft:
This behavior is caused by a known issue with Serialized Data Signing (https://learn.microsoft.com/en-us/exchange/plan-and-deploy/post-installation-tasks/security-best-practices/exchange-serialization-payload-sign?view=exchserver-2019#november-2023-su).

PR #1924, opened 2024-01-04
Make SSV validate PF content mailboxes

### LINKED ISSUE #829: [Issue] SourceSideValidations - Should validate ContentMailboxName and ContentMailboxGuid
We had a long-running case due to ContentMailboxName and ContentMailboxGuid being blank or invalid on some folders. These should be part of the SSV checks.

bill-long:
Fixes #829 and #1922.


dpaulson45:
FYI this won't auto close both issues.

PublicFolders/SourceSideValidations/SourceSideValidations.ps1
```diff
@@ -117,29 +117,48 @@ try {
...
+        return
+    }
+
+    $folderData.MailboxToServerMap = $mailboxToServerMap
+
+    # Validate that all content mailboxes exist
+    $ipmSubtreeByMailboxGuid = $folderData.IpmSubtree | Group-Object ContentMailboxGuid
+    foreach ($group in $ipmSubtreeByMailboxGuid) {
+        if ([string]::IsNullOrEmpty($group.Name)) {
+            Write-Warning "$($group.Count) folders have no ContentMailboxGuid. "
```
dpaulson45, line 152: Is this a point to attempt to get the mailbox information after this point if the name is in fact null or empty?

Just trying to understand the reason for this `if` statement as the result will come back with a `$null` value and provide you with the same information on line 156.

bill-long: Good point. I added that when I was trying to produce a null value, but I was never able to actually do that. I think we can just let it fall through to the existing logic.

dpaulson45 (approved):
Due to new issues coming up regarding what this fix, we can't wait anymore on the test team. It has been over 3 weeks since this code has been ready for production.

PR #1953, opened 2024-01-23
Noderunner Memory Limit Check

### LINKED ISSUE #927: HealthChecker Check & Warn about Fast Search Memory Config Changes
Based on few articles in internet customers are changing Memory settings for Fast Search which causes Index to either Fail or remain in Unknown state or fails with Out Of memory Exception

Locate the Noderunner.exe.config file. By default, this file is located along the following path: C:\Program Files\Microsoft\Exchange Server\V15\Bin\Search\Ceres\Runtime\1.0
Edit the file, and then locate the following key: ```<nodeRunnerSettings memoryLimitMegabytes="0"/>```

Warning should be generated if the above value is changed from default.

### LINKED ISSUE #1686: [Work Item] NodeRunner Memory check on HealthChecker
**Describe the work**
Using build 23.05.02.2057 we don't appear to have any checks for the noderunner.exe.config located here - C:\Program Files\Microsoft\Exchange Server\V15\Bin\Search\Ceres\Runtime\1.0 to validate the value for "nodeRunnerSettings memoryLimitMegabytes"

**Additional Context**
**// Default configuration**
  
  
  **<nodeRunnerSettings memoryLimitMegabytes="0"/>**

**// Adjusted configuration**
  
  
  **<nodeRunnerSettings memoryLimitMegabytes="1024"/>**

dpaulson45:
**Issue:**
We want to call out if there is a configuration issue for limiting memory usage with the `noderunner.exe.config`. If this is set, we can run into odd performance issues with search. However, in some scenarios, it can be useful to limit it. But depending on the cause for needing to limit, it should be temporary. 

**Fix:**
Include collection of `noderunner.exe.config` and check for the `memoryLimitMegabytes` to see if it set to anything other than 0. 

image

image

Resolved #1686 
Resolved #927 

**Validation:**
Lab testing & Pester


docs/Diagnostics/HealthChecker/NodeRunnerMemoryLimitCheck.md
```diff
@@ -0,0 +1,13 @@
+# Noderunner.exe Memory Limit Check
+
+### Description
+
+This check is looking at the `C:\Program Files\Microsoft\Exchange Server\V15\Bin\Search\Ceres\Runtime\1.0\noderunner.exe.config` to look at the `memoryLimitMegabytes` value. This value should be set to 0 for the best performance. By having it set to 0, we do not limit the `noderunner.exe` processes. However, in some scenarios you might want to recommend to limit the process memory consumption to prevent server impact. If you do this, it is only recommended as a temporary fix.
```
lusassl-msft, line 5:
```suggestion
This check is looking at the `<ExchangeInstallPath>\Bin\Search\Ceres\Runtime\1.0\noderunner.exe.config` to look at the `memoryLimitMegabytes` value. This value should be set to 0 for the best performance. By having it set to 0, we do not limit the `noderunner.exe` processes. However, in some scenarios you might want to recommend to limit the process memory consumption to prevent server impact. If you do this, it is only recommended as a temporary fix.
```

lusassl-msft (approved):
Approved. Suggested a minor change on the documentation path but this isn't a blocker for merging this PR into main.

lusassl-msft (approved):

PR #1956, opened 2024-01-25
Adding to Test-ExchAVExclusions.ps1 new files to allow dlls 

iserrano76:
**Issue:**
Windows 2022 use a new dll that does not include Company.
Add a missing Oracle Dll from FIPS

**Reason:**
Avoid false positives

**Fix:**
Added dlls to the allowed list

**Validation:**
Test in lab:
Previous version:
image

image

New Version:
image


Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -341,6 +342,7 @@ while ($currentDiff -gt 0) {
         $ModuleAllowList.add("SCCOLE2.dll")
         $ModuleAllowList.add("SCCSD.dll")
         $ModuleAllowList.add("SCCXT.dll")
+        $ModuleAllowList.add("vshtml.dll")
```
dpaulson45, line 345: looks like there are a lot of possible dlls that we can load from that location (`C:\Program Files\Microsoft\Exchange Server\V15\FIP-FS\Bin\TE_v.8.5.3.0`) might be best to avoid this happening again moving forward is to account for this somehow.

dpaulson45: Let me know if you want it to be apart of this PR or not.

iserrano76: I think we can check on FIP-FS if the company is Oracle, as we have several dlls there from oracle 

image

image

dpaulson45 (dismissed):
This looks good to me. But want another approver as well.

bill-long:
Unrelated to these changes, the process filter on line 341 should be up on line 311, so it's immediately clear to the reader that other processes are being excluded, and so the loop doesn't call Write-Progress 229 times (in my lab) when we're only looking at 51 processes. This can be done in a future PR. Just noting it here while I'm looking at it.

image

image

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -360,6 +346,9 @@ while ($currentDiff -gt 0) {
                 # Remove Microsoft modules
                 $ProcessModules = $ProcessModules | Where-Object { $_.FileVersionInfo.CompanyName -ne "Microsoft Corporation." -and $_.FileVersionInfo.CompanyName -ne "Microsoft" -and $_.FileVersionInfo.CompanyName -ne "Microsoft Corporation" }
 
+                # Remove Oracle modules on FIPS
+                $ProcessModules = $ProcessModules | Where-Object { (($_.FileName -notlike "*\FIP-FS\Bin\*" -and $_.FileVersionInfo.CompanyName -ne "Oracle Corporation")) }
```
bill-long, line 350: Should we care if there are Oracle modules loaded into the process and those modules are _not_ in FIP-FS\Bin? Because this will exclude them no matter where they are.

image

bill-long (changes requested):

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
dpaulson45: Correct, so the logic that we need to have is the following: 

```powershell
(-not($_.FileName -like "*\FIP-FS\Bin\*" -and $_.FileVersionInfo.CompanyName -eq "Oracle Corporation")) 
```

@iserrano76 please make this change and we can approve it.

dpaulson45:
> > Unrelated to these changes, the process filter on line 341 should be up on line 311, so it's immediately clear to the reader that other processes are being excluded, and so the loop doesn't call Write-Progress 229 times (in my lab) when we're only looking at 51 processes. This can be done in a future PR. Just noting it here while I'm looking at it.
> > image
> > image
> 
> Agree, indeed we could simplify this with something like: image
> 
> As it is just a simple line, I can do it here or do another PR as you prefer.

let's do this in a different PR.

dpaulson45 (approved):

PR #2010, opened 2024-02-17
Rewrite the module Allow list

iserrano76:
**Issue:**
In some cases IpcManagedAPIAsync.ni.dll is load in the Exchange process, as the company name is "Microsoft Corp."
we need to include that string in the Allow list company.
image

**Reason:**
We have some files in the Exchange folders that are not included in the Allow list.

**Fix:**
To avoid this error and future errors I analyzed all files EXE and DLLs in the Exchange folder to extract all company names in 2019/2016 to avoid future issues:
2019:
image

2016:
image

Additionally, I collected all files without company name to be included in the Allow module list.

**Validation:**
Tested in lab:
image


lusassl-msft:
@iserrano76 I'm not sure how accurate this is. This is what I get if I do the same on an Exchange 2019 box:

image

dpaulson45 (approved):

PR #2025, opened 2024-02-27
Under some specific circumstances test is not detected correctly

iserrano76:
**Issue:**
Under some circumstances the AMSI request in the test is not detected correctly.

**Reason:**
Avoid false positives.

**Fix:**
Change the way we do the test request.

**Validation:**
Test in lab.


dpaulson45:
Please explain why doing it this way avoids the false positives from occurring. It just looks like we are making the URL request stream more complex is all.

dpaulson45 (approved):

PR #1930, opened 2024-01-07
Remove permissions without piping

bill-long:
Fixes #1881

Once upon a time, we had to pipe invalid permissions to get them to remove. Testing in my lab on current versions of Exchange indicates this is no longer the case, so I'm removing the pipe. This will allow permissions removal to work in Remote PowerShell.

If we find that piping is still necessary for some scenarios, we may have to revisit this change. This will depend on user feedback.

PR #2039, opened 2024-03-13
SourceSideValidations SDS related requirements added

lusassl-msft:
**Issue:**
`./SourceSideValidations.ps1 -RemoveInvalidPermissions` will fail to run if it was executed on an Exchange Server that has the management tools installed but no other Exchange Server role.

**Reason / Fix:**
In `-RemoveInvalidPermission`, we pipe the invalid permissions to `Remove-PublicFolderClientPermission` precisely because there is no other way to remove invalid permissions. You cannot specify them by user, because the user is an orphaned SID. Specifying that as the `-User` parameter causes us to try to look up the SID to validate the parameter, which fails, so the cmdlet does not succeed. Thus, `ForEach-Object` will not work here. 
  
Piping the bad permission to `Remove-PublicFolderClientPermission` is the only way to get it to treat that orphaned object as valid input. 
  
`SourceSideValidations.ps1` will need to be run from a Mailbox role going forward, at least if you need to use the `-RemoveInvalidPermissions` switch.


lusassl-msft:
Closing this PR since this issue was fixed via:

PR #1930  and PR #1924

PR #2061, opened 2024-04-15
ShaneFe CalLogSummary - split files

Shanefe:
**Issue:**
Main file was really long file, splitting into multiple files. 

**Reason:**
Following best practices, easier PR's 

**Fix:**
No code was changed, just moved about

**Validation:**
Validated vs my test tenant


Calendar/CalLogHelpers/Build-CalLogCSV.ps1
```diff
@@ -0,0 +1,266 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+# Build-CalLogCSV.ps1
+# This script is used to support the Get-CalendarDiagnosticObjectsSummary.ps1 script.
```
dpaulson45, line 4: Keep in mind that this will be included in script in the build process. So, this line will be repeated over and over again, which might not look the best. Our Copyright lines are removed so they aren't repeated.

Calendar/CalLogHelpers/Build-CalLogCSV.ps1
```diff
@@ -0,0 +1,266 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+# Build-CalLogCSV.ps1
```
dpaulson45, line 3: This doesn't need to change, but `Build-CalLogCSV` isn't a function within this file. If it is just a collection of functions, a short name + functions is what we typically do. 

Example: https://github.com/microsoft/CSS-Exchange/blob/main/Shared/LoggerFunctions.ps1

Not saying this is what you need to do, but just something to consider.

Shanefe: Done.   Thanks.  That is better than forcing the File names into the Verb-Nown  of PS actions.

Calendar/CalLogHelpers/Build-CalLogCSV.ps1
Shanefe: Done.

Calendar/CalLogHelpers/MeetingSummaryFunctions.ps1
```diff
@@ -0,0 +1,105 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+# Build-MeetingSummary.ps1
+# This script is used to support the Get-CalendarDiagnosticObjectsSummary.ps1 script.
```
dpaulson45, line 4: Since you removed them everywhere else, let's do it here as well.

Shanefe: Done.

PR #2068, opened 2024-04-22
Extended Protection Updates

### LINKED ISSUE #2033: Extended Protection - Add SuppressExtendedProtection registry key check and Fix Get-OutlookAnywhere check

Found out about this key: https://learn.microsoft.com/en-us/troubleshoot/windows-server/windows-security/authentication-fails-non-windows-ntlm-kerberos-server

We need to check for this key, if it is set, we need to block configuring Exchange for EP, as it will break the environment. 

Also need to block configuring EP if an already configured EP server, that we aren't currently configuring, has `SSLOffloading` set to `true`. This was reported by a customer. 

**Additional context**
NA

dpaulson45:
**Issue:**
We want to block EP from being configured if this key `SuppressExtendedProtection` is set.

Also needed to address issue with Outlook Anywhere not blocking on already configured EP servers.

**Reason:**
Avoid customers from enabling EP anymore without addressing the problems. 

Resolved #2033 

**Validation:**
Lab tested

### DIFF
--- a/Security/src/ExchangeExtendedProtectionManagement/DataCollection/Get-ExtendedProtectionPrerequisitesCheck.ps1
+++ b/Security/src/ExchangeExtendedProtectionManagement/DataCollection/Get-ExtendedProtectionPrerequisitesCheck.ps1
@@ -75,6 +75,7 @@ function Get-ExtendedProtectionPrerequisitesCheck {
 
                 if ($null -eq $lmValue) { $lmValue = 3 }
 
+                Write-Verbose "Server $($server.FQDN) LmCompatibilityLevel set to $lmValue"
                 $registryValues.SuppressExtendedProtection = $epValue
                 $registryValues.LmCompatibilityLevel = $lmValue
             } else {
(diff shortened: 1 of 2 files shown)


Security/src/ExchangeExtendedProtectionManagement/DataCollection/Get-ExtendedProtectionPrerequisitesCheck.ps1
```diff
@@ -60,6 +65,18 @@ function Get-ExtendedProtectionPrerequisitesCheck {
...
+                    SubKey      = "SYSTEM\CurrentControlSet\Control\Lsa"
+                }
+
+                $lmValue = Get-RemoteRegistryValue @params -GetValue "LmCompatibilityLevel" -ValueType "DWord"
+                [int]$epValue = Get-RemoteRegistryValue @params -GetValue "SuppressExtendedProtection"
+
+                if ($null -eq $lmValue) { $lmValue = 3 }
+
+                $registryValues.SuppressExtendedProtection = $epValue
+                $registryValues.LmCompatibilityLevel = $lmValue
```
lusassl-msft, line 79: @dpaulson45 we query the `LmCompatibilityLevel` value but don't write the result out or use it for further testing. Is this something that we'll do with an future update?

lusassl-msft (approved):

PR #2075, opened 2024-04-24
Improve handling of ECC certificates in HealthChecker

lusassl-msft:
**Issue:**
Starting with the Exchange Server April 2024 Hotfix Update (HU), ECC certificates can be used on Exchange Server 2016 and Exchange Server 2019. The latest version of HealthChecker shows an error if an ECC certificate is assigned to an Exchange service (e.g., SMTP).

**Reason:**
The `PublicKeySize` is returned as `0` for ECC certificates (by design).

**Fix:**
Detect if a certificate is an ECC certificate. We do this by comparing the `WellKnownOid`. If it matches `1.2.840.10045.2.1` (Oid for ECC), we exclude the Key Size check for this certificate.

**Validation:**
Lab

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityExchangeCertificates.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityExchangeCertificates.ps1
@@ -106,7 +106,10 @@ function Invoke-AnalyzerSecurityExchangeCertificates {
             Add-AnalyzedResultInformation @params
         }
 
-        if ($certificate.PublicKeySize -lt 2048) {
+        # We show the 'Key Size' if a certificate is RSA or DSA based but not for ECC certificates where it would be displayed with a value of 0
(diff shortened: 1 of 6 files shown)


Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityExchangeCertificates.ps1
```diff
@@ -130,6 +133,15 @@ function Invoke-AnalyzerSecurityExchangeCertificates {
             Add-AnalyzedResultInformation @params
         }
 
+        if ($null -ne $certificate.IsEccCertificate) {
```
dpaulson45, line 136: `$certificate.IsEccCertificate` appears to be a `bool` so if set to `false`, this will still display. What are the scenarios where this is going to be `null`?

lusassl-msft: Right. The null-value check was a leftover from initial testing. Will be removed with the next commit.

dpaulson45 (approved):

PR #2078, opened 2024-04-30
Calendar RBA update

Shanefe:
Fix spacing in Output file.
Save Summary in file.
Translate CN to SMTP Addresses.

### DIFF
--- a/Calendar/Get-RBASummary.ps1
+++ b/Calendar/Get-RBASummary.ps1
@@ -46,10 +46,12 @@ function ValidateMailbox {
     # check we get a response
     if ($null -eq $script:Mailbox) {
         Write-Host -ForegroundColor Red "Get-Mailbox returned null. Make sure you Import-Module ExchangeOnlineManagement and  Connect-ExchangeOnline. Exiting script."
+        Stop-Transcript
         exit
     } else {
         if ($script:Mailbox.RecipientTypeDetails -ne "RoomMailbox" -and $script:Mailbox.RecipientTypeDetails -ne "EquipmentMailbox") {
             Write-Host -ForegroundColor Red "The mailbox is not a Room Mailbox / Equipment Mailbox. RBA will only work with these. Exiting script."
+            Stop-Transcript
             exit
         }
         if ($script:Mailbox.ResourceType -eq "Workspace") {
@@ -67,6 +69,7 @@ function ValidateMailbox {
         Write-Host -ForegroundColor Red "Make sure you are running from the correct forest.  Get-Place does not cross forest boundaries."
         Write-Host "Hint Forest is likely something like: [$($script:Mailbox.Database.split("DG")[0])]."
         Write-Error "Exiting Script."
+        Stop-Transcript
         exit
     }
 
@@ -85,6 +88,7 @@ function ValidateInboxRules {
         Write-Host -NoNewline "Rule to look into: "
         Write-Host -ForegroundColor Red "$($rules.Name -like "Delegate Rule*")"
         Write-Host -ForegroundColor Red "Exiting script."
+        Stop-Transcript
         exit
     } elseif ($rules.Name -like "REDACTED-*") {
         Write-Host -ForegroundColor Yellow "Warning: No PII Access to MB so cannot check for Delegate Rules."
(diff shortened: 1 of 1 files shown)


Calendar/Get-RBASummary.ps1
```diff
@@ -92,14 +92,14 @@ function ValidateInboxRules {
         exit
     } elseif ($rules.Name -like "REDACTED-*") {
         Write-Host -ForegroundColor Yellow "Warning: No PII Access to MB so cannot check for Delegate Rules."
-        Write-Host -ForegroundColor Red " --- Inbox Rules needs to be checked manually for any Delegate Rules. --"
         Write-Host -ForegroundColor Yellow "To gain PII access, Mailbox is located on $($mailbox.Database) on server $($mailbox.ServerName)"
-        if ($rules.count -eq 1) {
+        if ($null -eq $rules.count) {
```
dpaulson45, line 96: I think it would be better to include both statements here. 

```
$null -eq $rules.count -or $rules.count -eq 1
``

This way if `$rules.count` does equal 1 we catch it here.

File: Calendar/Get-RBASummary.ps1 [hunk 2]
```diff
@@ -266,7 +280,10 @@ function InPolicyProcessing {
         Write-Host "`t BookInPolicy:                     {$($RbaSettings.BookInPolicy)}"
     } else {
         Write-Host "`t BookInPolicy:                     These $($RbaSettings.BookInPolicy.count) accounts do not require the delegate approval."
-        foreach ($BIPUser in $RbaSettings.BookInPolicy) { Write-Host " `t `t $BIPUser " }
+        foreach ($BIPUser in $RbaSettings.BookInPolicy) {
+            $BIPUserMB = Get-Mailbox -Identity $BIPUser
+            Write-Host " `t `t [$($BIPUserMB.DisplayName)] -- $($BIPUserMB.PrimarySmtpAddress)"
+        }
```
dpaulson45, line 286: You are doing this same loop over and over again. Just a different list. I would make this info a function that you pass a list of loop over that and have it do everything you want.

Calendar/Get-RBASummary.ps1
```diff
@@ -32,17 +32,26 @@ if (Test-ScriptVersion -AutoUpdate) {
 
 Write-Verbose "Script Versions: $BuildVersion"
 
+$SummaryFilename = "RBA-Summary-For_$($Identity.Split('@')[0])_$((Get-Date).ToString('yyyy-MM-dd_HH-mm-ss')).txt"
+Write-Host "`r`nRBA Summary Output saved as [" -NoNewline
+Write-Host -ForegroundColor Cyan $SummaryFilename -NoNewline
+Write-Host "] in the current directory."
+Start-Transcript -Path $SummaryFilename
```
dpaulson45, line 39: I never really used `Start-Transcript` but does it include `Write-Verbose` in the log output? If it doesn't, you should look at the logging function that we created in Shared.

Calendar/Get-RBASummary.ps1
Shanefe: PS is weird if you have an "Array" with count of 1.  $rules.count will not return any value, if there is only 1 value. 
Let me look if I can predefine it as an array to by pass this weirdness.

Calendar/Get-RBASummary.ps1
Shanefe: Done

Calendar/Get-RBASummary.ps1
dpaulson45: Yeah, it is auto defining the then as NOT array, so count doesn't exist. So, the best way to handle this is by making it always be an array. 

```
$rules = @(Get-InboxRule)

OR 

[array]$rules = Get-InboxRule
```

Calendar/Get-RBASummary.ps1
Shanefe: Start-Transcript does include Verbose output (assuming you pass in -Verbose). 
It is quick and easy, but I will look for the Logging function in Shared for future scripts.

Calendar/Get-RBASummary.ps1
Shanefe: Ah... now I see what you are saying. 
-Fixed.

Shanefe: Belt and suspenders now.

dpaulson45 (approved):

PR #2090, opened 2024-05-09
Added M365 connections to shared

iserrano76:
**Issue:**
Added two functions to Shared in case any script needs a connection to M365.
One function for Exchange Online another function for AzureAD

**Reason:**
Unify the way we connect to M365.

**Fix:**
Both functions will do the following:
- Verify if the module is installed, if not, install it after request confirmation.
- Verify if the module is loaded, if not, load it. 
- Lastly, verify if any connection is present (we verify only one connection is present); if none, we offer to connect.
Important: we do not disconnect at anytime we just verify if we have an active session

### DIFF
--- a/Shared/Connect-M365.ps1
+++ b/Shared/Connect-M365.ps1
@@ -1,0 +1,159 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+function Test-EXOConnection {
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
+    [OutputType([bool])]
+    param (
+        [Switch]$Force
+    )
+    #Validate EXO V2 is installed
+    if ((Get-Module -ListAvailable | Where-Object { $_.Name -like "ExchangeOnlineManagement" }).count -ge 1) {
+        Write-Host "ExchangeOnlineManagement Powershell Module installed"
+    } else {
+        if ($Force -or $PSCmdlet.ShouldContinue("Do you want to install the module?", "ExchangeOnlineManagement Powershell Module not installed")) {
+            Install-Module -Name ExchangeOnlineManagement -Force -ErrorAction SilentlyContinue -Scope CurrentUser
+            if ((Get-Module -ListAvailable | Where-Object { $_.Name -like "ExchangeOnlineManagement" }).count -ge 1) {
+                Write-Host "ExchangeOnlineManagement Powershell Module installed"
+            } else {
+                Write-Host "ExchangeOnlineManagement Powershell Module installation failed" -ForegroundColor Red
+                return $false
+            }
+        } else {
+            Write-Host "We cannot continue without ExchangeOnlineManagement Powershell module" -ForegroundColor Red
+            return $false
+        }
+    }
+
+    #Validate EXO V2 is loaded
+    if ((Get-Module | Where-Object { $_.Name -like "ExchangeOnlineManagement" }).count -ge 1) {
+        Write-Host "ExchangeOnlineManagement Powershell Module loaded"
+    } else {
+        Import-Module ExchangeOnlineManagement -ErrorAction SilentlyContinue -Force
+        if ((Get-Module | Where-Object { $_.Name -like "ExchangeOnlineManagement" }).count -ge 1) {
+            Write-Host "ExchangeOnlineManagement Powershell Module Imported"
+        } else {
+            Write-Host "ExchangeOnlineManagement Powershell Module Import failed" -ForegroundColor Red
+            return $false
+        }
+    }
+
+    #Validate EXO V2 is connected or try to connect
+    $connection = $null
+    $connection = Get-ConnectionInformation -ErrorAction SilentlyContinue
+    if ($null -eq $connection) {
+        Write-Host "Please use Global administrator credentials" -ForegroundColor Yellow
+        if ($Force -or $PSCmdlet.ShouldContinue("Do you want to connect?", "We need a ExchangeOnlineManagement connection")) {
+            Connect-ExchangeOnline -ErrorAction SilentlyContinue
+            $connection = Get-ConnectionInformation -ErrorAction SilentlyContinue
+            if ($null -eq $connection) {
+                Write-Host "Connection could not be established" -ForegroundColor Red
+                Write-Host "We cannot continue without ExchangeOnlineManagement Powershell session" -ForegroundColor Red
+                return $false
+            } else {
+                Write-Host "Connected to EXO V2"
+                Write-Host "Session details"
+                Write-Host "Tenant Id: $($connection.TenantId)"
+                Write-Host "User: $($connection.UserPrincipalName)"
+                return $true
+            }
+        } else {
+            Write-Host "We cannot continue without ExchangeOnlineManagement Powershell session" -ForegroundColor Red
+            return $false
+        }
+    } elseif ($connection.count -eq 1) {
+        Write-Host "Connected to EXO V2"
+        Write-Host "Session details"
+        Write-Host "Tenant Id: $($connection.TenantId)"
+        Write-Host "User: $($connection.UserPrincipalName)"
+        return $true
+    } else {
+        Write-Host "You have more than one EXO sessions please use just one session" -ForegroundColor Red
+        return $false
+    }
+}
+
+function Test-AADConnection {
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
+    [OutputType([bool])]
+    param (
+        [Switch]$Force
+    )
+
+    #Validate AzureAD is installed
+    if ((Get-Module -ListAvailable | Where-Object { $_.Name -like "AzureAD" }).count -ge 1) {
+        Write-Host "AzureAD Powershell Module installed"
+    } else {
+        if ($Force -or $PSCmdlet.ShouldContinue("Do you want to install the module?", "AzureAD Powershell Module not installed")) {
+            Install-Module -Name AzureAD -Repository PSGallery -AllowClobber -Force -ErrorAction SilentlyContinue -Scope CurrentUser
+            if ((Get-Module -ListAvailable | Where-Object { $_.Name -like "AzureAD" }).count -ge 1) {
+                Write-Host "AzureAD Powershell Module installed"
+            } else {
+                Write-Host "AzureAD Powershell Module installation failed" -ForegroundColor Red
+                return $false
+            }
+        } else {
+            Write-Host "We cannot continue without AzureAD Powershell module" -ForegroundColor Red
+            return $false
+        }
+    }
+
+    #Validate AzureAD is loaded
+    if ((Get-Module | Where-Object { $_.Name -like "AzureAD" }).count -ge 1) {
+        Write-Host "AzureAD Powershell Module loaded"
+    } else {
+        Import-Module AzureAD -ErrorAction SilentlyContinue -Force
+        if ((Get-Module | Where-Object { $_.Name -like "AzureAD" }).count -ge 1) {
+            Write-Host "AzureAD Powershell Module Imported"
+        } else {
+            Write-Host "AzureAD Powershell Module Import failed" -ForegroundColor Red
+            return $false
+        }
+    }
+
+    #Validate AzureAD is connected or try to connect
+    try {
+        $connection = $null
+        $connection = Get-AzureADTenantDetail -ErrorAction SilentlyContinue
+        if ($null -eq $connection) {
+            Write-Host "Not connected to AzureAD" -ForegroundColor Red
+            Write-Host "We cannot continue without AzureAD Powershell session" -ForegroundColor Red
+            return $false
+        } else {
+            if ($connection.count -eq 1) {
+                Write-Host "Connected to AzureAD"
+                Write-Host "Session details"
+                Write-Host "Tenant: $($connection.DisplayName)"
+                return $true
+            } else {
+                Write-Host "You have more than one AzureAD sessions please use just one session" -ForegroundColor Red
+                return $false
+            }
+        }
+    } catch [Microsoft.Open.Azure.AD.CommonLibrary.AadNeedAuthenticationException] {
+        Write-Host "Please use Global administrator credentials" -ForegroundColor Yellow
+        if ($Force -or $PSCmdlet.ShouldContinue("Do you want to connect?", "We need a AzureAD connection")) {
+            Connect-AzureAD -ErrorAction SilentlyContinue
+            try {
+                $connection = Get-AzureADTenantDetail -ErrorAction SilentlyContinue
+                if ($null -eq $connection) {
+                    Write-Host "Connection could not be established" -ForegroundColor Red
+                    Write-Host "We cannot continue without AzureAD Powershell session" -ForegroundColor Red
+                    return $false
+                } else {
+                    Write-Host "Connected to AzureAD"
+                    Write-Host "Session details"
+                    Write-Host "Tenant: $($connection.DisplayName)"
+                    return $true
+                }
+            } catch [Microsoft.Open.Azure.AD.CommonLibrary.AadNeedAuthenticationException] {
+                Write-Host "Connection could not be established" -ForegroundColor Red
+                Write-Host "We cannot continue without AzureAD Powershell session" -ForegroundColor Red
+                return $false
+            }
+        } else {
+            Write-Host "We cannot continue without AzureAD Powershell session" -ForegroundColor Red
+            return $false
+        }
+    }
+}


Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+                return $false
+            }
+        } else {
+            Write-Host "We cannot continue without AzureAD Powershell module" -ForegroundColor Red
+            return $false
+        }
+    }
+
+    #Validate AzureAD is loaded
+    if ((Get-Module | Where-Object { $_.Name -like "AzureAD" }).count -ge 1) {
```
lusassl-msft, line 102: The AzureAD PowerShell module is deprecated. We should use PowerShell Graph SDK instead or using the existing shared Graph functions to acquire a token and run Graph API calls.

More information: https://techcommunity.microsoft.com/t5/microsoft-entra-blog/important-update-deprecation-of-azure-ad-powershell-and-msonline/ba-p/4094536

dpaulson45: Good point, we should maybe also require a min version of the modules?

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+# Licensed under the MIT License.
+
+function Test-EXOConnection {
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
+    [OutputType([bool])]
+    param (
+        [Switch]$Force
+    )
+    #Validate EXO V2 is installed
+    if ((Get-Module -ListAvailable | Where-Object { $_.Name -like "ExchangeOnlineManagement" }).count -ge 1) {
```
dpaulson45, line 11: We are doing `Get-Module` a lot here. I would prefer to see a function that handles this.

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+        Write-Host "ExchangeOnlineManagement Powershell Module installed"
```
dpaulson45, line 12: For the Test cmdlets, let's avoid writing to the screen of information. This should be done in Verbose information.

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+                return $false
+            }
+        } else {
+            Write-Host "We cannot continue without ExchangeOnlineManagement Powershell module" -ForegroundColor Red
+            return $false
+        }
+    }
+
+    #Validate EXO V2 is loaded
+    if ((Get-Module | Where-Object { $_.Name -like "ExchangeOnlineManagement" }).count -ge 1) {
```
dpaulson45, line 29: Might be better to check this first, then handle if the module just needs to be imported or installed.

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+            return $false
+        }
+    }
+
+    #Validate EXO V2 is connected or try to connect
+    $connection = $null
+    $connection = Get-ConnectionInformation -ErrorAction SilentlyContinue
+    if ($null -eq $connection) {
+        Write-Host "Please use Global administrator credentials" -ForegroundColor Yellow
+        if ($Force -or $PSCmdlet.ShouldContinue("Do you want to connect?", "We need a ExchangeOnlineManagement connection")) {
```
dpaulson45, line 46: Avoid writing to screen. Handle inside `ShouldContinue`

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+            Write-Host "ExchangeOnlineManagement Powershell Module Imported"
+        } else {
+            Write-Host "ExchangeOnlineManagement Powershell Module Import failed" -ForegroundColor Red
+            return $false
+        }
+    }
+
+    #Validate EXO V2 is connected or try to connect
+    $connection = $null
+    $connection = Get-ConnectionInformation -ErrorAction SilentlyContinue
```
dpaulson45, line 43: `Get-ConnectionInformation` appears to be only for EXO v3. Might need to make sure we have that version. 

https://learn.microsoft.com/en-us/powershell/module/exchange/get-connectioninformation?view=exchange-ps

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+        return $true
+    } else {
+        Write-Host "You have more than one EXO sessions please use just one session" -ForegroundColor Red
+        return $false
+    }
+}
+
+function Test-AADConnection {
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
+    [OutputType([bool])]
```
dpaulson45, line 78: Not sure if we always need to have AzureAD/Graph module with our scripts. Might be best to break these up into their own files.

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+    param (
+        [Switch]$Force
```
dpaulson45, line 80: Again, let's not do this.

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+    )
+
+    #Validate AzureAD is installed
+    if ((Get-Module -ListAvailable | Where-Object { $_.Name -like "AzureAD" }).count -ge 1) {
+        Write-Host "AzureAD Powershell Module installed"
+    } else {
+        if ($Force -or $PSCmdlet.ShouldContinue("Do you want to install the module?", "AzureAD Powershell Module not installed")) {
```
dpaulson45, line 87: We are virtually doing the same thing for both of these modules, if possible should see if we can use a function to handle this as well. Just an idea.

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+                Write-Host "Connected to AzureAD"
+                Write-Host "Session details"
+                Write-Host "Tenant: $($connection.DisplayName)"
+                return $true
+            } else {
+                Write-Host "You have more than one AzureAD sessions please use just one session" -ForegroundColor Red
+                return $false
+            }
+        }
+    } catch [Microsoft.Open.Azure.AD.CommonLibrary.AadNeedAuthenticationException] {
```
dpaulson45, line 133: Need to handle other types of catches here.

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
...
+                } else {
+                    Write-Host "Connected to AzureAD"
+                    Write-Host "Session details"
+                    Write-Host "Tenant: $($connection.DisplayName)"
+                    return $true
+                }
+            } catch [Microsoft.Open.Azure.AD.CommonLibrary.AadNeedAuthenticationException] {
+                Write-Host "Connection could not be established" -ForegroundColor Red
+                Write-Host "We cannot continue without AzureAD Powershell session" -ForegroundColor Red
+                return $false
```
dpaulson45, line 152: Basically repeating, should make this into a function to avoid duplicate code.

Shared/Connect-M365.ps1
```diff
@@ -0,0 +1,159 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+function Test-EXOConnection {
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
+    [OutputType([bool])]
+    param (
+        [Switch]$Force
```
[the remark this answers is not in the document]
iserrano76: ShouldContinue is inside the function and if we do not include $force parameter we got a warning:
image
That was the main reason to include it

Shared/Connect-M365.ps1
iserrano76: I will rethink how to split both functions in two different files and use common functions to reduce the duplication

lusassl-msft (changes requested):
Requesting changes as there are a few things that should be changed (see comments).

Shared/Connect-M365.ps1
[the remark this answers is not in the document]
dpaulson45: Gotcha, I think we should be using `ShouldProcess` instead of `ShouldContinue` that might be why that is coming up.

Shared/Connect-M365.ps1
dpaulson45: This is still here, needs to be addressed.

Shared/Connect-M365.ps1
dpaulson45: use `ShouldProcess`

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,51 @@
...
+# Licensed under the MIT License.
+
+function Request-Module {
+    [OutputType([bool])]
+    param (
+        [Parameter(Mandatory = $true)]
+        [string[]]$Modules,
+        [Parameter(Mandatory = $false)]
+        [System.Version]$MinModuleVersion = $null
+    )
```
dpaulson45, line 11: We actually shouldn't need to use all of this. I would think this is all we need to do: 

https://powershellisfun.com/2023/04/24/using-the-requires-statement-in-powershell/

dpaulson45 (changes requested):
Please address all open comments then let me know when this is ready for the next review.

Shared/ModuleHandle.ps1
dpaulson45: Let's update this before we attempt to review this again.

Shared/M365/EXOConnection.ps1
```diff
@@ -0,0 +1,96 @@
...
+        }
+        $newConnection = $connections
+    }
+
+    Write-Verbose "Connected session to Exchange Online"
+    $newConnection.PSObject.Properties | ForEach-Object { Write-Verbose "$($_.Name): $($_.Value)" }
+    if (-not $DoNotShowConnectionDetails) {
+        Show-EXOConnection -Connection $newConnection
+    }
+    return $newConnection
```
dpaulson45, line 81: Why are we returning a connection here? Trying to understand this.

dpaulson45: Big reason why is trying to understand this section of code: 

```
    if ($null -eq $connections -or $AllowMultipleSessions) {
        if ($connections.ModulePrefix -contains $Prefix) {
            Write-Host "You already have a session with the prefix $Prefix" -ForegroundColor Red
            return $null
        } else {
```

dpaulson45 (commented):
There might be some more things that I find later on, but here is a good starting point to address.

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,56 @@
...
+                    Install-Module -Name $module -Scope CurrentUser
+                    $confirmed = Get-InstalledModule -Name $module -ErrorAction Stop
+                    if (-not $confirmed) {
+                        Write-Host "We could not install module: $module" -ForegroundColor Red
+                        $confirmed = $true
+                    }
+                }
+            } catch {
+                Write-Host "Installation process fails. Error: `n$_" -ForegroundColor Red
+                return $false
```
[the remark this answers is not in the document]
iserrano76: I do not break the flow in any function, I return null and leave the decision to the one that use the function to stop if they do not get what they expect.
What do you think?

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,56 @@
...
+            }
+        }
+    }
+    if ($foundError) {
+        return $false
+    }
+    return $true
```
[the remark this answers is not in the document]
iserrano76: About the first comment, agree, it is easier. I will simplify it.

About second one, similar to previous one, if works I return true if fails I return false and leave the decision on the one that use the function.
I think maybe you can try to request another version and it is more flexible. 
What do you think?

Shared/M365/GraphConnection.ps1
```diff
@@ -0,0 +1,178 @@
...
+        [Parameter(Mandatory = $true)]
+        [string[]]$Scopes,
+        [Parameter(Mandatory = $true)]
+        [string[]]$Modules,
+        [Parameter(Mandatory = $false)]
+        [string[]]$TenantId = $null,
+        [Parameter(Mandatory = $false)]
+        [switch]$DoNotShowConnectionDetails,
+        [Parameter(Mandatory = $false)]
+        [System.Version]$MinModuleVersion = 2.0.0.0
```
dpaulson45, line 52: Let's add a script validation here for the min version supported.

Let's also add some version information to get the module that we are currently using, just in case things change with Graph.

Shared/M365/GraphConnection.ps1
```diff
@@ -0,0 +1,178 @@
...
+            if ($connection.TenantId -ne $TenantId) {
+                Write-Host "Connected to $($connection.TenantId). Not expected tenant: $TenantId" -ForegroundColor Red
+                return $null
+            } else {
+                Write-Verbose "TenantId is correct"
+            }
+        }
+
+        if (-not $DoNotShowConnectionDetails) {
+            $connection.PSObject.Properties | ForEach-Object { Write-Verbose "$($_.Name): $($_.Value)" }
```
dpaulson45, line 97: Move this to outside the `if` statement. We should always display this type of information in verbose.

Shared/M365/GraphConnection.ps1
```diff
@@ -0,0 +1,178 @@
...
+            $foundError = $true
+        }
+    }
+
+    if ($foundError) {
+        return $false
+    } else {
+        Write-Verbose "All expected scopes are present."
+        return $true
+    }
```
dpaulson45, line 161:
```suggestion
    Write-Verbose "All expected scopes are $(if($foundError){ "NOT "})present."
    return (-not $foundError)
```

Shared/M365/EXOConnection.ps1
```diff
@@ -0,0 +1,144 @@
...
+                try {
+                    $newConnections = Get-ConnectionInformation -ErrorAction Stop
+                } catch {
+                    Write-Host "We cannot check connections. Error:`n$_" -ForegroundColor Red
+                    return $null
+                }
+                foreach ($testConnection in $newConnections) {
+                    if ($connections -notcontains $testConnection) {
+                        $newConnection = $testConnection
+                    }
```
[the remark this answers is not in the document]
iserrano76: Yes, this is the purpose, in 84 we just get the connection with the expected prefix if it exist before.
In line 106 we are looking for the new connection, we are looking for changes before the list of connection before and after the script execution. the new connection is the connection that we create. 
we are looking only for the connection that we have created.

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,56 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+function Request-Module {
+    [OutputType([bool])]
+    param (
+        [Parameter(Mandatory = $true)]
+        [string[]]$Modules,
+        [Parameter(Mandatory = $false)]
+        [System.Version]$MinModuleVersion = $null
```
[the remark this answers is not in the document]
dpaulson45: Let's also address this prior to merging.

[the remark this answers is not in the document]
dpaulson45: I guess we did address this by being in a loop, but the chances of you having a module that you are requesting be at the same min version is pretty low.

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,84 @@
...
+        Write-Verbose "Checking $module"
+        if ($MinModuleVersion) {
+            $getParams["MinimumVersion"] = $MinModuleVersion
+            Write-Verbose "with minimum version $minModuleVersion"
+        } else {
+            Write-Verbose "without minimum version"
+        }
+        $installed = Get-InstalledModule @getParams
+
+        if ($null -eq $installed -or $installed.Name -notcontains $module) {
```
dpaulson45, line 53: If we did find the module with the version that we are looking for, we should do `Write-Verbose` here that states at least the version to know if it is something newer and possibly something changed with the cmdlets that broke something.

dpaulson45 (commented):
Almost ready to go, after we get the next commit, let's also rebase this down to a single commit.

Shared/ModuleHandle.ps1
dpaulson45: @iserrano76 let's do this and rebase and then we should be good for one last review before merge.

Shared/ModuleHandle.ps1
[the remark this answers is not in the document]
dpaulson45: Are they the same version? The chances that you need to request multiple modules with the same exact version number is slim.

Shared/ModuleHandle.ps1
```diff
@@ -29,19 +29,18 @@ function Request-Module {
...
     param (
         [Parameter(Mandatory = $true)]
-        [string[]]$Modules,
+        [string[]]$Module,
         [Parameter(Mandatory = $false)]
         [System.Version]$MinModuleVersion = $null
     )
 
     $noFoundError = $true
     foreach ($module in $Modules) {
```
dpaulson45, line 38: You broke this loop because you changed the parameter name to be `$Module` instead of `$Modules`.

iserrano76: Thanks for point to this, I am modifying the parameter to singular as you recommend in a previous PR from the Development Guidelines.
I will fix this and rebase

dpaulson45:
@iserrano76 let's go ahead and rebase and squash this down to a single commit.

Shared/ModuleHandle.ps1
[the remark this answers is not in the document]
iserrano76: yes,
 alt text
we have 3moduels and at this point we have only version 1.0, 3 modules with same version.
is this what you mean, rigth?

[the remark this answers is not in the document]
dpaulson45: Okay, I guess I stand corrected that these modules are all on the same version.

Also, the latest version is 2.25.

dpaulson45:
Pending review from Lukas as well.

Shared/ModuleHandle.ps1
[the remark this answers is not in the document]
lusassl-msft: We could also think about removing the `MinModuleVersion` parameter and instead passing a key-value pair via `Module` parameter. This could be a dictionary (because they are strongly typed):

```
$modules = New-Object "System.Collections.Generic.Dictionary[String, Version]"
$modules.Add("ModuleToLoad", "3.0.0.0")
```

We could then progress them one by one

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,85 @@
...
+            Name        = $m
+            ErrorAction = 'SilentlyContinue'
+        }
+        if ($MinModuleVersion) {
+            $getParams["MinimumVersion"] = $MinModuleVersion
+            Write-Verbose "with minimum version $minModuleVersion"
+        } else {
+            Write-Verbose "without minimum version"
+        }
+        $installed = Get-InstalledModule @getParams
```
lusassl-msft, line 50: We run the `Get-InstalledModule` multiple times. Can we run it just one time to query information about all available modules before we enter the `foreach`, cache the information and then continue processing based on the cached information within the `foreach`?

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,85 @@
...
+
+        if ($null -eq $installed -or $installed.Name -notcontains $m) {
+            Write-Host "The following module is missing: $m" -ForegroundColor Yellow
+            $confirmed = $null
+            try {
+                Write-Verbose "Installing $m"
+                $installParams = @{
+                    Name        = $m
+                    Scope       = "CurrentUser"
```
lusassl-msft, line 59: Not sure if this could be useful in future, can we make the scope adjustable via parameter (`CurrentUser` could be set as default here). What do you think?

lusassl-msft (changes requested):

Shared/ModuleHandle.ps1
iserrano76: It should work, I will test it. Whit the current script we will check every time only will be better if we request the same module or in similar situation. Looks better your suggestion.

Shared/ModuleHandle.ps1
iserrano76: I use CurrentUser because if we use Alluser it requires elevation and we will need to check it before.
in the current scripts we do not need elevation to have the script working and that was the main reason to avoid Alluser, but if you think it could be useful for other scripts I can include as a parameter.

Shared/ModuleHandle.ps1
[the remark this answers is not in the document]
iserrano76: Version is not mandatory.
If you want to check one by one you can use 
Request-Module -Module moduletoload1 -Version 1.0.0
Request-Module -Module moduletoload2 -Version 2.0.0

but if both has the same version just do the request in the same one:
Request-Module -Module moduletoload1, moduletoload2 -Version 1.0.0

It looks simplier in this way but if you think a key-value is easier we can change it

dpaulson45:
@lusassl-msft to review the final result, then we will squash this back down a commit and then merge.

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,98 @@
...
+        [switch]$InstallAllUsersIfNotAvailable
+    )
+
+    $noFoundError = $true
+
+    foreach ($m in $Module) {
+        Write-Verbose "Checking $m PowerShell Module"
+        $getParams = @{
+            Name        = $m
+            ErrorAction = 'SilentlyContinue'
```
lusassl-msft, line 47: @iserrano76 should we change this to `ErrorAction = Stop` and put the `Get-InstalledModule` into a `try/catch`? If we fail to query the modules for whatever reason, we treat it the same as if the module was not found. I think we should change this behavior.

Shared/ModuleHandle.ps1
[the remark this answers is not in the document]
lusassl-msft: I'm okay with both approaches. If you want to keep the current approach I'm okay with it.

Shared/ModuleHandle.ps1
lusassl-msft: I think there are use cases in the future but I'm okay with keeping the current approach for the first version of this script.

lusassl-msft (changes requested):

Shared/ModuleHandle.ps1
dpaulson45: If we are handling it properly, does it matter if we do the `try/catch` version?

lusassl-msft: No, it shouldn't matter then. If I understand the code correctly, we're trying to install the module if `$installed` is `$null` or does not contain `$m` (module x). Installed would be `$null` in case that `Get-InstalledModule` fail for whatever reason. This would mean we're trying installing a module which might be already there. That's my understanding here. Please correct me if I'm wrong.

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,112 @@
...
+        [string[]]$Module,
+        [Parameter(Mandatory = $false)]
+        [System.Version]$MinModuleVersion = $null,
+        [Parameter(Mandatory = $false)]
+        [switch]$InstallAllUsersIfNotAvailable
+    )
+
+    $noFoundError = $true
+
+    foreach ($m in $Module) {
```
lusassl-msft, line 43: Sorry to be so petty. Wouldn't it be better to return a final statement for each module that was passed when calling the `Request-Module` function? We support processing multiple modules in here but return just one state. If one of the modules fail and the remaining ones are processed successfully, we just return `$false` because one of them has failed. I think it would be better to put the module and final state to a hashtable and finally return it. What do you think?

iserrano76: Hi @lusassl-msft do not worry, it is a healthy discussion looking for the best solution, no problem at all 😉
Maybe it is easier to simplify and just allow one module each time instead of an array and return true or false for each one.
I think we are not going to request several modules.
What do you think?

dpaulson45: That could make it more complicated for the caller to use. If that is the case, I would just adjust to only accept a single module and you have to test against each one.

lusassl-msft: Yep, let’s move to the one module approach then 👍

dpaulson45: @iserrano76 are you able to make this change?

iserrano76: Yes, I am trying to find some time for testing.

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,105 @@
...
+# Licensed under the MIT License.
+
+<#
+.SYNOPSIS
+This script defines a function `Request-Module` that checks for the presence of specified PowerShell modules.
+If a module is not found, it attempts to install it on current user scope.
+The function accepts a list of module names and an optional minimum version for the modules.
+It returns a boolean indicating whether all specified modules are installed successfully.
+
+.PARAMETER Modules
```
dpaulson45, line 11:
```suggestion
.PARAMETER Module
```

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,105 @@
...
+ Mandatory array of strings specifying the names of the modules to check and install if necessary.
+.PARAMETER MinModuleVersion
+ Optional parameter to specify the minimum version of the modules (default is null).
+
+.OUTPUTS
+bool. A boolean indicating whether all specified modules are installed successfully.
+
+.EXAMPLE
+$requestModule = Request-Module -Modules "ExchangeOnlineManagement"
```
dpaulson45, line 20:
```suggestion
$requestModule = Request-Module -Module "ExchangeOnlineManagement"
```

Shared/ModuleHandle.ps1
```diff
@@ -0,0 +1,105 @@
...
+This example checks if the "ExchangeOnlineManagement" module is installed. If it is not found, the script attempts to install it.
+
+.EXAMPLE
+$requestModule = Request-Module -Modules "ExchangeOnlineManagement" -MinModuleVersion $MinModuleVersion
```
dpaulson45, line 24:
```suggestion
$requestModule = Request-Module -Module "ExchangeOnlineManagement" -MinModuleVersion $MinModuleVersion
```

dpaulson45:
@lukak-msft please review as soon as you can.

@iserrano76 please squash the commits down to 1 so we can clean this up and make it ready to merge once we are done reviewing.

dpaulson45:
@iserrano76 please squash this down to a single commit, then we will review and merge.

Shared/M365/EXOConnection.ps1
```diff
@@ -0,0 +1,150 @@
...
+
+    if ($null -eq $connections -or $AllowMultipleSessions) {
+        if ($connections.ModulePrefix -contains $Prefix) {
+            Write-Host "You already have a session" -ForegroundColor Yellow -NoNewline
+            if ($Prefix) {
+                Write-Host " with the prefix $Prefix." -ForegroundColor Yellow
+            } else {
+                Write-Host " without prefix." -ForegroundColor Yellow
+            }
+            $newConnection = $connections | Where-Object { $_.ModulePrefix -eq $Prefix }
```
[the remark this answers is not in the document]
iserrano76: to be honest it should never happens becuase Exchange module handle to have have only one connection per prefix but could be fine

dpaulson45:
@iserrano76 please squash this down to a single commit, then we will review and merge.

dpaulson45:
@iserrano76 please squash this down to a single commit, then we will review and merge. If you don't do this soon, we will close this PR.

PR #2113, opened 2024-06-19
Update Exchange CalLog

Shanefe:
Making them easier to read and process.
Fixed client mapping
Check For known issue: Outlook Bifurcation

### DIFF
--- a/.build/cspell-words.txt
+++ b/.build/cspell-words.txt
@@ -10,6 +10,7 @@ bincirc
 BSTR
 cach
 cachtokn
+Calendrier
 Clixml
 cmatch
 comparables
@@ -45,6 +46,7 @@ FYDIBOHF
 GCDO
 Get-AuthenticodeSignature
 globalsequence
+Kalender
 Hashtable
 HHMM
 HKCR
--- a/.build/cspell-words.txt
+++ b/.build/cspell-words.txt
@@ -10,7 +10,6 @@ bincirc
 BSTR
 cach
 cachtokn
-Calendrier
 Clixml
 cmatch
 comparables
@@ -46,7 +45,6 @@ FYDIBOHF
 GCDO
 Get-AuthenticodeSignature
 globalsequence
-Kalender
 Hashtable
 HHMM
 HKCR
--- a/Calendar/CalLogHelpers/CalLogCSVFunctions.ps1
+++ b/Calendar/CalLogHelpers/CalLogCSVFunctions.ps1
@@ -73,11 +73,7 @@ function CreateExternalMasterIDMap {
 
         if ($AllFolderNames.count -gt 1) {
             # We have 2+ FolderNames, Need to find the best one. Remove 'Calendar' from possible names
-            $AllFolderNames = $AllFolderNames | Where-Object {
-                $_ -notmatch 'Calendar' -and
-                $_ -notmatch 'Calendrier' -and
-                $_ -notmatch 'Kalender'
-            } # Need a better way to do this for other languages...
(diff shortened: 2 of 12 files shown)


.build/cspell-words.txt
```diff
@@ -10,6 +10,7 @@ bincirc
 BSTR
 cach
 cachtokn
+Calendrier
```
dpaulson45, line 13: I would like to see if there is a way to avoid this.

Calendar/CalLogHelpers/CalLogCSVFunctions.ps1
```diff
@@ -193,37 +188,26 @@ function BuildCSV {
...
             'AttendeeCount'                  = ($CalLog.DisplayAttendeesAll -split ';').Count
             'AppointmentState'               = $CalLog.AppointmentState.ToString()
-            'ResponseType'                   = $ResponseType
-            'SentRepresentingEmailAddress'   = $CalLog.SentRepresentingEmailAddress
-            'SentRepresentingSMTPAddress'    = GetSMTPAddress($CalLog.SentRepresentingEmailAddress)
-            'SentRepresentingDisplayName'    = $CalLog.SentRepresentingDisplayName
-            'ResponsibleUserSMTPAddress'     = GetSMTPAddress($CalLog.ResponsibleUserName)
-            'ResponsibleUserName'            = $CalLog.ResponsibleUserName
-            'SenderEmailAddress'             = $CalLog.SenderEmailAddress
-            'SenderSMTPAddress'              = GetSMTPAddress($CalLog.SenderEmailAddress)
```
dpaulson45, line 213: We don't want to know about the sender who requested the change?

Calendar/CalLogHelpers/CalLogCSVFunctions.ps1
```diff
@@ -80,8 +72,12 @@ function CreateExternalMasterIDMap {
...
 
         if ($AllFolderNames.count -gt 1) {
-            # We have 2+ FolderNames, Need to find the best one. #remove Calendar
-            $AllFolderNames = $AllFolderNames | Where-Object { $_ -notmatch 'Calendar' } # This will not work for non-english
+            # We have 2+ FolderNames, Need to find the best one. Remove 'Calendar' from possible names
+            $AllFolderNames = $AllFolderNames | Where-Object {
+                $_ -notmatch 'Calendar' -and
+                $_ -notmatch 'Calendrier' -and
+                $_ -notmatch 'Kalender'
+            } # Need a better way to do this for other languages...
```
dpaulson45, line 80: I am thinking we should be able to line up between `Get-CalendarDiagnosticObjects` output and `Get-MailboxFolderStatistics -Identity NAME -FolderScope Calendar` on the name. That way it should catch all languages without any issues.

dpaulson45 (changes requested):

Calendar/CalLogHelpers/CalLogCSVFunctions.ps1
Shanefe: These are all the RAW forms of information.  The Processed fields are output earlier, Organizer, From, Sender, etc. These were here in case the other did not work, but they are still preserved in the RAW output, so they are just duplicate.

.build/cspell-words.txt
Shanefe: Agree.  Will remove until we get a complete solution.

Calendar/CalLogHelpers/CalLogCSVFunctions.ps1
Shanefe: I have done this before.  I will revert to English only until we can get a complete solution.

dpaulson45 (approved):

PR #2115, opened 2024-06-20
Rba update check

Shanefe:
Update a few minor bits of the RBA diag

### DIFF
--- a/Calendar/Get-RBASummary.ps1
+++ b/Calendar/Get-RBASummary.ps1
@@ -280,9 +280,8 @@ function OutputMBList {
     )
     foreach ($User in $MBList) {
         # MS Support will error as we need the Organization to process from CN
-        if ($Identity -like "*@*") {
-            $Org = $Identity.Split('@')[1]
-        }
+        $Org = $Identity.Split('@')[1]
+
         if ($null -ne $Org) {
             $User = Get-Mailbox -Identity $User -organization $Org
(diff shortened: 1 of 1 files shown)


Calendar/Get-RBASummary.ps1
```diff
@@ -273,14 +272,24 @@ function RBAProcessingValidation {
...
         [string[]]$MBList
     )
     foreach ($User in $MBList) {
-        $User = Get-Mailbox -Identity $User
-        Write-Host " `t `t [$($User.DisplayName)] -- $($User.PrimarySmtpAddress)"
+        # MS Support will error as we need the Organization to process from CN
+        if ($Identity -like "*@*") {
+            $Org = $Identity.Split('@')[1]
+        }
+        if ($null -ne $Org) {
```
dpaulson45, line 286: Bug, need to set `$Org` to `$null` each time in the loop, otherwise `$Org` could be the last `$Identity` in the `$MBList`. 

You should be able to also do this: 
```
# No need for 'if ($Identity -like "*@*")' as split() doesn't return an error
$Org = $Identity.Split("@")[1]
if ($null -ne $Org) {...
```

This way `$Org` is always set correctly in the loop.

dpaulson45: @Shanefe just sending out a reminder.

dpaulson45 (approved):

PR #2116, opened 2024-06-20
Check oldest log date entry to throw a warning

### LINKED ISSUE #2022: [Feature request] Check max log size of the application event log and show a warning if it's set to the default (20 MB)
**Describe The Request**
It was requested that HealthChecker should check if the max log size is set to the recommended value. By default, the maximum application event log size is set to 20 MB which could be problematic in troubleshooting scenarios.

**Additional context**
[Recommended settings for event log sizes in Windows](https://learn.microsoft.com/it-IT/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/dd349798(v=ws.10)#recommended-settings-for-event-log-sizes-in-windows)

dpaulson45:
**Issue:**
If the application and system event logs are getting rolled over too quickly, the customer needs to be aware of this. 

**Reason:**
To bubble up some concerns with diagnostic information to be able to perform RCAs better for customers. 

**Fix:**
Check to see if the logs are at least 7 days old for the oldest entry in the log type. 

Resolved #2022 

**Validation:**
Lab tested/pester tested

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerOsInformation.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerOsInformation.ps1
@@ -487,10 +487,13 @@ function Invoke-AnalyzerOsInformation {
         $days = 7
         $testDate = (Get-Date).AddDays(-$days)
         foreach ($logType in $osInformation.EventLogInformation.Keys) {
-            if ($osInformation.EventLogInformation[$logType].LastLogEntry -gt $testDate) {
+            $logInfo = $osInformation.EventLogInformation[$logType]
+            # If the log isn't at the max limit, it is possible that they just cleared the logs or on a new server. This should be rare scenario.
+            if ($logInfo.LastLogEntry -gt $testDate -and
+                $logInfo.FileSize -ge $logInfo.MaxSize) {
                 $params = $baseParams + @{
                     Name             = "Event Log - $logType"
-                    Details          = "--ERROR-- Not enough logs to cover $days days. Last log entry is at $($osInformation.EventLogInformation[$logType].LastLogEntry)." +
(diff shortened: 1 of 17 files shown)


Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerOsInformation.ps1
```diff
@@ -482,4 +482,23 @@ function Invoke-AnalyzerOsInformation {
...
+        $days = 7
+        $testDate = (Get-Date).AddDays(-$days)
+        foreach ($logType in $osInformation.EventLogInformation.Keys) {
+            $logInfo = $osInformation.EventLogInformation[$logType]
+            # If the log isn't at the max limit, it is possible that they just cleared the logs or on a new server. This should be rare scenario.
+            if ($logInfo.LastLogEntry -gt $testDate -and
+                $logInfo.FileSize -ge $logInfo.MaxSize) {
+                $params = $baseParams + @{
+                    Name             = "Event Log - $logType"
+                    Details          = "--ERROR-- Not enough logs to cover $days days. Last log entry is at $($logInfo.LastLogEntry)." +
```
lusassl-msft, line 496: Should we make this a `Warning` instead of an `Error` as it doesn't cause any issue (but makes RCA harder in case of an issue).

lusassl-msft (approved):
Change looks good to me - approved. There is one question left but this isn't a blocker.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerOsInformation.ps1
dpaulson45: For now, let's leave it as an `Error` till we get feedback from it. In most cases, you should have more than 7 days' worth of application logs.

PR #2156, opened 2024-07-24
Avoid errors in management tools

iserrano76:
Issue:
If you run the script in a Exchange management Shell in a server with just administration tools you receive an error when the script try to get sites of servers.

Reason:
Avoid error in execution in management tools

Fix:
Use the same property for all executions.

Validation:
Execution in Previous version:
image

Execution in New version
image


Admin/Test-AMSI.ps1
```diff
@@ -648,11 +648,7 @@ process {
     }
 
     $uniqueSites = $null
-    if ($localServer) {
-        $uniqueSites = $SupportedExchangeServers.Site.Name | Get-Unique
-    } else {
-        $uniqueSites = $SupportedExchangeServers.Site | Get-Unique | ForEach-Object { $_.split('/')[-1] }
-    }
+    $uniqueSites = $SupportedExchangeServers.Site.Name | Sort-Object | Get-Unique
```
dpaulson45, line 651: This fix will not work. When you are on a tools box, sometimes, you get the proper Exchange Management objects. Other times, you do not and we need to account for both if we are to support the tools box with this script. 

image

dpaulson45 (changes requested):

Admin/Test-AMSI.ps1
iserrano76: I thought we change the behavior because I have 2 new new toolbox and I receive Full Exchange Objects in both.
do you know why sometimes we receive full objects and sometimes not?
I will prepare it in a different way.

dpaulson45: If the versions don't 100% match, it will not be the full object. If you install a SU on Exchange but not on the tools box, which happens all the time, that is why.

Admin/Test-AMSI.ps1
```diff
@@ -661,11 +663,7 @@ process {
                 Write-Warning "We did not find site $site"
             }
         }
-        if ($localServer) {
-            $fullList = ($SupportedExchangeServers | Where-Object { $Sites -contains $_.Site.Name } | Select-Object Name).Name
-        } else {
-            $fullList = ($SupportedExchangeServers | Where-Object { $Sites -contains $_.Site.split('/')[-1] } | Select-Object Name).Name
-        }
+        $fullList = ($SupportedExchangeServers | Where-Object { $Sites -contains $_.Site.Name } | Select-Object Name).Name
```
dpaulson45, line 666: Didn't fix this one.

Admin/Test-AMSI.ps1
iserrano76: oh, ok, now I can reproduce it and test it. Thanks

dpaulson45 (approved):

PR #2170, opened 2024-08-12
Measure email delay in mtl

Canthv0:
**Issue:**
Customers come with MTL files with 100s of messages and the statement that "email is slow"

**Reason:**
Providing a script that will run thru an MTL and show how long each message took to deliver.

**Validation:**
Tested script against multiple MTL file.


Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,172 @@
...
+# Test for the provided file and load it.
+if (Test-Path $MTLFile) {
+    Write-Output "Loading MTL file."
+    $mtl = Import-Csv $MTLFile -Encoding Unicode
+} else {
+    Write-Error "Unable to find specified file"
+}
+
+# Validate the MTL
+if (Test-CSVData -CSV $mtl -ColumnsToCheck "event_id", "source", "message_id", "date_time_utc") {}
```
dpaulson45, line 96: `$mtl` can be $null here which will result in unhandled errors.

dpaulson45, line 96: `"event_id", "source", "message_id", "date_time_utc"` are not typical values in a Message Tracking Log. Is this the expected output from a cmdlet? If so, we need to be more clear about this script/script name. If it isn't intended for detecting Email Delay in all types of Message Tracking Logs, including Exchange On Prem, then we need to make that clear with the name and the parameters being used.

dpaulson45, line 96: We should not have an empty action for an `if` statement. If the intent is to stop here, then put the reverse in the `if` statement and don't do an `else`

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,172 @@
...
+
+if ($uniqueMessageIDs.count -eq 0) {
+    Write-Error "No Unique MessageIDs found in data."
+}
+
+# Carve the data up into smaller collections to make searching faster.
+# Most of what is in the MTL we don't need for this.
+$SMTPReceive = $mtl | Where-Object { ($_.event_id -eq 'RECEIVE') -and ($_.source -eq 'SMTP') }
+$StoreDeliver = $mtl | Where-Object { ($_.event_id -eq 'DELIVER') -and ($_.source -eq 'STOREDRIVER') }
+$SMTPDeliver = $mtl | Where-Object { ($_.event_id -eq 'SENDEXTERNAL') -and ($_.source -eq 'SMTP') }
```
dpaulson45, line 112: I would assume it is faster to have this all in a single `foreach` loop and adding the data that you want to lists when matched vs looping through all the data 4 different times to add to individual lists.

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,172 @@
...
+
+    # extract the times for a message ID ... there can be more than one of each of these.
+    [array]$AllSentTimes = ($SMTPReceive | Where-Object { ($_.message_id -eq $id) }).date_time_utc
+    [array]$AllStoreDeliverTimes = ($StoreDeliver | Where-Object { ($_.message_id -eq $id) }).date_time_utc
+    [array]$AllRemoteDeliverTimes = ($SMTPDeliver | Where-Object { ($_.message_id -eq $id) }).date_time_utc
+
+    # Process the time sent
+    if ($AllSentTimes.count -eq 0) {
+        Write-Warning ($id.message_id.ToString() + "unable to find sent time. Discarding messageID")
+        quit
```
dpaulson45, line 130: `quit` isn't a valid cmdlet. 

image

image

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,172 @@
...
+    } else {
+        $TimeSent = Get-Date ($AllSentTimes | Sort-Object | Select-Object -First 1)
+    }
+
+    # If we didn't find any delivery information then drop the message ID
+    if ($AllStoreDeliverTimes.count -eq 0 -and $AllRemoteDeliverTimes.count -eq 0) {
+        Write-Warning ($id + "not able to find delivery time in MTL. Discarding messageID")
+    }
+    # Process the message information
+    else {
```
dpaulson45, line 140:
```suggestion
    } else {
     # Process the message information
```

dpaulson45: Correct formatting.

Transport/Measure-EmailDelayInMTL.ps1
dpaulson45, line 130: By the looks of things, this should be a `continue` to move to the next entry.

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,172 @@
...
+
+        # Combine all of the delivery times.
+        [array]$AllDeliveries = ($AllStoreDeliverTimes + $AllRemoteDeliverTimes) | Sort-Object
+
+        $report = [PSCustomObject]@{
+            MessageID    = $id
+            TimeSent     = $TimeSent
+            TimeReceived = (Get-Date $AllDeliveries[-1])
```
dpaulson45, line 148: You are assuming that the time is in order. How do we know this is ALWAYS true?

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,172 @@
...
+    [array]$AllSentTimes = ($SMTPReceive | Where-Object { ($_.message_id -eq $id) }).date_time_utc
+    [array]$AllStoreDeliverTimes = ($StoreDeliver | Where-Object { ($_.message_id -eq $id) }).date_time_utc
+    [array]$AllRemoteDeliverTimes = ($SMTPDeliver | Where-Object { ($_.message_id -eq $id) }).date_time_utc
+
+    # Process the time sent
+    if ($AllSentTimes.count -eq 0) {
+        Write-Warning ($id.message_id.ToString() + "unable to find sent time. Discarding messageID")
+        quit
+    } else {
+        $TimeSent = Get-Date ($AllSentTimes | Sort-Object | Select-Object -First 1)
```
dpaulson45, line 132: `$AllSentTimes` is an array of string values. If you are trying to sort this correctly, you must convert the data to a `DateTime` type object.

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,172 @@
...
+            MessageDelay = (Get-Date $AllDeliveries[0]) - $TimeSent
+        }
+
+        [array]$output = [array]$output + $report
+    }
+}
+
+# Export the data to the output file
+$outputFile = (Join-Path -Path $ReportPath -ChildPath "MTL_report.csv")
+$output | Export-Csv -IncludeTypeInformation:$false -Path $outputFile
```
dpaulson45, line 158: what happens if there is already a report there?

Transport/Measure-EmailDelayInMTL.ps1
Canthv0: Added null/empty check to Test-CSVData function.

Canthv0: removed empty if

Transport/Measure-EmailDelayInMTL.ps1
Canthv0: Yep needed a continue not a quit (oddly enough the quit was working) ... that is what I get for working off memory.

Transport/Measure-EmailDelayInMTL.ps1
Canthv0: not assuming that ... the array is sorted before we pull from it.
But I moved the logic around a bit and changed the variable names to call out that they are sorted.

Transport/Measure-EmailDelayInMTL.ps1
Canthv0: All testing to this point has had no issues when loading in Unicode files from EXO.
Testing now is failing ... so guessing I had some other error that was masking this problem.

Added code to convert to [DateTime] type

Transport/Measure-EmailDelayInMTL.ps1
Canthv0: Most email will only have two events that we care about.
The rest is just garbage for the purpose of this script (~4-8 lines depending)
Doing it this way quickly gets rid of the events that we don't want so that when we filter by messageID we are not reading thru 100s or 1000s of events that are not relevant.

Six of one half dozen the other this is just how it makes sense to do this to me.

Transport/Measure-EmailDelayInMTL.ps1
Canthv0: added onprem MTL detection.  Detects if the MTL is from an onprem PS command and updates the info in memory to work just like an EXO one.

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,172 @@
...
+            TimeReceived = (Get-Date $AllDeliveries[-1])
+            MessageDelay = (Get-Date $AllDeliveries[0]) - $TimeSent
+        }
+
+        [array]$output = [array]$output + $report
+    }
+}
+
+# Export the data to the output file
+$outputFile = (Join-Path -Path $ReportPath -ChildPath "MTL_report.csv")
```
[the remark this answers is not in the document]
Canthv0: Added code to validate output path at start of script.

Transport/Measure-EmailDelayInMTL.ps1
Canthv0: Initially thought was to just overwrite and move on.
Now we append filetime that is going to be unique.

Transport/Measure-EmailDelayInMTL.ps1
Canthv0: all formatting passed.

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,218 @@
...
+    else {
+        Write-Host "Loaded CSV without Unicode"
+    }
+} else {
+    Write-Host "Loaded MTL with Unicode"
+}
+
+# Detecting if this is an onprem MTL
+if (Test-CSVData -CSV $mtl -ColumnsToCheck "eventid", "source", "messageId", "timestamp") {
+    Write-Host "On Prem message trace detected; Updating property names"
```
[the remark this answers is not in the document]
dpaulson45: Please provide examples in docs of how we should be expecting this script to work for Exchange OnPrem message tracking.

docs/Transport/Measure-EmailDelayInMTL.md
```diff
@@ -0,0 +1,58 @@
...
+
+## PARAMETER
+
+**-MTLFile**
+
+MTL File to process.
+
+**-ReportPath**
+
+Folder path for the output file.
```
dpaulson45, line 21: Let's use a similar format from SourceSideValidations.md. That way we are a little consistent with our work.

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,218 @@
...
+# If it is null then we need to try without Unicode
+if ($null -eq $mtl) {
+    Write-Host "Failed to Load as Unicode; trying normal load"
+    $mtl = Import-Csv $MTLFile
+    # If we still have nothing then log an error and fail
+    if ($null -eq $mtl) {
+        Write-Error "Failed to load CSV" -ErrorAction Stop
+    }
+    # Need to know that we loaded without Unicode.
+    else {
```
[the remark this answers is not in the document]
Canthv0: Moved comment out of the else and above the if.

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,218 @@
...
+Write-Host "Converting date_time_utc values"
+for ($i = 0; $i -lt $mtl.Count; $i++) {
+    $mtl[$i].date_time_utc = Get-Date($mtl[$i].date_time_utc)
+}
+
+# get all of the unique message IDs in the file.
+[array]$uniqueMessageIDs = $mtl | Select-Object -ExpandProperty message_id | Sort-Object | Get-Unique
+
+if ($uniqueMessageIDs.count -eq 0) {
+    Write-Error "No Unique MessageIDs found in data."
```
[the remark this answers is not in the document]
Canthv0: Good catch ... moved all Write-Errors to -ErrorAction Stop

Transport/Measure-EmailDelayInMTL.ps1
```diff
@@ -0,0 +1,218 @@
...
+
+    # Build report object
+    $report = [PSCustomObject]@{
+        MessageID      = $id
+        TimeSent       = $TimeSent
+        TimeReceived   = $SortedTimeSent
+        MessageDelay   = $SortedTimeDelivered - $SortedTimeSent
+
+        # Build output object
+        [array]$output = [array]$output + $report
```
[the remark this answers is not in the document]
Canthv0: Totally reasonable ... updated.

Transport/Measure-EmailDelayInMTL.ps1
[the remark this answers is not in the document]
Canthv0: Documentation updated.

docs/Transport/Measure-EmailDelayInMTL.md
Canthv0: Documentation updated.

Canthv0:
Closing Pull request without Merge to spawn new on that is Squashed.

PR #2174, opened 2024-08-20
Move Update-PublicFolderPermissions.ps1 into CSS-Exchange

bill-long:
We're moving this script into CSS-Exchange. Highlights:

* New -PropagateAll switch.
* Tons of -Verbose output.
* Supports -WhatIf.

I plan to immediately open a couple of issues for future work. There's an opportunity for performance improvement, and it would be nice to replace the progress log with something that contains more than just entry IDs. But this is how the script has worked for years, so that stuff can wait.


docs/PublicFolders/Update-PublicFolderPermissions.md
```diff
@@ -0,0 +1,65 @@
+# ValidateExoPfDumpster
```
dpaulson45, line 1: Wrong name

dpaulson45 (changes requested):

dpaulson45 (approved):

lusassl-msft (approved):

PR #2199, opened 2024-09-30
[MDO] Add a new tool to resend failed emails 

iserrano76:
**Issue:**
Users can have legitimate emails blocked for various reasons, and once the cause is mitigated, they need to resend the affected emails.

**Reason:**
Make live easier to admins in EXO in case of unexpected blocked emails.

**Fix:**
Create a script to resend emails in failure state.
You can find additional information in the documentation.
https://github.com/iserrano76/CSS-Exchange/blob/MDO/docs/M365/MDO/ResendFailedMail.md

**Validation:**
Test it in a lab tenant. 
you can find screenshots in the documentation,.


M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,330 @@
...
+                    #Send-MgUserMail -UserId $failedMessage.SenderAddress -Message $fullMessage -ErrorAction Stop
+                    Send-MgUserMessage -UserId $failedMessage.SenderAddress -MessageId $fullMessage.Id -ErrorAction Stop
+                    Write-Host "Resent Message: $($failedMessage.MessageId) - Subject: $($fullMessage.Subject)"
+                } catch {
+                    Write-Host "Error resending message $($failedMessage.MessageId) - Subject: $($fullMessage.Subject)" -ForegroundColor Red
+                    Write-Verbose "$_"
+                }
+            }
+        }
+    }
```
dpaulson45, line 327: At the end, we should provide the count of the successful and failed to resends.

docs/M365/MDO/ResendFailedMail.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+## Parameters - all parameters are optional
+
+Parameter | Description |
+----------|-------------|
+Sender | Filter emails based on the sender's address.
+Recipient | Filter emails based on the recipient's address.
+Subject | Filter emails based on the email Subject.
+MessageId | Filter emails based on the MessageId address. **You must put the MessageId in double quotes**
```
dpaulson45, line 124: Why does the `MessageId` need to be in double quotes?

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,330 @@
...
+            Write-Verbose "Scopes $expectedScope is present."
+        } else {
+            Write-Host "The following scope is missing: $expectedScope" -ForegroundColor Red
+            $validScope = $false
+        }
+    }
+    return $validScope
+}
+
+if (-not $SkipConnectionCheck) {
```
dpaulson45, line 155: What is the scenario that we should support were we shouldn't validate this?

dpaulson45 (changes requested):

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
+function Write-HostLog ($message) {
+    if (![string]::IsNullOrEmpty($message)) {
+        $Script:HostLogger = $Script:HostLogger | Write-LoggerInstance $message
+    }
+    # all write-host should be logged in the debug log as well.
+    Write-DebugLog $message
+}
```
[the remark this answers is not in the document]
iserrano76: this is new, right? 
I will take a look to use it, it looks simplier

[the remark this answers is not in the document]
dpaulson45: A little new, created it about 7 months ago.

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,330 @@
...
+        try {
+            $tempAddress = Get-EXORecipient $failedMessage.SenderAddress -ErrorAction Stop
+            if ($null -eq $tempAddress) {
+                Write-Host "Sender $($failedMessage.RecipientAddress) is not a recipient in this tenant." -ForegroundColor Yellow
+                Write-Host "Discarded $($failedMessage.MessageId) - Subject: $($failedMessage.Subject)" -ForegroundColor Yellow
+            } else {
+                Write-Verbose "Added to cache Recipient $($failedMessage.SenderAddress) with Id $($failedMessage.SenderAddress)"
+                $recipientCache[$failedMessage.SenderAddress] = $failedMessage.SenderAddress
+                Write-Verbose "Verified $($failedMessage.SenderAddress)"
+                $verifiedAcceptedSenderMessages += $failedMessage
```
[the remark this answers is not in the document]
dpaulson45: If verifying the sender is time consuming, wouldn't that be true if we also fail to find it? I am talking about if we failed to find the sender `$null -eq $tempAddress`, why don't we cache that as well?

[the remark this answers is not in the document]
iserrano76: oh, ok , I get your point. 
I will check it

docs/M365/MDO/ResendFailedMail.md
```diff
@@ -0,0 +1,130 @@
...
+      - When created, the API permissions should look like this:<br>
+
+        !['No Logical inconsistencies found'](img/API-perms.png)
+
+    - Create a new client secret for the app under `Manage | Certificates & secrets`.
+
+> [!IMPORTANT]
+> Save the Value field of the secret **immediately** after creating it; you can't retrieve it later.
+
+> [!TIP]
```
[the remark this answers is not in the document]
rosspa05: Ok, perfect. We will change it as I think it looks better too.

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,330 @@
...
+        if (Test-GraphContext -Scopes $graphConnection.Scopes -ExpectedScopes $expectedScopes) {
+            Write-Host "Connected to Graph"
+            Write-Host "Session details"
+            Write-Host "TenantID: $(($graphConnection).TenantId)"
+            Write-Host "AuthType: $(($graphConnection).AuthType)"
+        } else {
+            Write-Host "We cannot continue without Graph Powershell session without Expected Scopes." -ForegroundColor Red
+            Write-Host "Verify that you have Microsoft.Graph.Mail and Microsoft.Graph.Users.Actions modules installed and loaded." -ForegroundColor Yellow
+            Write-Host "You could use:" -ForegroundColor Yellow
+            Write-Host "`tConnect-MgGraph -TenantId ""[YOUR TENANT ID HERE]"" -ClientSecretCredential (New-Object -TypeName System.Management.Automation.PSCredential -ArgumentList ""[YOUR APP ID HERE]"", (ConvertTo-SecureString -String ""[VALUE FIELD OF YOUR SECRET HERE]"" -AsPlainText -Force)) -NoWelcome" -ForegroundColor Yellow
```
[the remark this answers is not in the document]
rosspa05: We will follow your suggestion on this and change it.

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,330 @@
...
+$failedMessages = $null
+$failedMessages = Get-MessageTrace -StartDate (Get-Date).AddDays(-$Days) -EndDate (Get-Date) | Where-Object { $_.Status -eq "Failed" }
+
+if ($Sender) { $failedMessages = $failedMessages | Where-Object { $Sender -contains $_.SenderAddress } }
+if ($Subject) { $failedMessages = $failedMessages | Where-Object { $Subject -contains $_.Subject } }
+if ($Recipient) { $failedMessages = $failedMessages | Where-Object { $Recipient -contains $_.RecipientAddress } }
+if ($MessageId) { $failedMessages = $failedMessages | Where-Object { $MessageId -contains $_.MessageId } }
+if (-not $IncludeDuplicates) { $failedMessages = $failedMessages | Sort-Object MessageId -Unique }
+
+$failedMessages = $failedMessages | Where-Object { $acceptedDomainList -contains $_.SenderAddress.Split("@")[1] }
```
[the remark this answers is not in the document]
rosspa05: We'll take a look.

M365/MDO/ResendFailedMail.ps1
rosspa05: Ok, will do.

docs/M365/MDO/ResendFailedMail.md
rosspa05: This is documented and I have confirmed it tests: 
https://learn.microsoft.com/en-us/powershell/module/exchange/get-messagetrace?view=exchange-ps#-messageid 

image

M365/MDO/ResendFailedMail.ps1
rosspa05: The customer might not want or need to check the connection every time they run it in order to save time. If they run it once to see which emails would be sent, for example, why should they need to do the connection check again immediately afterwards.

docs/M365/MDO/ResendFailedMail.md
dpaulson45: That is for the cmdlet itself, which you don't provide the `MessageID` parameter when running the command `Get-MessageTrace`. You are just filtering it afterwards and it isn't required, as you defined it as a string array already, so it converts it to a string, which is all the quotation marks are doing. 

image

M365/MDO/ResendFailedMail.ps1
dpaulson45: But there isn't a valid reason to assume that is done within a reasonable time or even in the same PowerShell session.

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,320 @@
...
+    [Parameter(Mandatory = $false, ParameterSetName = "ResendCopyFailedDays")]
+    [switch]$SkipVersionCheck,
+    [Parameter(Mandatory = $true, ParameterSetName = "ScriptUpdateOnly")]
+    [switch]$ScriptUpdateOnly
+)
+
+$Script:DualLoggingEnabled = $true
+. $PSScriptRoot\..\..\Shared\GenericScriptStartLogging.ps1
+
+$versionsUrl = "https://aka.ms/ResendFailedMail-VersionsURL"
```
dpaulson45, line 96: `$versionsUrl` isn't used.

dpaulson45 (changes requested):

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,322 @@
...
+    Status = "Failed"
+}
+
+if ($StartDate) { $traceParams["StartDate"] = $StartDate }
+if ($EndDate) { $traceParams["EndDate"] = $EndDate }
+if ($RecipientAddress ) { $traceParams["RecipientAddress"] = $RecipientAddress }
+if ($SenderAddress) { $traceParams["SenderAddress"] = $SenderAddress }
+if ($MessageId) { $traceParams["MessageId"] = $MessageId }
+
+$failedMessages = New-Object System.Collections.Generic.List[object]
```
dpaulson45, line 221: For all the `New-Object System.Collections.Generic.List[object]` you are creating a list, but then doing the action of `+=` to attempt to add to the list. However, this process in PowerShell will convert the variable to an `Array` which is a static list vs a dynamic one.

You actually only need to create a single list in the 3 locations that you have attempted to use this.

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,322 @@
...
+
+try {
+    $failedMessages += Get-MessageTrace @traceParams
```
dpaulson45, line 224:
```suggestion


try {
    [array]$failedMessages = Get-MessageTrace @traceParams
```

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,322 @@
...
+
+if ($failedMessages.count -ge 1000) {
+    Write-Host "We get more than 1000 messages, please limit your search." -ForegroundColor Red
+    exit
+}
+
+if (-not $IncludeDuplicates) {
+    $uniqueMessages = New-Object System.Collections.Generic.List[object]
+    $uniqueMessages += $failedMessages | Sort-Object MessageId -Unique
+    $failedMessages = $uniqueMessages
```
dpaulson45, line 238:
```suggestion

    [array]$failedMessages = $failedMessages | Sort-Object MessageId -Unique

```

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,322 @@
...
+$totalMessages = $failedMessages.Count
+foreach ($failedMessage in $failedMessages) {
+    $count++
+    Write-Progress -Activity "Checking Progress" -Status "$count of $totalMessages" -PercentComplete ($count / $totalMessages * 100) -CurrentOperation "Checking message $($failedMessage.MessageId) - Subject: $($failedMessage.Subject)"
+    Write-Verbose "Checking $($failedMessage.SenderAddress)"
+    $tempAddress = $null
+    if ($recipientCache.ContainsKey($failedMessage.SenderAddress)) {
+        Write-Verbose "RecipientAddress  $($failedMessage.SenderAddress) found in cache"
+        if ($recipientCache[$failedMessage.SenderAddress]) {
+            $verifiedAcceptedSenderMessages += $failedMessage
```
dpaulson45, line 252:
```suggestion
            $verifiedAcceptedSenderMessages.Add($failedMessage)
```

dpaulson45: This is the correct way to add to the dynamic list.

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,322 @@
...
+            $tempAddress = Get-EXOMailbox $failedMessage.SenderAddress -ErrorAction Stop
+            if ($null -eq $tempAddress) {
+                Write-Verbose "Sender $($failedMessage.SenderAddress) is not a recipient in this tenant."
+                Write-Verbose "Discarded $($failedMessage.MessageId) - Subject: $($failedMessage.Subject)"
+                $recipientCache[$failedMessage.SenderAddress] = $false
+            } else {
+                Write-Verbose "Added to cache Recipient $($failedMessage.SenderAddress) with Id $($failedMessage.SenderAddress)"
+                $recipientCache[$failedMessage.SenderAddress] = $true
+                Write-Verbose "Verified $($failedMessage.SenderAddress)"
+                $verifiedAcceptedSenderMessages += $failedMessage
```
dpaulson45, line 268:
```suggestion
                $verifiedAcceptedSenderMessages.Add($failedMessage)
```

M365/MDO/ResendFailedMail.ps1
dpaulson45, line 224: We should probably have `-ErrorAction Stop` here.

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,322 @@
...
+        $resendCount = 0
+        foreach ( $failedMessage in $verifiedAcceptedSenderMessages ) {
+            $count++
+            Write-Progress -Activity "Resending Progress" -Status "$count of $totalMessages" -PercentComplete ($count / $totalMessages * 100) -CurrentOperation "Resending message $($failedMessage.MessageId) - Subject: $($failedMessage.Subject)"
+            try {
+                $fullMessage = $null
+                $fullMessage = Get-MgUserMessage -UserId $failedMessage.SenderAddress -Filter "InternetMessageId eq '$($failedMessage.MessageId)'" -ExpandProperty Attachments -ErrorAction Stop | Sort-Object ReceivedDateTime | Select-Object -First 1
+            } catch {
+                Write-Host "Error getting message $($failedMessage.MessageId) - Subject: $($failedMessage.Subject)" -ForegroundColor Red
+                Write-Verbose "$_"
```
dpaulson45, line 298: We should add a `continue` here, as we failed to get the message. Move onto the next message in the loop.

M365/MDO/ResendFailedMail.ps1
```diff
@@ -0,0 +1,322 @@
...
+                Write-Host "Message not found for $($failedMessage.MessageId)" -ForegroundColor Yellow
+            } else {
+                Write-Verbose "Resending message $($failedMessage.MessageId) - Subject: $($fullMessage.Subject)"
+                try {
+                    Send-MgUserMessage -UserId $failedMessage.SenderAddress -MessageId $fullMessage.Id -ErrorAction Stop
+                    Write-Host "Resent Message: $($failedMessage.MessageId) - Subject: $($fullMessage.Subject)"
+                    $resendCount++
+                } catch {
+                    Write-Host "Error resending message $($failedMessage.MessageId) - Subject: $($fullMessage.Subject)" -ForegroundColor Red
+                    Write-Verbose "$_"
```
dpaulson45, line 310: Because we are already at the end of the loop, you don't need to have a `continue` here.

M365/MDO/ResendFailedMail.ps1
dpaulson45: Not required, but something to think about to quickly move to the next item in the loop.

M365/MDO/ResendFailedMail.ps1
dpaulson45: @iserrano76 need to know what happens if the `SkipConnectionCheck` is used, when we aren't connected to EXO.

docs/M365/MDO/ResendFailedMail.md
iserrano76: this is about double quote for message ids:
It fails without quotes and works with double quotes:
image

Not only in the script in the original cmdlet as well:
image

dpaulson45: Didn't know that PowerShell treats that differently. Good to know.

dpaulson45 (approved):

PR #2202, opened 2024-10-07
Fix Script Version Bug

Shanefe:
Version bug fix
Link to Public Docs in MD
Update RBA script based on feedback

### DIFF
--- a/docs/Calendar/Get-CalendarDiagnosticObjectsSummary.md
+++ b/docs/Calendar/Get-CalendarDiagnosticObjectsSummary.md
@@ -51,4 +51,4 @@ Will create file like  `.\123456_CalLogSummary_<MeetingID>.xlsx` in current dire
 
 More Documentation on collecting CalLogs and Analyzing them:
  [How to Get Calendar Logs](https://aka.ms/GetCalLogs)
- [How to Analyse Calendar Logs](https://aka.ms/AnalyzeCalLogs)
+ [How to Analyze Calendar Logs](https://aka.ms/AnalyzeCalLogs)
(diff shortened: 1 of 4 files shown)


docs/Calendar/Get-CalendarDiagnosticObjectsSummary.md
```diff
@@ -48,3 +48,7 @@ Get-CalendarDiagnosticObjectsSummary.ps1 -Identity $Users -MeetingID $MeetingID
 ```
 Will create file like  `.\123456_CalLogSummary_<MeetingID>.xlsx` in current directory.
 
+
+More Documentation on collecting CalLogs and Analyzing them:
+https://learn.microsoft.com/en-us/exchange/troubleshoot/calendars/get-calendar-diagnostic-logs 
+https://learn.microsoft.com/en-us/exchange/troubleshoot/calendars/analyze-calendar-diagnostic-logs
```
dpaulson45, line 54: Recommend to put in aka.ms links here. 

I would also recommend to put in an aka.ms link on the main doc for the download. You can point it directly to the download, or you can point it to the docs page on mkdocs. But if you go directly to the download, then I would create another link for the mkdocs page to reference for more information. 

https://learn.microsoft.com/en-us/exchange/troubleshoot/calendars/get-calendar-diagnostic-logs#use-the-get-calendardiagnosticobjectssummaryps1-script

dpaulson45 (changes requested):
Request small changes for aka.ms links in docs. This way the page can move, and the aka.ms link just needs to be updated and nothing needs to occur here.

dpaulson45 (commented):
Test Comment for Build Pipeline

docs/Calendar/Get-CalendarDiagnosticObjectsSummary.md
dpaulson45: Still have the old links here as well.

dpaulson45 (approved):

PR #2206, opened 2024-10-16
New script to retrieve folder statistics for mailboxes with large numbers of folders

cboonham:
New script to retrieve folder statistics for mailboxes with large numbers of folders, working around potential timeouts.

**Issue:**
Customers experience timeouts when running Get-MailboxFolderStatistics against large mailboxes and raise support tickets.

**Reason:**
Introducing a new store script to use the -ResultSize and -SkipCount parameters to retrieve the folder statistics in batches.

**Fix:**
New script.

**Validation:**
Tested successfully on a number of customer escalations.

### DIFF
--- a/M365/Get-LargeMailboxFolderStatistics.ps1
+++ b/M365/Get-LargeMailboxFolderStatistics.ps1
@@ -42,7 +42,7 @@ param(
 )
 
 process {
-    $allContentFolders = New-Object System.Collections.Generic.List[System.Management.Automation.PSCustomObject]
+    $allContentFolders = New-Object System.Collections.Generic.List[object]
     $start = Get-Date
     Write-Host "$start Running Get-MailboxFolderStatistics for $Identity $MailboxType locations, in batches of $BatchSize"
 
@@ -54,7 +54,10 @@ process {
             do {
                 $skipCount = $BatchSize * $loopCount
                 $batch = Get-MailboxFolderStatistics -Identity $($location.Identity) -ResultSize $batchSize -SkipCount $skipCount
-                $allContentFolders += $batch | Where-Object { $_.ContentFolder -eq "TRUE" } | Select-Object -Property $Properties
+                [System.Array]$contentFolders = $batch | Where-Object { $_.ContentFolder -eq "TRUE" } | Select-Object -Property $Properties
+                if ($contentFolders.Count -gt 0) {
+                    $allContentFolders.AddRange($contentFolders)
+                }
                 Write-Host "$(Get-Date):$loopCount Found $($batch.Count) content folders from $($location.MailboxLocationType):$($location.MailboxGuid)"
                 $loopCount += 1
             }
--- a/M365/Get-LargeMailboxFolderStatistics.ps1
+++ b/M365/Get-LargeMailboxFolderStatistics.ps1
@@ -41,7 +41,7 @@ param(
     [string[]]$Properties = @("Name", "FolderPath", "ItemsInFolder", "FolderSize", "FolderAndSubfolderSize")
 )
 
-Process {
+process {
     $allContentFolders = New-Object System.Collections.Generic.List[System.Management.Automation.PSCustomObject]
     $start = Get-Date
     Write-Host "$start Running Get-MailboxFolderStatistics for $Identity $MailboxType locations, in batches of $BatchSize"
--- a/M365/Get-LargeMailboxFolderStatistics.ps1
+++ b/M365/Get-LargeMailboxFolderStatistics.ps1
@@ -1,6 +1,8 @@
 ﻿# Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+#Requires -Modules @{ ModuleName="ExchangeOnlineManagement"; RequiredVersion="3.4.0" }
+
 <#
 .SYNOPSIS
     Retrieves folder statistics for mailboxes with large numbers of folders.
@@ -25,7 +27,7 @@
 .EXAMPLE
     $folderStats = .\Get-LargeMailboxFolderStatistics.ps1 -Identity fred@contoso.com
     $folderStats = .\Get-LargeMailboxFolderStatistics.ps1 -Identity fred@contoso.com -MailboxType Primary
-    $folderStats = .\Get-LargeMailboxFolderStatistics.ps1 -Identity fred@contoso.com -MailboxType Archive -BatchSize 5000 -Properties "Name, FolderPath"
+    $folderStats = .\Get-LargeMailboxFolderStatistics.ps1 -Identity fred@contoso.com -MailboxType Archive -BatchSize 5000 -Properties @("Name", "FolderPath")
 #>
 param(
     [Parameter(Mandatory = $true, Position = 0)]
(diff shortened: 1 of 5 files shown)


Shanefe (approved):
:shipit:

Store/Get-LargeMailboxFolderStatistics.ps1
```diff
@@ -0,0 +1,68 @@
...
+    $MailboxType = "Archive",
+    [Parameter(Mandatory = $false, Position = 2)]
+    $BatchSize = 5000,
+    [Parameter(Mandatory = $false, Position = 3)]
+    $Properties = "Name, FolderPath, ItemsInFolder, FolderSize, FolderAndSubfolderSize"
+)
+
+Process {
+    $allContentFolders = $null
+    $propertyArray = $Properties -split ',' | ForEach-Object { $_.Trim() }
```
dpaulson45, line 44: Would be best to have the parameter handle this already and accept it as a string array instead of doing this.

Store/Get-LargeMailboxFolderStatistics.ps1
```diff
@@ -0,0 +1,68 @@
...
+param(
+    [Parameter(Mandatory = $true, Position = 0)]
+    $Identity,
+    [Parameter(Mandatory = $false, Position = 1)]
+    [ValidateSet("Primary", "Archive")]
+    $MailboxType = "Archive",
+    [Parameter(Mandatory = $false, Position = 2)]
+    $BatchSize = 5000,
+    [Parameter(Mandatory = $false, Position = 3)]
+    $Properties = "Name, FolderPath, ItemsInFolder, FolderSize, FolderAndSubfolderSize"
```
dpaulson45, line 39:
```suggestion
    [string[]]$Properties = @("Name", "FolderPath", "ItemsInFolder", "FolderSize", "FolderAndSubfolderSize")
```

Store/Get-LargeMailboxFolderStatistics.ps1
```diff
@@ -0,0 +1,68 @@
+﻿# Copyright (c) Microsoft Corporation.
```
dpaulson45, line 1: Let's avoid making a new folder structure here. Since this script is geared towards O365/M365 we should have it in that folder instead.

You can just place it at the root of that folder, and we will change the structure of that folder later. That should not be done within this PR.

Store/Get-LargeMailboxFolderStatistics.ps1
```diff
@@ -0,0 +1,68 @@
...
+# Licensed under the MIT License.
+
```
dpaulson45, line 3: This might be a good location to make sure that the EXO Module is installed with the correct version.

Store/Get-LargeMailboxFolderStatistics.ps1
```diff
@@ -0,0 +1,68 @@
...
+    $mailboxLocations = Get-MailboxLocation -User $Identity
+
+    foreach ($location in $mailboxLocations) {
+        if (($location.MailboxLocationType -like '*Archive' -and $MailboxType -like 'Archive') -or ($location.MailboxLocationType -like '*Primary' -and $MailboxType -like 'Primary')) {
+            $loopCount = 0
+            do {
+                $skipCount = $BatchSize * $loopCount
+                $batch = Get-MailboxFolderStatistics -Identity $($location.Identity) -ResultSize $batchSize -SkipCount $skipCount
+                $allContentFolders += $batch | Where-Object { $_.ContentFolder -eq "TRUE" } | Select-Object -Property $propertyArray
```
dpaulson45, line 56: Because you can be dealing with a large number of folders/objects being added to this list, it is recommended for the better performance to use `System.Collections.Generic.List[object]` for this collection.

dpaulson45 (changes requested):
You also need to update the `mkdocs.yml` file to include your new script doc file.

Store/Get-LargeMailboxFolderStatistics.ps1
cboonham: Updated

Store/Get-LargeMailboxFolderStatistics.ps1
cboonham: Done

Store/Get-LargeMailboxFolderStatistics.ps1
cboonham: Done

Store/Get-LargeMailboxFolderStatistics.ps1
dpaulson45: Yeah, we are trying to avoid adding too many components

dpaulson45:
@cboonham please address the following:

image


You can run this script manually as well on your end to know that everything is good prior to doing the commit.

cboonham:
@dpaulson45 Thanks for the review, CodeFormatter issues should be fixed now.

M365/Get-LargeMailboxFolderStatistics.ps1
```diff
@@ -0,0 +1,69 @@
...
+
+    $mailboxLocations = Get-MailboxLocation -User $Identity
+
+    foreach ($location in $mailboxLocations) {
+        if (($location.MailboxLocationType -like '*Archive' -and $MailboxType -like 'Archive') -or ($location.MailboxLocationType -like '*Primary' -and $MailboxType -like 'Primary')) {
+            $loopCount = 0
+            do {
+                $skipCount = $BatchSize * $loopCount
+                $batch = Get-MailboxFolderStatistics -Identity $($location.Identity) -ResultSize $batchSize -SkipCount $skipCount
+                $allContentFolders += $batch | Where-Object { $_.ContentFolder -eq "TRUE" } | Select-Object -Property $Properties
```
dpaulson45, line 57: Because of the way PowerShell works, this is going to change the type of `$allContentFolders` back to a generic array type. You should do something like this: 

```suggestion
                $batch = Get-MailboxFolderStatistics -Identity $($location.Identity) -ResultSize $batchSize -SkipCount $skipCount | 
                Where-Object { $_.ContentFolder -eq "TRUE" } | 
                Select-Object -Property $Properties
                $allContentFolders.AddRange($batch)
```

https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.addrange?view=net-8.0

M365/Get-LargeMailboxFolderStatistics.ps1
```diff
@@ -0,0 +1,69 @@
...
+    [ValidateSet("Primary", "Archive")]
+    $MailboxType = "Archive",
+    [Parameter(Mandatory = $false, Position = 2)]
+    $BatchSize = 5000,
+    [Parameter(Mandatory = $false, Position = 3)]
+    [string[]]$Properties = @("Name", "FolderPath", "ItemsInFolder", "FolderSize", "FolderAndSubfolderSize")
+)
+
+process {
+    $allContentFolders = New-Object System.Collections.Generic.List[System.Management.Automation.PSCustomObject]
```
dpaulson45, line 45:
```suggestion
    $allContentFolders = New-Object System.Collections.Generic.List[object]
```

Needs to change, if you are going to do `AddRange()` for the better performance.

M365/Get-LargeMailboxFolderStatistics.ps1
cboonham: I had to play around with this to get it to work in all scenarios. I found the cast to [System.Array] was necessary to force single value results to an array and also AddRange on arrays with no elements also threw an error hence the check there before the AddRange. If I get a chance I will try it on a large archive to see if it actually makes a difference.

M365/Get-LargeMailboxFolderStatistics.ps1
cboonham: Done, see above.

dpaulson45 (approved):

PR #2218, opened 2024-11-06
[MDO] Fix empty subject by sorting on the first to last  message and get first one (oldest should be the original)

iserrano76:
**Issue:**
If you have more that one copy in message trace not always shows subject

**Reason:**
With the current filter we do not take always the original one, the oldest one.

**Fix:**
Fix empty subject by sorting on the first to last  message and get first one (oldest should be the original and has the original subject)

**Validation:**
Check it in lab

### DIFF
--- a/M365/MDO/ResendFailedMail.ps1
+++ b/M365/MDO/ResendFailedMail.ps1
@@ -231,7 +231,7 @@ if ($failedMessages.count -ge 1000) {
 }
 
 if (-not $IncludeDuplicates) {
-    [array]$failedMessages = $failedMessages | Sort-Object MessageId -Unique
+    [array]$failedMessages = $failedMessages | Group-Object -Property MessageId | ForEach-Object { $_.Group | Sort-Object -Property Received | Select-Object -First 1 }
 }
 
 $verifiedAcceptedSenderMessages = New-Object System.Collections.Generic.List[object]


rosspa05 (approved):
Tested, reviewed, and discussed. Looks good.

rosspa05:
Discussed, tested, and reviewed. Looks better.

M365/MDO/ResendFailedMail.ps1
```diff
@@ -231,7 +231,7 @@ if ($failedMessages.count -ge 1000) {
 }
 
 if (-not $IncludeDuplicates) {
-    [array]$failedMessages = $failedMessages | Sort-Object MessageId -Unique
+    [array]$failedMessages = $failedMessages | Group-Object -Property MessageId | ForEach-Object { $_.Group | Sort-Object -Property Received | Select-Object -First 1 }
```
dpaulson45, line 234: How does this address empty subjects?

iserrano76: It looks an issue with Get-MessageTrace:
for all original messages it is able to get subject, but for some resend messages is not able.
With this we will get the oldest one, the original with that messageId

dpaulson45: Would it be better to include a where statement when the subject isn't null?

dpaulson45:
Looks like the build pipeline is having an issue with checking out your branch. Not sure if this is due to the close or not.

M365/MDO/ResendFailedMail.ps1
iserrano76: I though that but if the original message is null, it will not work

iserrano76:
> Looks like the build pipeline is having an issue with checking out your branch. Not sure if this is due to the close or not.
I try to run it again but it fail.
should we created it again?

M365/MDO/ResendFailedMail.ps1
dpaulson45: We talking about original message subject or original message MessageId

dpaulson45:
Got the pipeline to work after the merge.

PR #2227, opened 2024-11-21
New scripts to help manage mailbox extended property quota issues

cboonham:
**Issue:**
Mailboxes can hit extended property (nee named properties) quota issue, due to apps registering too many extended properties.

**Reason:**
These scripts help diagnose if there is an extended properties quota issue and help to address the issue by searching for a removing the unwanted extended properties.

**Fix:**
New scripts and documentation.

**Validation:**
Tested in own tenant.


M365/Check-MailboxExtendedProperty.ps1
```diff
@@ -0,0 +1,133 @@
...
+    .\Check-MailboxExtendedProperty.ps1 -Identity fred@contoso.com
+    .\Check-MailboxExtendedProperty.ps1 -Identity fred@contoso.com -Threshold 0.95
+    .\Check-MailboxExtendedProperty.ps1 -Identity fred@contoso.com -Threshold 0.7 -SelectFirst 20
+
+#>
+param(
+    [Parameter(Mandatory = $true, Position = 0)]
+    $Identity,
+    [Parameter(Mandatory = $false, Position = 1)]
+    $Threshold = 0.9,
```
dpaulson45, line 32: Let's have a `ValidateScript` here to make sure the value is between `.5` and less than `1`. Or whatever that makes sense here. But we don't want to have an end user provide something like `90` when they really wanted `.9`.

M365/Check-MailboxExtendedProperty.ps1
```diff
@@ -0,0 +1,133 @@
+﻿# Copyright (c) Microsoft Corporation.
```
dpaulson45, line 1: We should probably change this script name to `Test-MailboxExtendedProperty`. We tend to follow the guidelines of approved verbs for PowerShell, at least for the public name of the scripts/functions.  

https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.4

M365/Remove-MailboxExtendedProperty.ps1
```diff
@@ -0,0 +1,62 @@
...
+    A list of mailbox items and their extended property, that are to be removed.
+
+.EXAMPLE
+    $mailboxExtendedProperty = Get-MailboxExtendedProperty -Identity fred@contoso.com | Where-Object { $_.PropertyName -like '*Some Pattern*' }
+    $messagesWithExtendedProperty = .\Search-MailboxExtendedProperty.ps1 -MailboxExtendedProperty $mailboxExtendedProperty
+    .\Remove-MailboxExtendedProperty.ps1 -MessagesWithExtendedProperty $messagesWithExtendedProperty
+
+#>
+param(
+    [Parameter(Mandatory = $true, Position = 0)]
```
dpaulson45, line 25: Because we are doing a remove here, we need to include a `ShouldProcess` within the script.

M365/Search-MailboxExtendedProperty.ps1
```diff
@@ -0,0 +1,138 @@
...
+.PARAMETER MailboxExtendedProperty
+    One of more mailbox extended properties to search for in the mailbox, returned by Get-MailboxExtendedProperty.
+
+.EXAMPLE
+    $mailboxExtendedProperty = Get-MailboxExtendedProperty -Identity fred@contoso.com | Where-Object { $_.PropertyName -like '*Some Pattern*' }
+    $messagesWithExtendedProperty = .\Search-MailboxExtendedProperty.ps1 -MailboxExtendedProperty $mailboxExtendedProperty
+#>
+param(
+    [Parameter(Mandatory = $true, Position = 0)]
+    $MailboxExtendedProperty
```
dpaulson45, line 27: After running `$mailboxExtendedProperty = Get-MailboxExtendedProperty -Identity fred@contoso.com | Where-Object { $_.PropertyName -like '*Some Pattern*' }` what is the type of `$mailboxExtendedProperty`? We should either include that type, if it is specific to a class as a parameter, or we should validate the information passed.

dpaulson45 (changes requested):
Quick review of the scripts, make those changes and will come back and go in deeper into it.

M365/Check-MailboxExtendedProperty.ps1
cboonham: Added parameter validation

M365/Remove-MailboxExtendedProperty.ps1
cboonham: Implemented SupportsShouldProcess

M365/Check-MailboxExtendedProperty.ps1
cboonham: Renamed to Test-

M365/Search-MailboxExtendedProperty.ps1
cboonham: Added parameter Type validation

dpaulson45:
Pipeline failed due to spell check: 

image

Must use camelCase or PascalCase

cboonham:
Fixed the other spellings, but 'readwrite' is in a Graph Scope, i.e. mail.readwrite, so need to stay.

cboonham:
> Pipeline failed due to spell check:
> 
> image
> 
> Must use camelCase or PascalCase

Done

dpaulson45:
Code formatting currently failing with the build process: 

image

Run the `CodeFormatter.ps1` in the `.build` directory

dpaulson45:
@lusassl-msft will review and test tomorrow

M365/Search-MailboxExtendedProperty.ps1
```diff
@@ -0,0 +1,145 @@
+﻿# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+#Requires -Modules @{ ModuleName="ExchangeOnlineManagement"; RequiredVersion="3.4.0" }
+#Requires -Modules @{ ModuleName="Microsoft.Graph.Users"; RequiredVersion="2.24.0" }
+#Requires -Modules @{ ModuleName="Microsoft.Graph.Mail"; RequiredVersion="2.24.0" }
```
lusassl-msft, line 6: @cboonham do we require exactly this version or is this the minimum version that is required? If it's the minimum version, it would be better to use `ModuleVersion` instead as this specifies a minimum acceptable version of the module.

M365/Remove-MailboxExtendedProperty.ps1
```diff
@@ -0,0 +1,72 @@
...
+            }
+        })]
+    $MessagesWithExtendedProperty
+)
+
+process {
+    # Get the current Microsoft Graph context
+    $context = Get-MgContext
+    if ($null -eq $context) {
+        Write-Host -ForegroundColor Red "No valid context. Please connect to Microsoft Graph first."
```
lusassl-msft, line 41: Do you want to write out a log file? You can use the shared `GenericScriptStartLogging.ps1` helper functions to write `Write-Host` and `Write-Verbose` out to a log file. This would be helpful in case of troubleshooting issues with the script.

M365/Remove-MailboxExtendedProperty.ps1
```diff
@@ -0,0 +1,72 @@
...
+                $true
+            } else {
+                throw "The parameter MailboxExtendedProperty doesn't appear to be the result from running 'Search-MailboxExtendedProperty'."
+            }
+        })]
+    $MessagesWithExtendedProperty
+)
+
+process {
+    # Get the current Microsoft Graph context
```
lusassl-msft, line 38: You can add the shared `GenericScriptUpdate.ps1` helper functions to the script. This will add auto-update capabilities in case that we release a newer version of the script.

dpaulson45: If we don't think the scripts are going to have a lot of repeat traffic, we might not need to have the update script feature. @cboonham I would leave that up to your discretion.

M365/Search-MailboxExtendedProperty.ps1
cboonham: OK, done. I was undecided whether to do this originally, as it means testing with later versions, but it should be fine.

M365/Remove-MailboxExtendedProperty.ps1
cboonham: I will consider this for any larger scripts that may need a log file. Doesn't this create download dependencies?

lusassl-msft: Nope this will not cause download dependencies. What we basically do is we override the, for example, `Write-Host` cmdlet with our own implementation which can then be used to display something and writing the same text out to a debug log file.

dpaulson45 (approved):

PR #2234, opened 2024-11-27
Improve hybrid transport certificate testing

### LINKED ISSUE #1931: TlsCertificateName not configured for cloud mail enabled connector - Health Checker
**Provide Version Number**
v23.12.08.2252

**Describe the issue**
Health Checker reports _TlsCertificateName_ is not configured for an Edge Transport Server-based hybrid send connector, and missing the tested internal Exchange mailbox server.

**Expected behavior**
The script should recognize the source transport servers as Edge transport servers, and skip any local TLS certificate checks as the certificate is not enabled for SMTP on internal mailbox servers.

**Script Output**
image

### LINKED ISSUE #1932: Hybrid Tls Certificate not found on server - Health Checker
**Provide Version Number**
v23.12.08.2252

**Describe the issue**
The script lists the configured TlsCertificate for the cloud-enabled send connector as missing on other servers besides the very first mailbox server. The certificate is installed on a single server for HCW reference, as the org uses Edge transport servers for hybrid SMTP connectivity.

**Expected behavior**
The script should ignore the missing certificate on other mailbox servers, as the certificate is only installed on one server for HCW reference.

**Script Output**
image

### LINKED ISSUE #2035: New ExoConnectorCheck vs. documentaion - Health Checker
**Provide Version Number**
24.02.15.1640

**Describe the issue**
The New ExoConnectionCheck reports an issue if a send connector is present that routes mail to Exchange Online when there is no TLS certificate explicitly configured on the connector.
The [documentation link](https://learn.microsoft.com/en-us/exchange/mail-flow-best-practices/use-connectors-to-configure-mail-flow/set-up-connectors-to-route-mail#2-set-up-your-email-server-to-relay-mail-to-the-internet-via-microsoft-365-or-office-365) referred does not mention explicitly setting the TLS certificate (at least not on the send connector), nor does the PowerShell command include the TlsCertificateName parameter.

**Expected behavior**
If it is now required to explicitly specify a TlsCertificateName on the send connector, documentation should be updated accordingly.
If not, this should not be reported as an issue (but additional checks may be required to verify correct configuration).

**Script Output**
	EXO Connector Present: True
		Send Connector - <connectorname>: Misconfigured to send authenticated internal mail to M365.
			CloudServicesMailEnabled: True
			TLSCertificateName set: False
		More Information: https://aka.ms/HC-ExoConnectorIssue

**Additional context**
Exchange automatically selects a suitable certificate that matches the FQDN on the send connector, as long as it is valid and enabled for SMTP.
So the check might instead (or additionally) involve verifying if such a certificate is present, and if it is issued by a public authority (e.g. not self-signed and not issued by a AD CA).

lusassl-msft:
**Issue:**
The transport certificate testing wasn't working accurately. HealthChecker, for example was looking for the hybrid transport certificate even if the computer was not configured as `SendingTransportServers` or `ReceivingTransportServers`. We also showed a warning if `TLSCertificateName` wasn't set on a send connector. This is a valid use-case as Exchange server allows implicit or explicit configuration. With an implicit configuration, Exchange Server looks for the best matching certificate is selected based on the Fqdn set on the connector.

Resolve #1932 
Resolve #1931 
Resolve #2035 

**Fix:**
Excluded the server from some of the checks related to TLS certificate if it doesn't exist in the `SendingTransportServers` or `ReceivingTransportServers` collection. Improved the TLSCertificateCheck to only show a red error if the `TLSCertificateName` is not set and the `Fqdn` is not set on the connector (computers Fqdn is used in this scenario and this could be an internal one).

**Validation:**
Lab

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
@@ -23,6 +23,12 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
     $tcpKeepAlive = $osInformation.RegistryValues.TCPKeepAlive
     $organizationInformation = $HealthServerObject.OrganizationInformation
 
+    # cSpell:disable
+    $eopDomainRegExPattern = "^[^.]+\.mail\.protection\.(outlook\.com|partner\.outlook\.cn|office365\.us)$"
+    $remoteRoutingDomainRegExPattern = "^[^.]+\.mail\.(onmicrosoft\.com|partner\.onmschina\.cn|onmicrosoft\.us)$"
+    $serviceDomainRegExPattern = "^mail\.protection\.(outlook\.com|partner\.outlook\.cn|office365\.us)$"
+    # cSpell:enable
+
     $baseParams = @{
         AnalyzedInformation = $AnalyzeResults
         DisplayGroupingKey  = (Get-DisplayResultsGroupingKey -Name "Frequent Configuration Issues"  -DisplayOrder $Order)
@@ -299,8 +305,8 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
     $sendConnectors = $exchangeInformation.ExchangeConnectors | Where-Object { $_.ConnectorType -eq "Send" }
 
     foreach ($sendConnector in $sendConnectors) {
-        $smartHostMatch = ($sendConnector.SmartHosts -match "^[^.]+\.mail\.protection\.(outlook\.com|partner\.outlook\.cn|office365\.us)$").Count -gt 0
-        $dnsMatch = $sendConnector.SmartHosts -eq 0 -and ($sendConnector.AddressSpaces.Address -match "^[^.]+\.mail\.(onmicrosoft\.com|partner\.onmschina\.cn|onmicrosoft\.us)$").Count -gt 0
+        $smartHostMatch = ($sendConnector.SmartHosts -match $eopDomainRegExPattern).Count -gt 0
+        $dnsMatch = $sendConnector.SmartHosts -eq 0 -and ($sendConnector.AddressSpaces.Address -match $remoteRoutingDomainRegExPattern).Count -gt 0
 
         if ($dnsMatch -or $smartHostMatch) {
             $exoConnector.Add($sendConnector)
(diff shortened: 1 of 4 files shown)


Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
```diff
@@ -299,8 +299,8 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
     $sendConnectors = $exchangeInformation.ExchangeConnectors | Where-Object { $_.ConnectorType -eq "Send" }
 
     foreach ($sendConnector in $sendConnectors) {
-        $smartHostMatch = ($sendConnector.SmartHosts -like "*.mail.protection.outlook.com").Count -gt 0
-        $dnsMatch = $sendConnector.SmartHosts -eq 0 -and ($sendConnector.AddressSpaces.Address -like "*.mail.onmicrosoft.com").Count -gt 0
+        $smartHostMatch = ($sendConnector.SmartHosts -match "^[^.]+\.mail\.protection\.(outlook\.com|partner\.outlook\.cn|office365\.us)$").Count -gt 0
+        $dnsMatch = $sendConnector.SmartHosts -eq 0 -and ($sendConnector.AddressSpaces.Address -match "^[^.]+\.mail\.(onmicrosoft\.com|partner\.onmschina\.cn|onmicrosoft\.us)$").Count -gt 0
```
dpaulson45, line 303: Since this is rare, instead of adding `onmschina` to the dictionary, can we just set the pattern to a variable prior where we manually disable the spell check, then enable it afterwards?

lusassl-msft: Sure, done with the next commit.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerFrequentConfigurationIssues.ps1
```diff
@@ -315,12 +321,17 @@ function Invoke-AnalyzerFrequentConfigurationIssues {
...
-        # Misconfigured connector is if TLSCertificateName is not set or CloudServicesMailEnabled not set to true
-        if ($connector.CloudEnabled -eq $false -or
-            $connector.CertificateDetails.TlsCertificateNameStatus -eq "TlsCertificateNameEmpty") {
+
+        # If CloudServiceMailEnabled is not set to true it means the connector is misconfigured
+        # If no Fqdn is set on the connector, the Fqdn of the computer is used to perform best matching certificate selection
+        # There is a risk that the Fqdn is an internal one (e.g., server.contoso.local) which will lead to a broken hybrid mail flow in case that TlsCertificateName is not set
+        if (($connector.CloudEnabled -eq $false) -or
+            ($null -eq $connector.Fqdn -and
+            $connector.CertificateDetails.TlsCertificateNameStatus -eq "TlsCertificateNameEmpty")) {
```
lusassl-msft, line 330: @richfaj @stanalek does that work to make the check more accurate? We show a red error in case that `CloudServiceMailEnabled` is `false` or if no `Fqdn` is set on the connector => `Fqdn` of the computer will be used which is quite often an internal one (e.g., computer.domain.local) and `TlsCertificateName` is also not set => implicit matching will be performed.

PR #2265, opened 2025-01-11
Allow trailing back-slashes in paths and use proper join-path/io class to handle paths 

michelderooij:
**Issue:**
For $XMLDirectoryPath, you check if the path does not end in a back-slash. For $OutputFilePath, you do the same as well as test the path. The description mentions the trailing "requirement" of one, but not the other - inconsistent. Test-Path for both suffices. 

**Reason:**
Accepting trailing slashes is convenient, as tab-completion adds trailing back-slash automatically. Also removes inconsistency.

**Fix:**
Use Test-Path for both in parameter validation. Used Join-Path etc to properly construct paths, and [System.IO.Path]::changeExtension instead of string operations.

**Validation:**
Tested against local and remote servers, using local and UNC paths

### DIFF
--- a/Diagnostics/HealthChecker/HealthChecker.ps1
+++ b/Diagnostics/HealthChecker/HealthChecker.ps1
@@ -27,9 +27,7 @@
 	This optional parameter allows the target Exchange server to be specified.  If it is not the
 	local server is assumed.
 .PARAMETER OutputFilePath
-	This optional parameter allows an output directory to be specified.  If it is not the local
-	directory is assumed.  This parameter must not end in a \.  To specify the folder "logs" on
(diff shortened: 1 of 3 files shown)


bill-long:
Thank you for the PR. Please make CodeFormatter happy:

image

Looks good to me otherwise. You can run CodeFormatter locally prior to submission as noted in [CONTRIBUTING.md](https://github.com/microsoft/CSS-Exchange/blob/main/CONTRIBUTING.md).

dpaulson45:
@michelderooij please address the code formatting rule as soon as you can.

dpaulson45 (approved):

PR #2272, opened 2025-01-21
Update Invoke-AnalyzerIISInformation.ps1 to verify Exchange Back End website binding [Issue #2209]

### LINKED ISSUE #2209: HealthChecker should validate IIS bindings for "All Unassigned"
Healthchecker should check if IIS bindings are correctly configured, especially on the backend website.
Binding should be "all unassigned" and not for a specific IP address.

**Describe The Request**
The correct default values on backend should be like this:
       Exchange Back End  Started  False         http  - *:81:  - NULL
                                                  https - *:444: - 69DE5CA602D04D03939DA4E2F5BE2D076609B906
Incorrect setting on backend:
        Exchange Back End  Started  False         http  - *:81:              - NULL
                                                  https - **192.168.10.48**:444: - 69DE5CA602D04D03939DA4E2F5BE2D076609B906

This becomes important when extended protection is enabled. When it is, and the binding is set to a specific IP address of the server the backend will drop the connection with the well known SEC_E_BAD_BNDINGS error; 2148074310.

**Additional context**
Checked with Exchange Health Checker version 24.10.15.1910

shasagar:
Modified the code to verify Exchange Backend Site binding and notify if website is bind with single IP.

**Issue:**
Healthchecker should check if IIS bindings are correctly configured, especially on the backend website.
Binding should be "all unassigned" and not for a specific IP address.

**Reason:**
The correct default values on backend should be like this:
Exchange Back End Started False http - *:81: - NULL
https - *:444: - 69DE5CA602D04D03939DA4E2F5BE2D076609B906
Incorrect setting on backend:
Exchange Back End Started False http - *:81: - NULL
https - 192.168.10.48:444: - 69DE5CA602D04D03939DA4E2F5BE2D076609B906

This becomes important when extended protection is enabled. When it is, and the binding is set to a specific IP address of the server the backend will drop the connection with the well known SEC_E_BAD_BNDINGS error; 2148074310.

**Fix:**
Updated "Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerIISInformation.ps1" to check Exchange Back End website binding
Resolved #2209 

**Validation:**
Performed test in Lab by removing 'All Unassigned' binding from Exchange Back End website and binding single ip address with port 444

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerIISInformation.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerIISInformation.ps1
@@ -78,6 +78,7 @@ function Invoke-AnalyzerIISInformation {
     Write-Verbose "Working on IIS Web Sites"
     $outputObjectDisplayValue = New-Object System.Collections.Generic.List[object]
     $problemCertList = New-Object System.Collections.Generic.List[string]
+    $exchangeBackEndBindingsList = New-Object System.Collections.Generic.List[string]
     $iisWebSites = $exchangeInformation.IISSettings.IISWebSite | Sort-Object ID
     $bindingsPropertyName = "Protocol - Bindings - Certificate"
 
@@ -107,6 +108,10 @@ function Invoke-AnalyzerIISInformation {
                     } elseif ($cert.LifetimeInDays -lt 0) {
                         $problemCertList.Add("'$certHash' Has expired and will cause problems.")
                     }
+                }
+
+                if ($website.name -eq "Exchange Back End" -and $($_.Protocol) -eq "https") {
+                    $exchangeBackEndBindingsList.Add($($_.bindingInformation))
                 }
 
                 1..(($protocolLength - $_.Protocol.Length) + 1) | ForEach-Object { $pSpace += " " }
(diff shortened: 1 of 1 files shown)


dpaulson45:
@shasagar let's make the commit name more appropriate, `Update Invoke-AnalyzerIISInformation.ps1` is very generic and isn't useful in the commit history.

Let's also include the issue number in the main PR comment to have it be linked.

Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerIISInformation.ps1
```diff
@@ -78,6 +78,7 @@ function Invoke-AnalyzerIISInformation {
     Write-Verbose "Working on IIS Web Sites"
     $outputObjectDisplayValue = New-Object System.Collections.Generic.List[object]
     $problemCertList = New-Object System.Collections.Generic.List[string]
+    $exchangeBackEndBindings = @()
```
lusassl-msft, line 81: Can we use a List<T> here instead of an array?

dpaulson45: Yeah, probably should being that the other lists above are that way. I was going to say, there is only going to be a few items at most in this list so there really isn't a performance benefit, but it will make more sense if we do it that way based off the other code.

dpaulson45 (approved):

dpaulson45:

PR #2280, opened 2025-02-10
Update Test-ExchAVExclusions.ps1 - Add AMSI detection

Ardasar:
**Issue:**
As you know, AMSI dll injection can occur in several ways in different antiviruses. By default, we believe that AMSI libraries should be injected using a supported approach (like AMSI providers, etc.).

However, many antiviruses make injections bypassing the standard approach.

The current version of the script does not allow engineers to determine whether the AMSI library was added as a legitimate provider or the antivirus simply scans the process (makes an injection) using the same library (very common).

**Reason**
When analyzing connectivity/performance issues, it is important to understand how exactly the antivirus is integrated into the w3wp processes (supported/unsupported way).

**Fix:**
The script will show the current injected AMSI dlls and providers settings on the server.
At the moment, the script only outputs the data to a file, but at the end of the script, several commented lines have been added that will allow the information to be output to the screen as well.

### DIFF
--- a/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
+++ b/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
@@ -1,4 +1,4 @@
-# Copyright (c) Microsoft Corporation.
+﻿# Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
 <#
--- a/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
+++ b/Diagnostics/AVTester/Test-ExchAVExclusions.ps1
@@ -8,6 +8,7 @@
     Major Release History:
         06/16/2021 - Initial Release
         06/26/2023 - Added ability to scan processes
+        02/10/2025 - Added AMSI detection
 
 .SYNOPSIS
 Uses EICAR files to verify that all Exchange paths that should be excluded from AV scanning are excluded.
@@ -313,6 +314,7 @@ $currentDiff = $initialDiff
 $firstExecution = $true
 $SuspiciousProcessList = New-Object Collections.Generic.List[string]
 $SuspiciousW3wpProcessList = New-Object Collections.Generic.List[string]
+$SuspiciousAMSIinW3wpProcessList = New-Object Collections.Generic.List[string]
 
 Write-Host "Analyzing Exchange Processes"
 while ($currentDiff -gt 0) {
@@ -437,6 +439,18 @@ while ($currentDiff -gt 0) {
         $FIPCompanyNameAllowList.Add("Oracle Corporation")
         $FIPCompanyNameAllowList.Add("Oracle Corp.")
 
+        #AMSI modules list
+        $AMSIModulesList = New-Object Collections.Generic.List[string]
+
+        #Microsoft Defender
+        $AMSIModulesList.add("MpOav.dll")
+
+        #McAfee/Trelix
+        $AMSIModulesList.add("MfeAmsiProvider.dll")
+
+        #Kaspersky
+        $AMSIModulesList.add("amsiprovider.dll")
+
         # cSpell:enable
 
         Write-Verbose "Allow List Module Count: $($ModuleAllowList.count)"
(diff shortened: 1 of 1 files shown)


Ardasar:
@microsoft-github-policy-service agree company="Microsoft"

dpaulson45:
@iserrano76 please review and let's determine if we should be displaying the information or not.

dpaulson45:
image

dpaulson45:
Yeah, I think this script should just call out anything that isn't out of the box default that is loaded into our process.

We should include information about the AMSI stuff to know if this is related to that from the registry, but I think that is all that we should be doing.

dpaulson45:
Personally, I don't like that we are including the 3rd party here. This will be a never-ending list, and it could provide false positive information saying that it is okay when it really isn't.

iserrano76:
> The Test-AMSI script is intended for other purposes.
> 
> The changes proposed in this request first of all are aimed at obtaining information about the AMSI configuration, which makes life easier for our engineers.

Correct me if I am wrong, when you collect AMSI information you took "HKLM:\SOFTWARE\Microsoft\AMSI\Providers" registry key and look for this information in HKLM:\SOFTWARE\Classes\CLSID\$_ to check the AMSI providers is in the correct place.

If this is the case it is what Test-AMSI is doing with the parameter -CheckAMSIConfig
image

Ardasar:
The main idea is to obtain a list of modules loaded via AMSI for subsequent display in the list of third-party modules loaded in w3wp. Similarly, it is intended to select third-party modules loaded in edgetransport that function as transport agents. While the implementation for transport agents is clear, with AMSI, I aim to gather as many examples as possible to write correctly functioning code. Additionally, running two scripts for the same purpose seems somewhat redundant.

Moreover, few people are aware of the Test-AMSI script, and it was originally created for other tasks. Attempting to rewrite it correctly would be akin to **duplicating** Test-ExchAVExclusions. On the other hand, engineers frequently use the Test-ExchAVExclusions script, and it is already ideally suited for determining the legitimacy of AMSI module loading.

iserrano76:
I think I found an alternative solution to avoid use a list of 3rd party modules.
If the AMSI provider is correctly implemented it should include the dll in the registrykey:
image
so we can get the dll and verify if w3wp has that dll.
I sent you and example.

dpaulson45:
> I think I found an alternative solution to avoid use a list of 3rd party modules. If the AMSI provider is correctly implemented it should include the dll in the registrykey: image so we can get the dll and verify if w3wp has that dll. I sent you and example.

We can provide that it is an AMSI loaded DLL and correctly configured, but we still should be calling it out as it is 3rd party and it can still cause possible issues within our product. I do like this approach better than have a list we maintain.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -313,6 +314,37 @@ $currentDiff = $initialDiff
...
+    $regex = "[0-9A-Fa-f\-]{36}"
+
+    # Filter the subKeys that match the regular expression and get only their names
+    $matchingSubKeys = $subKeys | Where-Object { $_.Name -match $regex } | ForEach-Object { $_.PSChildName }
+
+    if ($matchingSubKeys) {
+        $AMSIDll = $null
+        $AMSIDll = $matchingSubKeys | ForEach-Object { (Get-Item "HKLM:\SOFTWARE\Classes\ClSid\$_\InprocServer32" -ErrorAction SilentlyContinue).GetValue("") }
+        if ($AMSIDll) {
+            $AMSIDlls += $AMSIDll.trim('"')
```
dpaulson45, line 338: `$AMSIDlls` was created as a generic list, but you are using it as a static array causing it to switch. Please use `$AMSIDlls.Add(($AMSIDll.trim('"')))` instead.

dpaulson45 (changes requested):

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
iserrano76: if fact it is not need it, AMSIDll has the information need it

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -313,6 +314,34 @@ $currentDiff = $initialDiff
...
+    $regex = "[0-9A-Fa-f\-]{36}"
+
+    # Filter the subKeys that match the regular expression and get only their names
+    $matchingSubKeys = $subKeys | Where-Object { $_.Name -match $regex } | ForEach-Object { $_.PSChildName }
+
+    if ($matchingSubKeys) {
+        $AMSIDll = $null
+        $AMSIDll = $matchingSubKeys | ForEach-Object { (Get-Item "HKLM:\SOFTWARE\Classes\ClSid\$_\InprocServer32" -ErrorAction SilentlyContinue).GetValue("").trim('"') }
+        if ($null -eq $AMSIDll) {
+            Write-Host 'No AMSI Dlls was found for $subKeys, possible AMSI misconfiguration"' -ForegroundColor Red
```
dpaulson45, line 337: `$subKeys` appears to be a list, therefore, when converting to a string this isn't going to be useful if it is a list.

Diagnostics/AVTester/Test-ExchAVExclusions.ps1
```diff
@@ -313,6 +314,34 @@ $currentDiff = $initialDiff
...
+        }
+    } else {
+        Write-Host 'No AMSI configuration was found for $subKeys, possible AMSI misconfiguration"' -ForegroundColor Red
```
dpaulson45, line 340: Same thing here.

dpaulson45 (approved):

PR #2287, opened 2025-02-24
Update Emerging-Issues.md

Batre-MSFT:
**Issue:**

**Reason:**

**Fix:**
Short description of the fix

**Validation:**
Provide if applicable


Batre-MSFT:
Added line about Extended protection issue

dpaulson45 (changes requested):
Information is already there about CU14, twice actually, we don't need this information for a 3 time.

dpaulson45 (approved):

PR #2294, opened 2025-03-10
Added show detailed explanation parameter

rosspa05:
**Issue:**
Added show detailed explanation parameter. 
Added functionality to check for illogical inclusions or exclusions for Outbound spam policy, which was missing.
Corrected policy lookup based on rule name to look up policy based on the policy property in the rule. 

**Reason:**
Detailed explanation switch helps users understand output. 
Fixed missing functionality and made lookup more robust. 

**Fix:**
Added verification and show more details. 
Fixed above issues. 

**Validation:**
Tested in lab.

### DIFF
--- a/M365/MDO/MDOThreatPolicyChecker.ps1
+++ b/M365/MDO/MDOThreatPolicyChecker.ps1
@@ -107,6 +107,21 @@ begin {
     # Cache of members to reduce number of calls to Get-MgGroupMember
     $memberCache = @{}
 
+    function Write-DetailedExplanationOption {
+        [CmdletBinding()]
+        param(
+            [Parameter(Mandatory = $true)]
+            [string]$Message,
+            [Parameter(Mandatory = $true)]
+            [switch]$ShowDetailedExplanation
+        )
+        if ($ShowDetailedExplanation) {
+            Write-Host "`t`t$message"
+        } else {
+            Write-Verbose $message
+        }
+    }
+
     function Get-GroupObjectId {
         [OutputType([string])]
         param(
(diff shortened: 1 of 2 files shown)


M365/MDO/MDOThreatPolicyChecker.ps1
```diff
@@ -449,7 +526,7 @@ begin {
     SetWriteVerboseAction ${Function:Write-DebugLog}
     SetWriteWarningAction ${Function:Write-HostLog}
 
-    $BuildVersion = ""
+    $BuildVersion = "25.01.23.2258"
```
iserrano76, line 529: We need to clean version

M365/MDO/MDOThreatPolicyChecker.ps1
```diff
@@ -918,5 +1009,8 @@ process {
             }
         }
     }
+    if (-not $ShowDetailedExplanation) {
+        Write-Host ("`nFor details about why a policy applies to a recipient, use the -ShowDetailedExplanation parameter and run this script again.")
+    }
     Write-Host " "
-}
+}
```
iserrano76, line 1016: It looks it doesnt ends with a empty end line:
image

dpaulson45:
When doing multiple changes, even when small, each relatable change should be their own commit at the very least. If the pull request is a lot of random changes, then that should even be broken out into multiple PRs to make things easier to review and easier to understand when trying to figure out why things were changed in the past.

M365/MDO/MDOThreatPolicyChecker.ps1
```diff
@@ -279,7 +288,12 @@ begin {
             $emailInRule = $emailExceptionInRule = $groupInRule = $groupExceptionInRule = $domainInRule = $domainExceptionInRule = $false
 
             if ($Outbound) {
-                Write-Verbose "Checking outbound rule $($rule.Name)"
+                $message = "Checking outbound spam rule: `"$($rule.Name)`""
+                if ($ShowDetailedExplanation) {
+                    Write-Host "`t`t$message"
+                } else {
+                    Write-Verbose $message
+                }
```
dpaulson45, line 296: Personally, instead of doing this same action over and over and over again, I would recommend making a function called `Write-DetailedExplanationOption` that would contain this `if` statement that would take the passed value message and then determine if we need to `Write-Host` or `Write-Verbose` depending on the `$ShowDetailedExplanation` value.

dpaulson45: By also doing it this way, it greatly reduces the lines of code within your script.

dpaulson45 (changes requested):

dpaulson45 (approved):

PR #2364, opened 2025-07-22
Update DLT365Groupsupgrade.ps1 script

hazemembaby:
**Issue:**
Describe what the issue is you are addressing "remove DebugMemberRestriction function as no longer required"

**Reason:**
Describe what the reason is for making the change "the condition is no longer required"

**Fix:**
Short description of the fix "removed the function code"

**Validation:**
Provide if applicable


lusassl-msft:
Could you explain why the condition is no longer required?

hazemembaby:
> Could you explain why the condition is no longer required?

As per last testing for the issue by SPM and on my lab tenant the condition is no longer effective in preventing the group from being migrated so there's no need to have that condition added any more on the script.

PR #2365, opened 2025-07-23
Fix error handling in -SlowTraversal path

bill-long:
When we hit an error getting statistics during slow traversal, we would dump an int to the pipeline which broke the script.

This fixes that issue and some related error handling.


bill-long:

PR #2371, opened 2025-08-01
Include GUID when sending emails to us

dpaulson45:
**Issue:**
When customers send emails to us, they tend to have a similar subject which make it harder to try to find the email chain. 

**Reason:**
This is to help with trying to find the conversation of an email

**Fix:**
Try to provide a suggested subject. 

**Validation:**
Lab tested


dpaulson45:
image

lusassl-msft (approved):

PR #2379, opened 2025-08-07
Configure 3p to have HU to be SU

dpaulson45:
**Issue:**
Need to switch 3p App script to be SU vs HU

**Validation:**
Tested


lusassl-msft:
Issue #2380

PR #2295, opened 2025-03-12
Health Checker to Multi Threaded operation

dpaulson45:
**Issue:**
Currently, the Exchange Health Checker is only working on a single task for a single server thus making it slow because we are collecting so much data and trying to analyze it. 

**Fix:**
Made the Exchange Health Checker to be multi-threaded to allow for quicker data collection and review to occur. 

**Validation:**
Lab tested/customers tested

PR #2387, opened 2025-08-14
Enhance ConfigureExchangeHybridApplication.ps1 based on customer and internal feedback

lusassl-msft:
**Description:**
- `ConfigureAuthServer` logic simplified
  - We now copy the `DomainName` values from the corresponding `MicrosoftACS` Auth Server if there is one
  - We now compare the `AcceptedDomain` on the on-premises side with the domains returned via Graph API call if no corresponding `MicrosoftACS` object was found and we use the matching domains
- Improvements to `Get-GraphAccessToken` to make it more robust
- Improvements to `Remove-CertificateFromAzureServicePrincipal` to return more context about the operation that was performed

**Validation:**
Lab

### DIFF
--- a/Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
+++ b/Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
@@ -240,8 +240,8 @@ begin {
         param()
 
         Write-Host ""
-        Write-Host "If the script was already run from an elevated Exchange Management Shell (EMS), ensure the Auth Certificate is valid" -ForegroundColor Yellow
-        Write-Host "For more details, see: https://aka.ms/MonitorExchangeAuthCertificate" -ForegroundColor Yellow
+        Write-Warning "If the script was already run from an elevated Exchange Management Shell (EMS), ensure the Auth Certificate is valid"
+        Write-Warning "For more details, see: https://aka.ms/MonitorExchangeAuthCertificate"
     }
 
     function Get-XForEnabledFeature {
@@ -300,7 +300,7 @@ begin {
     # Prevent the script from running on PowerShell Core - there are adjustments needed which must be tested before release
     # We can't use requires PSEdition Desktop because it's not compatible with PowerShell version 3 and 4
     if ($null -ne $PSVersionTable.PSEdition -and $PSVersionTable.PSEdition -eq "Core") {
-        Write-Host "This script isn't supported in PowerShell Core - use PowerShell 5.1 or earlier" -ForegroundColor Yellow
+        Write-Warning "This script isn't supported in PowerShell Core - use PowerShell 5.1 or earlier"
 
         return
     }
@@ -347,7 +347,7 @@ begin {
 
     if ($null -eq $organizationGuid -and
         -not $Script:ResetFirstPartyServicePrincipalKeyCredentials) {
-        Write-Host "Unable to query the guid of the Exchange organization - please try to run the script again" -ForegroundColor Yellow
+        Write-Warning "Unable to query the guid of the Exchange organization - please try to run the script again"
 
         return
     }
@@ -368,7 +368,7 @@ begin {
     #region Prerequisites
     # Make sure that PowerShell runs in elevated mode - if it doesn't we don't need to proceed further - stop the script run
     if (-not (Confirm-Administrator)) {
-        Write-Host "This script must be executed in elevated mode - start the PowerShell as an Administrator and try again" -ForegroundColor Yellow
+        Write-Warning "This script must be executed in elevated mode - start the PowerShell as an Administrator and try again"
 
         return
     }
@@ -427,9 +427,9 @@ begin {
 
         # Script must be executed via EMS if 'UpdateCertificate' parameter is used and CertificateMethod is set to 'Automated'
         if ($isAutomatedCertificateUpload) {
-            Write-Host "To perform the automated export and upload of the Auth Certificate, the script must be executed on an Exchange server" -ForegroundColor Yellow
-            Write-Host "Make sure to run the script from an elevated Exchange Management Shell (EMS)" -ForegroundColor Yellow
-            Write-Host "Otherwise you can specify the certificate by using the '-CertificateInformation' parameter" -ForegroundColor Yellow
+            Write-Warning "To perform the automated export and upload of the Auth Certificate, the script must be executed on an Exchange server"
+            Write-Warning "Make sure to run the script from an elevated Exchange Management Shell (EMS)"
+            Write-Warning "Otherwise you can specify the certificate by using the '-CertificateInformation' parameter"
             Show-AuthCertificateInvalidWarning
 
             return
@@ -437,7 +437,7 @@ begin {
 
         # Script must be executed via EMS if 'ConfigureAuthServer' parameter is used - this is because we need to run the 'Set-AuthServer' cmdlet
         if ($Script:ConfigureAuthServer) {
-            Write-Host ($notRunViaEmsString -f "Auth Server") -ForegroundColor Yellow
+            Write-Warning ($notRunViaEmsString -f "Auth Server")
             Show-AuthCertificateInvalidWarning
 
             return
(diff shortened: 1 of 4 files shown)


Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
```diff
@@ -541,7 +550,23 @@ begin {
...
+
+            if (-not [System.String]::IsNullOrEmpty($Script:CustomClientId)) {
+                Write-Verbose "CustomClientId $Script:CustomClientId was provided and will be used"
+                $getGraphAccessTokenParams.Add("ClientId", $Script:CustomClientId)
+            }
+
+            $graphAccessToken = Get-GraphAccessToken @getGraphAccessTokenParams
+
+            if ($null -eq $graphAccessToken) {
+                Write-Host "Failed to acquire an access token - the script cannot continue" -ForegroundColor Yellow
```
dpaulson45: By chance should we use `Write-Warning` for these instead?

Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
```diff
@@ -793,89 +818,123 @@ begin {
...
         try {
-            # Identify all Remote Routing Domains as we need to add them to the Auth Server object via DomainName parameter
-            $targetDeliveryDomains = Get-RemoteDomain | Where-Object {
-                $_.TargetDeliveryDomain -eq $true
-            }
+            $authServers = Get-AuthServer -ErrorAction Stop
         } catch {
-            Write-Host "We couldn't query the Remote Domains configured in your organization" -ForegroundColor Yellow
+            Write-Host "Unable to run the 'Get-AuthServer' cmdlet" -ForegroundColor Yellow
             Write-Verbose "We hit the following exception: $_"
```
dpaulson45: Since we don't have advanced logging going here like in HC, we should maybe get more details vs just the string value of the exception maybe? Just a thought.

Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
```diff
@@ -793,89 +818,123 @@ begin {
...
 
-            return
+            if ($evoStsAuthServer.Count -le 0) {
+                Write-Host "We did not find an Auth Server which is configured for your tenant" -ForegroundColor Yellow
+
+                return
+            }
+
+            if ($evoStsAuthServer.Count -gt 1) {
+                Write-Host "More than one EvoSTS Auth Server was found that is configured for your tenant" -ForegroundColor Yellow
```
dpaulson45: What should the customer or engineer do if they find this scenario? I am assuming that this should be rare, correct? But we should provide an idea to address this.

Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
```diff
@@ -793,89 +818,123 @@ begin {
...
+
+                return
+            }
         }
 
         # We've detected a matching Auth Server object which we'll configure for dedicated Exchange hybrid application feature
-        Write-Host "Auth Server: $($authServer.Identity) was identified as valid hybrid configured object"
+        Write-Host "'$($evoStsAuthServer.Identity)' was identified as matching Auth Server"
+        Write-Verbose "Previous DomainName entries: $([System.String]::Join(", ", $evoStsAuthServer.DomainName))"
```
dpaulson45: Recommended, not required. 

```suggestion
        Write-Verbose "Previous DomainName entries: $([System.String]::Join(", ", [array]$evoStsAuthServer.DomainName))"
```

This forces it to always work, found this out in other PR. 

https://github.com/microsoft/CSS-Exchange/pull/2295/commits/d77560ebd7ef7a882555a63951bfcf699db57b41

Shared/AzureFunctions/Get-GraphAccessToken.ps1
```diff
@@ -33,37 +36,119 @@ function Get-GraphAccessToken {
     begin {
         Write-Verbose "Calling $($MyInvocation.MyCommand)"
 
-        $clientId = "1950a258-227b-4e31-a9cf-717495945fc2" # Well-known Microsoft Azure PowerShell application ID
+        function ConvertFrom-QueryString {
```
dpaulson45, line 46: Let's add a comment here that explains with an example of what we are doing here.

Shared/AzureFunctions/Get-GraphAccessToken.ps1
```diff
@@ -33,37 +36,119 @@ function Get-GraphAccessToken {
...
+                Write-Host "State mismatch detected! Expected '$state', got '$returnedState'" -ForegroundColor Red
+
+                return
+            }
+
+            $code = $queryString["code"]
+
+            if (-not $code) {
+                Write-Host "Authorization code is missing in callback" -ForegroundColor Red
+
```
dpaulson45, line 151: Should we ever expect something like this to happen, if so should we provide what to do if this does happen?

Shared/GraphApiFunctions/Remove-CertificateFromAzureServicePrincipal.ps1
```diff
@@ -34,115 +34,130 @@ function Remove-CertificateFromAzureServicePrincipal {
...
-                $keyCredentialsList.Add([PSCustomObject]@{
-                        key   = $key.key
-                        type  = $key.type
-                        usage = $key.usage
-                    })
+                    # If the certificate has expired and RemoveExpiredCertificates is true, do not retain it
+                    if ($RemoveExpiredCertificates) {
+                        [DateTime]$expDate = $key.endDateTime
+
+                        if ($expDate -lt (Get-Date)) {
```
dpaulson45: Wouldn't the expired data be stored in UTC? We should make sure that we are comparing the correct time zone information, so we aren't removing a cert that is still valid for a few hours.

dpaulson45:
@lusassl-msft please update the docs to state if you run into issues to review the debug log and its location to have a better understanding of the issue.

Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
lusassl-msft: Will be done with the next commit.

Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
lusassl-msft: Good catch. Added!

Shared/AzureFunctions/Get-GraphAccessToken.ps1
lusassl-msft: Done.

Shared/AzureFunctions/Get-GraphAccessToken.ps1
lusassl-msft: The `state` parameter binds the authorization response to this specific client request and provides CSRF protection. The client generates a random, unguessable value, stores it locally, sends it in the /authorize request, and verifies it on callback. If the returned `state` doesn’t match the stored value, the response is ignored - preventing stray redirects (e.g., from another tab/app/tenant) and attacker attempts to have the client redeem unsolicited results.

So, yes - it's possible that we end up here but it's unexpected and this code should prevent the client from continue processing => we don't redeem the returned code for an access token.

This is a very, very rare situation.

Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
lusassl-msft: It's not a common thing. I'll add additional information with the next commit.

Shared/GraphApiFunctions/Remove-CertificateFromAzureServicePrincipal.ps1
lusassl-msft: Good catch. Fixed!

Hybrid/ConfigureExchangeHybridApplication/ConfigureExchangeHybridApplication.ps1
lusassl-msft: For now, I'd keep it as is but moved the exception to the `Write-Host` (now `Write-Warning`) so that it's visible to the user. We can improve logging with another update once we're clear which information we'd need.

PR #2389, opened 2025-08-14
Remove Invoke-ScriptBlockHandler from Shared to avoid Health Checker from using it

dpaulson45:
**Reason:**
With Health Checker being multiple threaded, we no longer want to use `Invoke-ScriptBlockHandler` so we need to remove it from all the Shared functions. 

**TODO**
- [x] Squash the commits properly
- [x] Update function descriptions for the ones that we changed
- [x] Verify that all the scripts modified that they are working properly

**Validation:**
Lab tested

### DIFF
--- a/Shared/Get-ExSetupFileVersionInfo.ps1
+++ b/Shared/Get-ExSetupFileVersionInfo.ps1
@@ -1,44 +1,42 @@
 ﻿# Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-. $PSScriptRoot\Invoke-ScriptBlockHandler.ps1
+<#
+.DESCRIPTION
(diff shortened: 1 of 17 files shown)


Shared/Get-ExSetupFileVersionInfo.ps1
```diff
@@ -1,46 +1,36 @@
 ﻿# Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-. $PSScriptRoot\Invoke-ScriptBlockHandler.ps1
+. $PSScriptRoot\Invoke-CatchActionError.ps1
```
dpaulson45: Need to look back into this because if this is intended to be inside of `Invoke-ScriptBlockHandler` this will not be there.

PR #2401, opened 2025-08-22
Add ExchangeOrgGUID parameter, PS elevation and Entra Auth response

trudolf-msft:
1. Added ExchangeOrgGUID parameter to manually use the GUID instead of gathering from AD (use case: T0 Entra admin in AVD is not domain-joined)

2. Moved the check if PowerShell is elevated to export certificate section (use case: T0 Entra admin doesn't have those permissions)

3. Added HTTP response note for 403 error about insufficient permissions


dpaulson45:
@trudolf-msft, you just committed a full script that is built out after the build process vs the individual source files that we utilize to make managing the source code easier for scripts. 

This is going to be rejected.

PR #2448, opened 2025-10-16
TcpListener in extra

jcoiffin:
**Issue:**
After updating to Windows 11 build 26200.6899, ExTRA fails with:
Exception calling "GetResult" with "0" argument(s): "The request is not supported."

**Reason:**
HttpListener appears to be legacy API and seems to be blocked by Windows 11 stricter network isolation
I don't have the exact reason on Windows 11

**Fix:**
Proposing this change using TcpListener instead of HttpListener  which appears to fix the issue

### DIFF
--- a/Diagnostics/ExTRA/ExTRA.ps1
+++ b/Diagnostics/ExTRA/ExTRA.ps1
@@ -76,49 +76,64 @@ if ($null -ne $alreadySelectedTags) {
 
 $selectionTableJson = $selectionTable | ConvertTo-Json -Depth 4
 
-$httpListener = New-Object System.Net.HttpListener
-$httpListener.Prefixes.Add($uri)
-$httpListener.Start()
+$listener = [System.Net.Sockets.TcpListener]::new([System.Net.IPAddress]::Loopback, 5002)
+$listener.Start()
 
-& explorer.exe $uri
+Start-Process "explorer.exe" $uri
 
 try {
-    while ($httpListener.IsListening) {
-        $task = $httpListener.GetContextAsync()
-        while (-not $task.AsyncWaitHandle.WaitOne(100)) {
+    while ($listener.Server.IsBound) {
+
+        if (-not $listener.Pending()) {
             Start-Sleep -Milliseconds 100
+            continue
         }
 
-        $context = $task.GetAwaiter().GetResult()
+        $client = $listener.AcceptTcpClient()
+        $stream = $client.GetStream()
+        $reader = New-Object IO.StreamReader($stream)
+        $writer = New-Object IO.StreamWriter($stream)
+        $writer.AutoFlush = $true
 
-        if ($context.Request.HttpMethod -eq "PUT") {
-            $context.Response.StatusCode = 200
-            $context.Response.Close()
+        # Read request line
+        $requestLine = $reader.ReadLine()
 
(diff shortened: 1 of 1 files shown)


dpaulson45:
@jcoiffin Please confirm if you are on the Insider Program.

jcoiffin:
> Insider Program

I don't think so.
let me know how to check

dpaulson45:
> > Insider Program
> 
> I don't think so. let me know how to check

image

Once you go into that section, it should show you if you are in it or not.

jcoiffin:
Thanks thus I am not in insider program.
My colleagues with same behavior when moving to this build are not part of this program as well

Diagnostics/ExTRA/ExTRA.ps1
```diff
@@ -76,49 +76,64 @@ if ($null -ne $alreadySelectedTags) {
...
 $selectionTableJson = $selectionTable | ConvertTo-Json -Depth 4
 
-$httpListener = New-Object System.Net.HttpListener
-$httpListener.Prefixes.Add($uri)
-$httpListener.Start()
+$listener = [System.Net.Sockets.TcpListener]::new([System.Net.IPAddress]::Loopback, 5002)
+$listener.Start()
 
-& explorer.exe $uri
+Start-Process "explorer.exe" $uri
```
bill-long, line 82: Why was the call operator changed to Start-Process?

Diagnostics/ExTRA/ExTRA.ps1
```diff
@@ -76,49 +76,64 @@ if ($null -ne $alreadySelectedTags) {
...
+        $headers = @{}
+        while (($line = $reader.ReadLine()) -and $line -ne "") {
+            $parts = $line.Split(":", 2)
+            if ($parts.Length -eq 2) { $headers[$parts[0].Trim()] = $parts[1].Trim() }
         }
 
-        if ($context.Request.HttpMethod -eq "GET") {
-            Write-Host "Showing tag selector UI in the default browser."
+        if ($requestLine -match "^PUT") {
+            Write-Host "Browser tab was closed without saving changes."
```
bill-long, line 114: Why was the pause removed from the PUT handler?

bill-long (changes requested):

bill-long (changes requested):
Also when I click Save I get an error. Did you test this @jcoiffin?

image

bill-long:
Due to all the issues with this code, we are closing this PR.

When submitting a PR, please ensure it is tested and all changes are explained. The PR should only change things related to the problem being addressed. For example, in this PR the pause was removed from the PUT handler for no apparent reason (it even had a comment explaining why it was there), and this change broke the ability to click Refresh.

Switching to TcpListener is the approach we plan to use, just not via this PR.

Diagnostics/ExTRA/ExTRA.ps1
jcoiffin: no reason , let's keep it
& explorer.exe $uri

Diagnostics/ExTRA/ExTRA.ps1
jcoiffin: You're right, I didn't handle the case of a user refreshing the page. I guess it is not very common.
I don't have the solution how to loop again after pausing in that case thus just closed for user to start again.

PR #2449, opened 2025-10-16
Add w3wp.exe to process exclusion list

DKhrebin:
Defender injects his modules to w3wp. This is not supported per https://learn.microsoft.com/en-us/troubleshoot/exchange/administration/avoid-unsupported-integration-methods This will nor prevent AMSI module(s) to be loaded.

**Reason:**
Load Defender's AMSI module in supported way.

**Fix:**
Add w3wp to processes exclusions list 

**Validation:**
manual add to script in lab


dpaulson45:
Why are we adding this here? Is this for the `Test-ExchangeAVExclusions`?

DKhrebin:
It is for Set-ExchAVExclusions to add w3wp process to Defender's processes exclusions list.

dpaulson45:
Then we need to remove lines 357 - 359 here.

https://github.com/microsoft/CSS-Exchange/blob/5fd95fc3733fe42484a54139a228aefc72c15029/Diagnostics/AVTester/Test-ExchAVExclusions.ps1#L357

DKhrebin:
No. It is 2 different things. 
I want to change Set-ExchAVExclusions to have w3wp in exclusions list.
Test AVExclusions should continue verify if in w3wp we have 3rd party modules.

For Test script I'll add code to process 3rd party modules loaded by AMSI to transport or Diagnostics applications separately.

P.S. Feel free to ping me in Teams in case of questions :)

PR #2455, opened 2025-10-22
Add more Modern Sharing to Check-SharingStatus.ps1

Shanefe:
**Issue:**
Make output more readable / useful

**Fix:**
Hide output that is not often needed

**Validation:**
Tested with internal access on many user issues.

### DIFF
--- a/.build/cspell-words.txt
+++ b/.build/cspell-words.txt
@@ -4,6 +4,7 @@ amsi
 Antispam
 anob
 asmx
+Authenticode
 authsspi
 autodiscover
 bincirc
@@ -45,7 +46,6 @@ fltmc
 freb
 FYDIBOHF
 GCDO
-Get-AuthenticodeSignature
 globalsequence
 Hashtable
 HHMM
--- a/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerIISInformation.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Invoke-AnalyzerIISInformation.ps1
@@ -565,7 +565,11 @@ function Invoke-AnalyzerIISInformation {
             $binSearchFolders = (([xml]($_.Content)).configuration.appSettings.add | Where-Object {
                     $_.key -eq "BinSearchFolders"
                 }).value
-            $paths = $binSearchFolders.Split(";").Trim()
+            $paths = $null
+
+            if ($null -ne $binSearchFolders) {
+                $paths = $binSearchFolders.Split(";").Trim()
+            }
             $installPath = $exchangeInformation.RegistryValues.MsiInstallPath
             foreach ($binTestPath in  @("bin", "bin\CmdletExtensionAgents", "ClientAccess\Owa\bin")) {
                 $testPath = [System.IO.Path]::Combine($installPath, $binTestPath)
--- a/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityIISModules.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityIISModules.ps1
@@ -37,10 +37,11 @@ function Invoke-AnalyzerSecurityIISModules {
                     $modulesWriteType = "Red"
 
                     $iisModulesOutputList.Add([PSCustomObject]@{
-                            Module = $m.Name
-                            Path   = $m.Path
-                            Signer = "N/A"
-                            Status = "Not signed"
+                            Module       = $m.Name
+                            Path         = $m.Path
+                            Signer       = "N/A"
+                            Status       = "Not signed"
+                            PathNotFound = $m.PathNotFound
                         })
                 } elseif (($m.SignatureDetails.IsMicrosoftSigned -eq $false) -or
                     ($m.SignatureDetails.SignatureStatus -ne 0) -and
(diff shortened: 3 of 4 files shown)


Calendar/Check-SharingStatus.ps1
```diff
@@ -337,6 +382,34 @@ function GetOwnerInformation {
...
+    $CalFolderProps = Get-MailboxCalendarFolder -Identity "${Owner}:\$OwnerCalendarName"
+    Write-Host -ForegroundColor DarkYellow "`t ExtendedFolderFlags: $($CalFolderProps.ExtendedFolderFlags)"
+    if ($CalFolderProps.ExtendedFolderFlags -like "*SharedOut*") {
+        Write-Host -ForegroundColor Green "`t Calendar is Shared Out using Modern Sharing."
+    } else {
+        Write-Host -ForegroundColor Red "`t Calendar is not Shared Out using Modern Sharing."
+    }
+
+    Write-Host -ForegroundColor DarkYellow "`t Running 'Get-CalendarActiveSharingInformation -Identity "${Owner}:\$OwnerCalendarName"'"
+    $OwnerActiveSharingInfo = Get-CalendarActiveSharingInformation -Identity "${Owner}:\$OwnerCalendarName"
```
dpaulson45, line 387: Spell Check is currently complaining about this. 

Image

Based off code review internally, this looks like the correct spelling of it. 

Therefore, we need to tell Spell Check to ignore this word. If this is the only location of this, then you can just add it to the file like so: 

```
# cSpell:ignore ActiveShareesDataSet
```

Or if we think this might be used in different locations down the road, let's just add it to the main database of words to ignore.  #Resolved

Calendar/Check-SharingStatus.ps1
```diff
@@ -121,18 +123,16 @@ function ProcessCalendarSharingAcceptLogs {
...
 
     # Call the Export-MailboxDiagnosticLogs cmdlet and store the output in a variable
     try {
         # Call the Export-MailboxDiagnosticLogs cmdlet and store the output in a variable
         # -ErrorAction is not supported on Export-MailboxDiagnosticLogs
         # $logOutput = Export-MailboxDiagnosticLogs $Identity -ComponentName AcceptCalendarSharingInvite -ErrorAction SilentlyContinue
-
-        $logOutput = Export-MailboxDiagnosticLogs $Identity -ComponentName AcceptCalendarSharingInvite
+        Write-Host "Collecting AcceptCalendarSharingInvite logs for [$Identity] ..."
+        $logOutput = Export-MailboxDiagnosticLogs $Identity -ComponentName AcceptCalendarSharingInvite 2>$null
```
dpaulson45: Why are we redirecting the `Error` stream here when we are in a `try` `catch` block? #Resolved

Calendar/Check-SharingStatus.ps1
```diff
@@ -171,12 +191,21 @@ function ProcessCalendarSharingAcceptLogs {
...
+    #$csvObject | Where-Object { [DateTime]$_.Timestamp -gt (Get-Date).AddDays(-180) } | Format-Table -a Timestamp, SharedCalendarOwner, FolderName
+
+    Write-Host "Receiver [$Identity] has accepted copies of the shared calendar from the following recipients in the last 180 days:"
+    if ($csvObject.Timestamp.Substring(0, 2) -gt 12) {
+        Write-Verbose "Trying European DateTime Format - dd/MM/yyyy HH:mm:ss"
+        $culture = [System.Globalization.CultureInfo]::CreateSpecificCulture("en-GB")
+    } else {
+        $culture = [System.Globalization.CultureInfo]::CreateSpecificCulture("en-US")
+    }
+    $csvObject | Where-Object { [DateTime]::Parse($_.Timestamp, $culture) -gt (Get-Date).AddDays(-180) } | Format-Table -a Timestamp, SharedCalendarOwner, FolderName
```
dpaulson45: We might want to include this in a `try` `catch` block to handle if we don't properly parse the `DateTime`.  #Resolved

Calendar/Check-SharingStatus.ps1
```diff
@@ -207,13 +236,16 @@ function ProcessInternetCalendarLogs {
         $logOutput = Export-MailboxDiagnosticLogs $Identity -ComponentName InternetCalendar
     } catch {
         # Code to run if an error occurs
-        Write-Error "An error occurred: $_"
+        Write-Error "An error occurred processing InternetCalendar logs."
```
dpaulson45: It might still be helpful to include the Inner Exception. Makes it easier to debug down the road if required.  #Resolved

Calendar/Check-SharingStatus.ps1
Shanefe: I was trying to capture the Error that happens when the requested Log is not found.  Even with the Catch and 2>$null it still throws a huge error msg.
I will remove the 2> as it does not seem to help.

dpaulson45: Have you tried to include `-ErrorAction Stop` within your `Export-MailboxDiagnosticLogs`?

Calendar/Check-SharingStatus.ps1
Shanefe: Updated to better error msg.
Just means that there is no InternetCalendar Log

Calendar/Check-SharingStatus.ps1
Shanefe: -ErrorAction is not supported on Export-MailboxDiagnosticLogs

dpaulson45:

PR #2463, opened 2025-10-29
Update supported HC Exchange Versions

dpaulson45:
**Reason:**
Want to include `Exchange Server SE` and correcting format.


dpaulson45:

PR #2466, opened 2025-10-29
Improve performance for old CVE checks

### LINKED ISSUE #2331: [Issue] - Health Checker - Improve looped version checks
**Describe the issue**
While trying to improve the performance of Health Checker with Multi-Threading, I am noticing a trend that it is taking a long time to process in sections like `Invoke-AnalyzerSecurityCveCheck` and `Invoke-AnalyzerKnownBuildIssues`. While this does not change the outcome of the script, if we can improve the time here, we likely can cut out a few seconds of time to process during the analyzer per server which will result in a faster run of Health Checker overall.

**Expected behavior**
To have the code perform the most efficient way possible. 

**Script Output**
Currently with some examples: 
`Completed: Invoke-AnalyzerKnownBuildIssues and took 0.7249042 seconds`
`Completed: Invoke-AnalyzerKnownBuildIssues and took 0.781123 seconds`
`Completed: Invoke-AnalyzerSecurityCveCheck and took 13.5136554 seconds`
`Completed: Invoke-AnalyzerSecurityCveCheck and took 11.7195886 seconds`

**Additional context**
This is going to take some time to review and test out possible improvements, however, this shouldn't hold up the multi-threading work being performed before we release. Which is why we are creating an issue for this to keep this in mind for down the road.

dpaulson45:
**Reason:**
Improving the performance of a large loop that we don't need to be processing because we are on a newer build of Exchange than the SU. 

**Fix:**
Skip over old SU years based off the CU Release Date that we have. 

We will still process some things, but we will also skip over a lot as well. 

Resolved #2331 

**Validation:**
Lab tested

### DIFF
--- a/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityCveCheck.ps1
+++ b/Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityCveCheck.ps1
@@ -167,7 +167,7 @@ function Invoke-AnalyzerSecurityCveCheck {
         $year = [int]$value.Substring(3, 2)
 
         # Only add values that are greater than or equal to our current CU year build.
-        $cuReleaseYear = [int]([string]$exchangeInformation.BuildInformation.VersionInformation.ReleaseDate.Year).Substring(2)
+        $cuReleaseYear = $exchangeInformation.BuildInformation.VersionInformation.ReleaseDate.Year % 100
 
         if ($year -lt $cuReleaseYear) {
             continue


Diagnostics/HealthChecker/Analyzer/Security/Invoke-AnalyzerSecurityCveCheck.ps1
```diff
@@ -165,6 +165,14 @@ function Invoke-AnalyzerSecurityCveCheck {
     foreach ($value in $unsortedKeys) {
         $month = $value.Substring(0, 3)
         $year = [int]$value.Substring(3, 2)
+
+        # Only add values that are greater than or equal to our current CU year build.
+        $cuReleaseYear = [int]([string]$exchangeInformation.BuildInformation.VersionInformation.ReleaseDate.Year).Substring(2)
```
lusassl-msft: The following might be better to avoid multiple casting and string operations. What do you think?

```suggestion
        $cuReleaseYear = $exchangeInformation.BuildInformation.VersionInformation.ReleaseDate.Year % 100
```

PR #2490, opened 2026-01-23
Enhance the AMSI Request Body Scanning analysis

lusassl-msft:
**Description:**

Enhances the AMSI Request Body Scanning analysis to provide more detailed and actionable information about the configuration state.

When AMSI Request Body Scanning is enabled for specific protocols (rather than globally via `EnabledAll`), the output now lists which protocols are configured:

image

I've also added yellow warning indicator when AMSI Request Body Scanning is not enabled. 

I've also fixed the analyzer logic of the AMSI body scanning feature. The logic in Exchange Server is the following:

`EnabledAll == true` means that AMSI body scanning is enabled for any protocol. Starting with the Aug25SU, `EnabledAll == true` is the default without explicitly setting the value via SO.

If you're running the Aug25Su and you want to disable AMSI body scanning for all protocols except selected protocols, the logic is the following:
1. Set `EnabledAll == false` via SO
2. Create SO for any protocol which you want to enable (e.g., `EnabledEcp == true`, `EnabledOwa == true`)

**Validation:**
Lab

### DIFF
--- a/azure-pipeline-merge.yml
+++ b/azure-pipeline-merge.yml
@@ -24,7 +24,7 @@ extends:
             cd .\.build
             .\docs.ps1
           displayName: "Docs Check"
-        - pwsh: Write-Host "Temp Skip of Spell Check, Manually Check!" # .\.build\SpellCheck.ps1
+        - pwsh: .\.build\SpellCheck.ps1
           displayName: "Spell Check"
         - pwsh: |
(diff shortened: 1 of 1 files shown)


dpaulson45:
looks good, is it worth our time to add in pester testing covering the problem that this new code does address?

PR #2502, opened 2026-02-11
Check for cmdlet

Shanefe:
**Issue:**
Check for cmdlet before calling
Handle large calendars (blows Int)
Improve culture checker 

**Validation:**
Ran on SAW machine.

### DIFF
--- a/Calendar/Check-SharingStatus.ps1
+++ b/Calendar/Check-SharingStatus.ps1
@@ -191,16 +191,16 @@ function ProcessCalendarSharingAcceptLogs {
         $timestamp = $entry.Timestamp
         if ([string]::IsNullOrEmpty($timestamp)) { continue }
 
-        $datePart = $timestamp.Split(" ")[0]
-        if ([string]::IsNullOrEmpty($datePart)) { continue }
+        $monthOrDay = $timestamp.Split(" ")[0]
+        if ([string]::IsNullOrEmpty($monthOrDay)) { continue }
 
-        $dayPart = $datePart.Split("/")[0]
-        if ([string]::IsNullOrEmpty($dayPart)) { continue }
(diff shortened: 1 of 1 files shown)


Calendar/Check-SharingStatus.ps1
```diff
@@ -185,11 +185,26 @@ function ProcessCalendarSharingAcceptLogs {
...
+
+        $monthOrDay = $timestamp.Split(" ")[0]
+        if ([string]::IsNullOrEmpty($monthOrDay)) { continue }
+
+        $firstValue = $monthOrDay.Split("/")[0]
+        if ([string]::IsNullOrEmpty($firstValue)) { continue }
+
+        $valueToTest = 0
+        if ([int]::TryParse($firstValue, [ref]$valueToTest)) {
+            if ($valueToTest -gt 12) {
```
dpaulson45, line 202: The problem with this logic is it doesn't account for dates at the beginning of the of month. 

Example: Feb 11th, 2024 -> 11/02/2024 in European format. 

Would these time stamps be determined by the mailbox's current settings at the time of the entry created by what is set at `Get-MailboxRegionalConfiguration`? Would it be better to get that information and try to parse the events this way?  #WontFix

Shanefe: I don't really care if the Region is correct, I just don't want the code to 'fail' if it is in 'other' format.

The code loops through all the dates and if any are 'European' time format it is European.

Basically it just prints a list of the calendar that it has been accepted in the last 180 days. 
I only need the date to get to 180 days.

Shanefe: I would like to leave it as "good enough for what it is"
i.e. Won't Fix

Will change if needed.

PR #2523, opened 2026-03-25
Change pagefile display write type color to Green

trudolf-msft:
If recommended value is correct it should show as green. Customer was confused and did check manually


dpaulson45:
Synced with some engineers about this. We don't need to include a `Green` flag on this because we don't do this for everything and we just need to draw attention to the critical/common misconfigurations.

PR #2535, opened 2026-05-05
Add device code fallback for Graph token auth

aadnehovda:
**Issue:**
`Get-GraphAccessToken` currently assumes a local browser can be launched for the authorization code callback. This fails on Server Core and other non-interactive/remote PowerShell sessions where browser-based authentication is not available.

**Reason:**
Server Core and remote administration scenarios need an authentication path that does not require a browser on the target machine.

**Fix:**
Adds an `-AuthFlow` parameter with `Auto`, `AuthorizationCode`, and `DeviceCode` options. `Auto` keeps the existing authorization code flow when available, but uses device code authentication when running in Server Core, non-interactive, or remote PowerShell contexts, and also falls back to device code if browser launch/local callback fails. The device code implementation uses `Invoke-RestMethod` and handles expected OAuth polling responses such as `authorization_pending`, `slow_down`, `expired_token`, and `authorization_declined`.

**Validation:**
- Ran `.build\CodeFormatter.ps1 -Save -Branch main`
- Ran `.build\Build.ps1`
- Ran PowerShell parser validation for `Shared\AzureFunctions\Get-GraphAccessToken.ps1`
- Ran mocked auth-flow validation for browser URL encoding, browser-failure fallback, and Server Core immediate device-code selection
- Ran live PowerShell Direct validation on a local Server Core Hyper-V VM; confirmed `Auto` selected device code auth and reached the Microsoft device-code endpoint
- Attempted `.build\SpellCheck.ps1`; local host does not have `npm`, so cspell could not run locally


aadnehovda:
@microsoft-github-policy-service agree

dpaulson45:
This is currently on the backburner to review. Our plate is current full, not allowing a quick turnaround for some of these reviews. 

We have to be a little more careful with Shared code that is being changed to avoid any main scripts having issues. 

At a quick glance, this appears to be good but not validated on our end just yet.

PR #2561, opened 2026-06-26
Change Quick Exceptions to be the default

Shanefe:
**Issue:**
Exception collection can be too slow (hours)

**Fix:**
Switch to only collect exceptions for 3 months

**Validation:**
Testing for months.


dpaulson45:
@Shanefe SpellCheck is failing with `Froms`. 

If it is duplicate From, then do Senders or FromList or something like that. 

image

Shanefe:
Thanks.
I have updated the code to fix this and another issue.

Shane
________________________________

dpaulson45 (approved):

PR #2589, opened 2026-09-18
Fix modern public folder sync reliability

angryhussord:
**Issue:**
`Sync-ModernMailPublicFolders.ps1` does not expose an authorization endpoint override, assumes `EmailAddresses` supports `ToStringArray()`, and can lose summary output when a CSV is locked. A summary-writing failure can also be confused with an Exchange operation failure.

**Reason:**
Public-folder synchronization should support explicit authentication endpoints without manual connection edits, handle supported address representations consistently, and preserve an accurate operation record without repeating completed Exchange operations.

**Fix:**
- Add optional `AzureADAuthorizationEndpointUri` forwarding while preserving `ConnectionUri`, credentials, and the `Remote` prefix. Stop on connection failures.
- Normalize strings, string arrays, collections, and proxy-address objects across creation, updates, and consolidation. Preserve SMTP casing, legacy X500 addresses, and existing cloud-only address rules; reject unsupported entries explicitly.
- Retry file-opening sharing violations a bounded number of times, then write to an exclusively created sibling CSV with headers and a clear warning. Preserve existing files and report every output path.
- Keep successful Exchange operation accounting separate from summary output, retain folder GUIDs in error rows, and stop before validation or migration when summary initialization fails, including under the default `Continue` error preference.
- Update parameter and usage documentation, including output-file behavior, and add regression coverage using synthetic data and mocked Exchange commands.

**Validation:**
- Targeted Pester suite: 76 passed, 0 failed on PowerShell 7.
- Windows PowerShell 5.1 compatibility coverage, including file-permission and default-`Continue` startup regressions.
- Repository formatter and PSScriptAnalyzer completed without findings.
- Repository spell-check: 803 files, 0 issues.
- Repository build and generated standalone-script parsing succeeded.
- No live Exchange tenant operations were performed.


PublicFolders/MailPublicFolderSync/Sync-ModernMailPublicFolders.ps1
```diff
@@ -108,23 +116,103 @@ function WriteVerboseMessage() {
...
+        try {
+            $stream = [System.IO.File]::Open($targetPath, $fileMode, [System.IO.FileAccess]::Write, [System.IO.FileShare]::Read)
+            break
+        } catch [System.IO.IOException] {
+            $errorCode = $_.Exception.HResult -band 0xFFFF
+            if ($Initialize -and $errorCode -in @(80, 183)) {
+                break
+            }
+
+            if ($errorCode -notin @(32, 33)) {
```
dpaulson45, line 159: [CLI] - What are these particular area codes about?

PublicFolders/MailPublicFolderSync/Sync-ModernMailPublicFolders.ps1
```diff
@@ -108,23 +116,103 @@ function WriteVerboseMessage() {
...
+                Start-Sleep -Milliseconds 200
+            }
+        }
+    }
+
+    if ($null -eq $stream) {
+        $fallbackName = "{0}.{1}.csv" -f [System.IO.Path]::GetFileNameWithoutExtension($originalPath), (New-Guid).ToString("N")
+        $targetPath = Join-Path -Path ([System.IO.Path]::GetDirectoryName($originalPath)) -ChildPath $fallbackName
+        $stream = [System.IO.File]::Open($targetPath, [System.IO.FileMode]::CreateNew, [System.IO.FileAccess]::Write, [System.IO.FileShare]::Read)
+    }
```
dpaulson45, line 173: [CLI] - Should we be concern here about the stream being `$null` if `$Initialize` is not set? Is that possible?

dpaulson45: Not a major concern — the null path is intentional and handled safely.

It is reachable in the non-`Initialize` (append) case, but only after three consecutive `ERROR_SHARING_VIOLATION`/`ERROR_LOCK_VIOLATION` retries against the original file. Any other `IOException` rethrows before we get here, so `$stream` can only be `$null` when we have decided the original file is unusable right now.

The fallback then opens a sibling CSV with `CreateNew`. Because that file is brand new, `$needsHeader = $stream.Length -eq 0` is true and the header is written before the appended row, and `SummaryFileChanged` surfaces the new path to the operator. This is the intentional "bounded retry, then exclusive sibling CSV" behavior called out in the PR description.

PublicFolders/MailPublicFolderSync/Sync-ModernMailPublicFolders.ps1
dpaulson45: These are the low 16 bits of the Windows `IOException.HResult`:

- `32` = `ERROR_SHARING_VIOLATION` — file is in use by another process.
- `33` = `ERROR_LOCK_VIOLATION` — another process holds a lock on part of the file.
- `80` = `ERROR_FILE_EXISTS` and `183` = `ERROR_ALREADY_EXISTS` — returned by `CreateNew` when the target file already exists.

So `notin @(32, 33)` means "not a transient contention issue, rethrow", and `Initialize -and in @(80, 183)` means "we asked for `CreateNew` and the file already exists, stop retrying and let the fallback sibling-CSV path take over".

dpaulson45 (approved):

